PR #2, opened 2019-08-02
Add check to type using to encode dtype

cjh1:
Files written with Python 2.7 will be numpy.bytes_, Python 3 will be numpy.str_. Decode according to this type.


cjh1:
@psavery This is an alternative to #1, rather than save a bytes, it adds a special can on loading for the old python 2.7 files. I don't have a strong preference either way.

psavery:
This is a potentially better solution than #1. You can choose, @joelvbernier.

PR #8, opened 2019-11-11
integrated new transforms module in the hexrd3 code base

ovillellas:
All the work for the new transforms module put inside the hexrd3 codebase.

As hexrd3 is not yet complete, I've added the new transforms module as "transforms".

The old transforms is renamed and appears as legacy_transforms. All calls I've found are redirected to legacy_transforms until they the new transforms can be integration-tested (herxd3 needs to be functional for that).

Structure-wise:
- All C code is placed under the extensions directory, isolated from the Python code. However, 
   legacy transforms C code remains in the legacy-transforms directory.
- At runtime, extensions will appear as "hexrd.extensions.<name of extension>". This follows
   what was already done for the legacy transforms.
- tests for transforms are under a tests directory that is outside the source code for the package
   so that it is not included in installation. This is different from other tests that are physically
   close with what they test (the same directory).


joelvbernier:
@ovillellas We will revisit this soon!

ZackAttack614:
@psavery I would like to close this PR out. Thoughts?

psavery:
@ZackAttack614 Yes we can close this.

PR #12, opened 2020-02-11
Chunk stats

donald-e-boyce:
This updates the imageseries.stats.median() function to use chunks and return an intermediate image at each point, allowing the calling function to update a progress bar. I will follow up with another request applying these changes to other functions such average, max or percentile. Here is an example of the usage:

```python
import numpy as np

from hexrd import imageseries


stats = imageseries.stats

a = np.ones((50, 10, 6))
for i in range(a.shape[1]): # set values by row
    a[:,i,:] = i
ims = imageseries.open(None, 'array', data=a)


def run_chunk(ims, nchunk):
    img = np.zeros(ims.shape)
    for i in range(nchunk):
        img = stats.median(ims, chunk=(i, nchunk, img))
        print("%d/%d\n" % (i, nchunk), img)
    return img
```

### DIFF
--- a/hexrd/imageseries/load/array.py
+++ b/hexrd/imageseries/load/array.py
@@ -52,7 +52,7 @@ class ArrayImageSeriesAdapter(ImageSeriesAdapter):
         return self._data.dtype
 
     def __getitem__(self, key):
-        return self._data[key]
+        return self._data[key].copy()
 
     def __iter__(self):
         return ImageSeriesIterator(self)
--- a/hexrd/imageseries/stats.py
+++ b/hexrd/imageseries/stats.py
@@ -1,30 +1,20 @@
 """aggregate statistics for imageseries
 
-The functions here operate on the frames of an imageseries.
-Because there may be an large number of images, the images
-are processed in chunks, some number of rows at a time.
-All of the functions take a keyword arguments of "chunk" and "nframes".
-If "nframes" is greater than 0, it means use only that many frames of
-the imageseries; otherwise use all the frames.
-The "chunks" argument may be either None or a 3-tuple (i, n, img), where:
+The functions here operate on the frames of an imageseries and return a
+single aggregate image. For each function, there is a corresponding iterable
+that allows you to run the function in smaller bits; the bits are either groups
+of frames or groups of rows, depending on the function. The iterable returns a
+sequence of images, the last being the final result.
 
- i is the current chunk to compute,
- n is the total number of chunks, and
- img is the current image with same shape as the imageseries
+For example:
 
-If chunks is not None, then the the function needs to be called n times,
-and after each call an intermediate image is returned, the last being complete.
+#  Using the standard function call
+img = stats.average(ims)
 
-If chunks is None, then only one call is needed, and the number of chunks is
-determined by the STATS_BUFFER variable.
-
-For example, to find the median image for an imageseries "ims" using "nchunk" chunks:
-
-    img = np.zeros(ims.shape)
-    for i in range(nchunk):
-        img = op(ims, chunk=(i, nchunk, img))
-        print("%d/%d" % (i, nchunk), img)
-
+# Using the iterable with 10 chunks
+for img in stats.average_iter(ims, 10):
+    # update progress bar
+    pass
 """
 import numpy as np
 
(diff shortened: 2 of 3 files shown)


donald-e-boyce:
The stats module looks good now. All functions can be called with the chunk argument. The module documentation explains the chunking interface.

psavery:
@donald-e-boyce So, I started trying this out in hexrdgui, and I think it is working. However, I noticed that there seem to be some performance hits, and I wonder if we might be able to do anything to make these functions run faster.

I wrote a script to perform some timings, and it is attached below. I'm timing our `multiruby2_6_ff1_000048-cachefile.npz` file (~1.3 MB), and I can send it to you if you need it.

For the `max` function:
* master branch: 16.53 seconds
* this branch with `chunk=None`: 76.07 seconds
* this branch with 51 chunks: 817.68 seconds

This could be problematic, as it might make the loading time a lot longer.

[test.py.gz](https://github.com/HEXRD/hexrd/files/4225816/test.py.gz)

donald-e-boyce:
OK, I'll check it out.

donald-e-boyce:
I pushed up a new commit. There was a logging statement I got rid of, and I dropped use of the processed imageseries rectangle. There was also a genuine bug that reset the number of frames to the length of the imageseries, even if a smaller number was specified. I don't see how that affected the all-frames case, but somehow it seemed to speed up after that.

Still, it looks like there is a cost to chunking.  Below are timings for the stats.max function for the frame-cache used above. I think what is happening is that I set up the chunking to chunk by rows. That is because the median function was my model, and you need all the frames to compute the median, but you can compute the median row by row. For the average, max and min, you can break it into frames and combine the results later. In fact the max() function (previously) only ever had two frames loaded at a time, the running max image and the current frame. 

Now, to compute the median over several rows for a large number of frames, you end up loading all the frames anyway because of the way imageseries gets the frames---to get a subset of rows, you have to get the full frame first.  For a frame-cache, all images are available in memory, but as sparse matrices, so there is also some overhead converting to a full image. Likewise for HDF5.

>  testing stats.max
> 
>  ===== 0 chunks, 100 frames
> elapsed time:  3.01 seconds
> 
>  ===== 1 chunks, 100 frames
>  0/1 
> done
> elapsed time:  2.95 seconds
> 
>  ===== 10 chunks, 100 frames
>  0/10 
> .........done
> elapsed time:  5.7 seconds
> 
>  ===== 50 chunks, 100 frames
>  0/50 
> ......... 10/50 
> ......... 20/50 
> ......... 30/50 
> ......... 40/50 
> .........done
> elapsed time:  18.35 seconds
> 
>  ===== 0 chunks, 1441 frames
> elapsed time:  70.93 seconds
> 
>  ===== 20 chunks, 1441 frames
>  0/20 
> ......... 10/20 
> .........done
> elapsed time:  136.74 seconds
> 
>  ===== 50 chunks, 1441 frames
>  0/50 
> ......... 10/50 
> ......... 20/50 
> ......... 30/50 
> ......... 40/50 
> .........done
> elapsed time:  268.78 seconds
>

donald-e-boyce:
Here is the test script I've been using.

psavery:
@donald-e-boyce I don't see an attached script. Could you try attaching it again?

Here are the updated timings from my script, using the latest from this branch:

For the `max` function on `multiruby2_6_ff1_000048-cachefile.npz`:
* master branch: 16.46 seconds
* this branch with `chunk=None`: 34.80 seconds
* this branch with 51 chunks: 109.66 seconds

Also note that for this `multiruby` example, when we load it in hexrdgui, there are two cache files to load (not just one, as is being done in this test), and more aggregations might need to be performed for the dark images.

donald-e-boyce:
I did some profiling on this. It looks like most of the overhead is indeed loading the frames. In fact, a significant part of that is numpy.zeros() function, which is called in the toarray() method of the scipy.sparse.csr_matrix (compressed sparse row format). So in pulling the frame, much of the work is writing zeros for the full matrix.

donald-e-boyce:
I updated the stats module creating an iterable for each function, as you suggested. It is much cleaner and performs much better. Each function has a corresponding iterable. For example, to find the max, you can either use stats.max() as before or stats.max_iter(), which returns and iterable for use with the progress bar. See module documentation.

psavery:
@donald-e-boyce I'm still working on this, but it has been working very well so far, Don. Thank you very much!

psavery:
@donald-e-boyce This has been working great, Don. Thank you for the work!

I have one request right now, though. It looks like if you pass in `nchunk=1` and `nframes=1`, you get an exception like below. Could you modify the code so that we don't get an exception if `nchunk=1` and `nframes=1`?
```python
  File "test.py", line 145, in <module>
    time_all(nframes, nchunks)
  File "test.py", line 35, in time_all
    time_iter("max iterable", stats.max_iter, ims, nc, nf)
  File "test.py", line 94, in time_iter
    for i, img in enumerate(op(ims, nchunk, nframes=nframes)):
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/imageseries/stats.py", line 48, in max_iter
    stop = stops[s0]
```

psavery:
Or maybe just a different exception. I realize that we could also just change it on our end to just return the image if `nframes=1` as well...

donald-e-boyce:
I fixed it. I am just checking it. I will push shortly.

psavery:
@donald-e-boyce Great! Thank you don. This simplifies things a little bit on our part.

psavery:
I think this PR is ready for merging for me, if it is ready for you guys.

@donald-e-boyce @joelvbernier

cjh1:
@donald-e-boyce @joelvbernier If this one is ready to merge, lets do it, so we merge the hexrdgui portion.

PR #14, opened 2020-02-18
Remove unused variable

psavery:
In PlanarDetector.make_powder_rings(), if the material's plane data
has a tThWidth of None, and no delta_tth is passed to the function,
an exception occurs when `np.degrees(pd.tThWidth)` is attempted.

Upon looking further, though, it appears that the `delta_tth`
variable is not used beyond that point, so it seems to be not
necessary to set it.

Remove the variable so that the exception goes away.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -1901,8 +1901,6 @@ class PlanarDetector(object):
 
             if delta_tth is not None:
                 pd.tThWidth = np.radians(delta_tth)
-            else:
-                delta_tth = np.degrees(pd.tThWidth)
 
             # conversions, meh...
             del_eta = np.radians(delta_eta)


psavery:
@joelvbernier I ran into this issue with some changes I was making to hexrdgui. Hopefully, this PR will be fine...

psavery:
@joelvbernier Can we get this merged? This is needed to prevent errors in the Cartesian plot when the `maxtThWidth` is `None`, as of HEXRD/hexrdgui#262

joelvbernier:
@psavery perhaps we should catch the exception as the function needs the width one way or another to work properly.

https://github.com/HEXRD/hexrd/blob/79ebd31e43ca4b96a7cad60ff61ff2cc4fb14fba/hexrd/instrument.py#L1902-L1905

we could replace `delta_tth = np.degrees(pd.tThWidth)` with
```
try:
    delta_tth = np.degrees(pd.tThWidth)
except(AttributeError):
    raise RuntimeError(
        "either the tThWidth attribute must be set on PlaneData, or delta_tth must be specified."
    )
```

psavery:
Oh okay. I had just noticed that the variable `delta_tth` isn't used at all past that point. Maybe it should be?

joelvbernier:
Ah, sorry -- I see.  Yes, it looks like past that point it is assuming that the tThWidth attribute on PlaneData is set to something, which is necessary for getTThRanges() to work https://github.com/HEXRD/hexrd/blob/79ebd31e43ca4b96a7cad60ff61ff2cc4fb14fba/hexrd/instrument.py#L1915

so maybe a better catch is to do the following:

```
if delta_tth is None and pd.tThWidth is None:
    raise RuntimeError(
        "either the tThWidth attribute must be set on PlaneData, or delta_tth must be specified."
    )
```

rather than assigning to the after-that-point-unused variable `delta_tth`.  The intention was to allow the user to override whatever might be stored on the supplied `PlaneData` class; it has been ages since I dealt with `PlaneData` objects that didn't have the `tThWidth` set, so this slipped through. 

I double checked that in the event you hand in a tuple `(tth_min, tth_max)` in lieu of a `PlaneData` object, it raises a RuntimeError if `delta_tth` is `None`

https://github.com/HEXRD/hexrd/blob/79ebd31e43ca4b96a7cad60ff61ff2cc4fb14fba/hexrd/instrument.py#L1880

joelvbernier:
If this makes sense to you, I can make the change and commit.

psavery:
@joelvbernier Yeah that sounds good. I'll make a follow-up PR in hexrdgui to fix the exception there. Thank you!

PR #21, opened 2020-06-03
space group symmetries+unitcell class+whole powder pattern 

saransh13:
There should only be a few merge conflicts in materials.py. Other than that don't see any major issues merging.


joelvbernier:
We might need to make a feature branch on the org so we can test this.  touches lots of things.

PR #22, opened 2020-06-05
Remove versioner and just use setuptools_scm

cjh1:


cjh1:
@joelvbernier This removes versoneer and use standard setuptools_scm to get the version. From now on when you tag, I would not include the 'v' prefix to avoid problems with the conda build s.

PR #26, opened 2020-06-15
Return results from find_orientations()

psavery:
Hi @joelvbernier. I started taking a look at the find_orientations() code, 
and I was able to run a test on it successfully (for the 8 detector example).

However, I noticed that the `find_orientations()` function writes out
the results rather than returning them. I thought that it might make
it easier for us to write tests, and plot things in the GUI, if the
`find_orientations()` returned the data. So in this PR, it does that,
and also, the writing out of data is now in the CLI code instead of in the 
`find_orientations()` function.

Let me know what you think of the changes.

### DIFF
--- a/hexrd/cli/find_orientations.py
+++ b/hexrd/cli/find_orientations.py
@@ -1,5 +1,12 @@
 from __future__ import print_function, division, absolute_import
 
+import os
+import numpy as np
+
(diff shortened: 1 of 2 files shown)


joelvbernier:
I'm down with this.

PR #30, opened 2020-06-17
Don't remove 'v'

cjh1:
We are no longer prefixing our versions with 'v', so don't try to
remove it!


cjh1:
@joelvbernier This should fix it

PR #34, opened 2020-06-24
Cache config data from files. Allow setting the data.

psavery:
When config data is loaded from a file for the first time, store
the data on a member variable so that if the data is requested a
second time, it doesn't need to be read from the file again.

This also allows applications, such as hexrdgui, to set the config
data if the data has already been loaded into the application,
so that the data doesn't need to be loaded from the files again.


psavery:
@joelvbernier @donald-e-boyce

PR #35, opened 2020-06-24
Add __iter__ method to ProcessedImageSeries

psavery:
This is required in order to run find_orientations with a
ProcessedImageSeries object.

If frames were specified, then a generator is returned which
yields the frames one at a time as they are iterated through.

Otherwise, the underlying image series' iterator is used.


psavery:
@joelvbernier @donald-e-boyce

PR #38, opened 2020-06-26
Make saving optional in generate_eta_ome_maps

psavery:
We will probably not be saving the eta omega maps generated
by the GUI here.


cjh1:
Would be good to get this in for the indexing work ...

PR #40, opened 2020-06-29
Add common save method for Eta Omega maps

psavery:
This adds a static save method on `hexrd.xrdutil.EtaOmeMaps` (copied
directory from `hexrd.instrument.GenerateEtaOmeMaps.save()`), that
is used both by `EtaOmeMaps` and `GenerateEtaOmeMaps` to save their
data.

This is mostly done so that `EtaOmeMaps` can have a `save()` method
too, which is helpful for the GUI...


cjh1:
This is also needed for the indexing work.

Issue #39, opened 2020-06-29
Multithread/multiprocess eta omega map generation


psavery:
I believe that is done in [hexrd.instrument.extract_polar_maps()](https://github.com/HEXRD/hexrd/blob/4ad3e965d55b5402f9c8f29549f8bc92fe48e7de/hexrd/instrument.py#L662-L664).

@cjh1

joelvbernier:
I tested it yesterday and seems kosher.  Will merge when confirmed!

joelvbernier:
This was merged in #41,correct @cjh1 ?

cjh1:
Yes

PR #41, opened 2020-06-30
Extract polar maps parallel

cjh1:
This PR parallelizes `extract_polar_maps`. @joelvbernier I used the finer grain parallize you suggest. One my 4 core laptop I see the time go from 157 sec to 50 sec, more cores will show even better improvement. I am also only calculating `eta_edges` once for the first detector and reusing it.


cjh1:
Fixes #39

cjh1:
I need validate this producing the same result as the serial version, before merge ...

cjh1:
Seem to produce the same values.

PR #46, opened 2020-07-10
Move code to create clustering params into method

psavery:
This moves the majority of the code used to create the min samples
and the mean reflections per grain, for the clustering, into
a separate function. This is done so that we can call this function
from the GUI.

I did a CLI `hexrd find-orientations ...` before and after test,
to make sure the output didn't change, and a diff shows that the
output appears to be the same.

### DIFF
--- a/hexrd/findorientations.py
+++ b/hexrd/findorientations.py
@@ -537,6 +537,79 @@ def generate_eta_ome_maps(cfg, hkls=None, save=True):
     return eta_ome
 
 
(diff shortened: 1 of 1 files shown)


psavery:
@cjh1 @joelvbernier

joelvbernier:
Nice -- I'll say that we should go one step further in the repo and just make a top-level `hedm` package for all of these hedm-specific functions (including the current config and cli packages, along with the associated callbacks).  It will make organizing the source easier moving forward.

Issue #42, opened 2020-07-08
Add basic CI tests


cjh1:
We will run these in github actions. @joelvbernier  and @donald-e-boyce Are going to provide the scripts.

cjh1:
Resolved by #47

PR #47, opened 2020-07-13
Add find_orientations tests

cjh1:
This PR add two pytests, one for `generate_eta_ome_maps` and one for `find_orientations`. Note that I separated out the two, so the `find_orientations` test used pre-computed eta_ome_maps. This also adds a GitHub action workflow to run the test for every PR. It also fixes to bugs that the test revealed!


cjh1:
Fixes #42

cjh1:
@joelvbernier Can we get this merged, if I fix the conflicts?

PR #55, opened 2020-07-20
New material

cjh1:


cjh1:
@saransh13 I fixed up the issue, not quite sure why it appeared in this PR, but its resolved.

hexrd/material.py
```diff
@@ -69,48 +75,111 @@ class Material(object):
...
+    the dmin parameter is used to figure out the maximum sampling for g-vectors
+    this parameter is in angstroms
+    '''
+    DFLT_DMIN = _angstroms(0.2)
 
-    def __init__(self, name=DFLT_NAME, cfgP=None):
+    '''
+    default space group setting
+    '''
+    DFLT_SGSETTING = 0
```
psavery, line 117: This is getting set twice. Once here, and once on [line 89](https://github.com/HEXRD/hexrd/pull/55/files#diff-9e92e5bccd03281b17dc651f596ca74bR89).

saransh13: default should be zero. removing the first one in the next commit.

hexrd/material.py
```diff
@@ -348,6 +699,14 @@ def _set_atominfo(self, v):
 #
 
 
+xtal_sys_dict = {'cubic': 1,
+                 'tetragonal': 2,
+                 'orthorhombic': 3,
+                 'hexagonal': 4,
+                 'rhombohedral': 5,
```
psavery, line 706: @saransh13 Should this one be "trigonal" instead of "rhombohedral"?

I currently get an error with the material.h5 file as follows:
```python
>>> from hexrd.material import Material
>>> Material('Gd_hp', 'material.h5').dump_material('out.h5')
  File "<stdin>", line 1, in <module>
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 529, in dump_material
    AtomInfo['xtal_sys'] = xtal_sys_dict[self.unitcell.latticeType.lower()]
KeyError: 'trigonal'
```

psavery:
@saransh13 It looks like there might be a few functions still around in the `Material` class that need updating/removing.

For instance, in the GUI, we have been setting the lattice parameters like this:
```python
mat.latticeParameters = params
```

Even though this property setter still exists, it produces an error due to the removed SG:
```python
>>> mat = Material('CeO2', 'material.h5')
>>> mat.latticeParameters = [5.2]
  File "<stdin>", line 1, in <module>
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 637, in _set_latticeParameters
    self._lparms = self._toSixLP(self.sgnum, v)
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 268, in _toSixLP
    tmpSG = SG(sgn)
NameError: name 'SG' is not defined
```

The `_toSixLP` and `_readCfg` functions still use `SG`, so they might need to be updated/removed.

It looks like I can maybe use the `_rqpDict` from the `unitcell.py` file when setting the lattice parameters (which helps us take into account the required ones). Let me know if I should do otherwise.

cjh1:
Just as a heads up that our preliminary integration of the materials package seems to slow down the GUI in terms of loading a changing material properties. We can look at this when the first stage of integration is complete, but just wanted to record it here.

hexrd/material.py
psavery: If so, I'll take care of fixing it.

saransh13: @psavery yep, should be trigonal. My bad.

saransh13:
I'll push these fixes today. Do you want to keep the property setter in the `Material` class the same?

psavery:
@saransh13 Sure! Sounds fine to me.

saransh13:
@psavery can you check the latest pushes?

psavery:
@saransh13 I think it looks great so far. Thank you!

hexrd/material.py
```diff
@@ -122,88 +184,297 @@ def __str__(self):
...
+        self._atominfo  = numpy.transpose(numpy.array(gid.get('AtomData'), dtype = numpy.float64))
+        self._U         = numpy.transpose(numpy.array(gid.get('U'), dtype = numpy.float64))
+
+        # read atom types (by atomic number, Z)
+        self._atomtype = numpy.array(gid.get('Atomtypes'), dtype = numpy.int32)
+        self._atom_ntype = self._atomtype.shape[0]
+
+        self._sgsetting = numpy.asscalar(numpy.array(gid.get('SpaceGroupSetting'), \
+                                        dtype = numpy.int32))
+        self._sgsetting -= 1
```
psavery, line 447: @saransh13 It looks like if I load in the HDF5 file and then dump it out immediately, the space group setting drops in value by 1, maybe because of this line here.

CeO2 for instance starts with "1". After load and dump, it is then "0". After load and dump again, it is then "-1".

This is the only main difference I found for round-tripping.

saransh13: `sgsetting` should always be 0. All calculations are performed in the standard setting, corresponding to the value 0. 
Will push a fix shortly.

saransh13: The problem was that that `sgsetting` was stored as "1" based indexing in the h5 file but used in "0" based indexing in the class. Everything is "0" based now. I've pushed the changes and also uploaded a "material_new.h5" on the hexrd google drive where everything is "0" based. You should use that file for testing instead.

psavery: @saransh13 Sounds great. I'll try it out. Thank you Saransh!

psavery:
Okay. Things are starting to look good in hexrdgui for the most part I think. I put up HEXRD/hexrdgui#496 that depends on this one.

cjh1:
@saransh13 @joelvbernier I think HEXRD/hexrdgui#496 is looking in good shape as a first past. So I think you can merge this when your are ready and then we will merge it.

cjh1:
@saransh13 @joelvbernier Can we merge this? Its already pretty large, would be great to get it in and iterate.

saransh13:
Yeah I'm okay with merging.

joelvbernier (approved):

PR #56, opened 2020-07-22
Fix name of find_orientations module

johnkit:
Fixes CLI exception when fit_grains runs find_orientations.


cjh1:
@joelvbernier Are we good to merge this?

PR #59, opened 2020-07-25
panel update via instrument

joelvbernier:


setup.py
```diff
@@ -8,7 +8,7 @@
 np_include_dir = os.path.join(numpy.get_include(), 'numpy')
 
 install_reqs = [
-    'fabio@git+https://github.com/joelvbernier/fabio.git@master',
```
cjh1, line 11: Are you changes now in a release, I didn't see one cut?

joelvbernier: Hmmm...  maybe not.  Just asked the silx maintainer when we can expect one.

cjh1: The rest of the changes look good, so if you want to remove this change, we can merge? Or wait for fabio ...

joelvbernier: Yeah, they are busy with the restart at ESRF, so it might be a bit...  I'll change it back.

PR #64, opened 2020-08-04
Add PyYAML Loader and Dumper to handle numpy.ndarray

### LINKED ISSUE #28: Handling array-like panel buffers (masks) in instrument config write
The current output method `instrument.HEDMInstrument().write_config()` is _not_ appropriate if the panel buffers are array-like.  There are two possible fixes, and perhaps we should implement both.
1. Add an HDF5 instrument config format, and store the buffer as a dataset (numpy ndarray)
2. In the case of array-like panel buffer, add a trigger in the YAML output to load buffer from a flat npy file that gets auto-generated.

Thoughts @cryos @psavery @cjh1 @donald-e-boyce ?

cjh1:
This is to support numpy array's in the instrument panel buffer.

Fixes #28

### DIFF
--- a/hexrd/config/dumper.py
+++ b/hexrd/config/dumper.py
@@ -1,0 +1,62 @@
+import yaml
+import numpy as np
+from pathlib import Path
+
+
+def _dict_path_by_id(d, value, path=()):
+    if id(d) == value:
+        return path
+    elif isinstance(d, dict):
+        for k, v in d.items():
+            p = _dict_path_by_id(v, value, path + (k, ))
+            if p is not None:
+                return p
+    elif isinstance(d, list):
+        for i, v in enumerate(d):
+            p = _dict_path_by_id(v, value, path + (str(i),))
+            if p is not None:
+                return p
+
+    return None
+
+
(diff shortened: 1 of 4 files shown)


joelvbernier:
@cjh1 -- any idea why the windows packaging test failed?

cjh1:
@joelvbernier I saw that yesterday, I tried restart it a could of times, no dice, I will dig into it some more ...

cjh1:
@joelvbernier A four restart did it!

joelvbernier:
@cjh1 -- do you have an example on how to use this?

cjh1:
@joelvbernier  So to dump out a dict that contains a numpy.ndarray, you would use the following:

```python
yaml.dump(d, stream=f, Dumper=NumPyIncludeDumper)
```

This will serialize the numpy array to a file relative to the YAML doc. Inside the YAML doc you will find the array replaced with something like `!include <numpy_file_path>`.

To load it back in:

```python
 yaml.load(f, Loader=NumPyIncludeLoader)
```

joelvbernier:
Ok, I will try it out!

PR #66, opened 2020-08-13
Allow specifying rmat_c and tvec_c in angles_to_cart

psavery:
This adds rmat_c and tvec_c as optional arguments to angles_to_cart.

My specific use case is that, when I compute the bounding box for the
Laue widths, I first compute the box in angles and then use
angles_to_cart to convert to Cartesian.

If the crystal's position is not at the origin, we need to specify
tvec_c in order for this to work properly.


psavery:
@joelvbernier

PR #70, opened 2020-08-14
Build package in container

cjh1:
This PR moves the build of the package into a older Linux container. This ensure that we build with an old version of glibc so the package will work on older distros as well. @joelvbernier this should resolve the issue your collaborator is seeing.

### DIFF
--- a/.github/workflows/container_build.sh
+++ b/.github/workflows/container_build.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+cd /github/workspace/
+yum install -y wget git gcc
+wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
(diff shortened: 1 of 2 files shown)


cjh1:
@psavery When this builds can you try the package produced for me, to check it works on your machine?

psavery:
Will do... on Monday :slightly_smiling_face:

joelvbernier:
This won't create any problems with current linux systems, correct?  If it is a pain to try to support older linux distros (like RHEL6), then we can just punt on this.

cjh1:
It won't, @psavery is going to do a final test for me.

psavery:
My review is not valuable enough to merit a merge...

cjh1:
@joelvbernier I am going to go ahead and merge this if that is ok?

PR #83, opened 2020-08-28
New distortion

joelvbernier:
This PR has the new distortion package implemented.  Tested instrument initialization, and applicaiton via the hexrdgui in polarview.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -63,9 +63,7 @@ from hexrd.crystallography import PlaneData
 from hexrd import constants as ct
 from hexrd.rotations import angleAxisOfRotMat, RotMatEuler
 from hexrd.config.dumper import NumPyIncludeDumper
-
-# FIXME: distortion kludge
-from hexrd.distortion import GE_41RT  # BAD, VERY BAD!!!
+from hexrd import distortion as distortion_pkg
 
 from skimage.draw import polygon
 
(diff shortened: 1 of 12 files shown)


hexrd/instrument.py
```diff
@@ -1832,10 +1835,7 @@ def distortion(self):
 
     @distortion.setter
     def distortion(self, x):
-        """
-        Probably should make distortion a class...
-        ***FIX THIS***
-        """
+        # FIXME: ne to reconcile check with new class type!
         assert len(x) == 2 and hasattr(x[0], '__call__'), \
```
joelvbernier, line 1839: incorrect assertion for now, need to fix this setter.

hexrd/xrdutil.py
```diff
@@ -774,8 +771,8 @@ def angularPixelSize(
         * assumes xy_det in UNWARPED configuration
         """
         xy_det = np.atleast_2d(xy_det)
-        if distortion is not None and len(distortion) == 2:
-            xy_det = distortion[0](xy_det, distortion[1])
+        if distortion is not None:  # !!! check this logic
```
joelvbernier, line 774: Seems ok.

joelvbernier:
@psavery -- didn't mean to burden you with this.  I am testing and it seems ok for now (espeially for such a big change!).

hexrd/transforms/xf.py
```diff
@@ -11,9 +11,9 @@
 #
```
joelvbernier, line 11: wtf is wrong with me; we don't even use this module!

psavery:
I'm gonna take a look today. Sorry for the delay...

joelvbernier:
This looks to be ok, and the places it touched are isolated to the overlays and the calibration utilities.  The latter are going to be moved into hexrd, and the former actually need to be modified to even use distortion anyhow (only affects the raw view when distortion is not `None`).  We can do that in a future PR.

PR #84, opened 2020-09-01
Add fit grains CI test

cjh1:


joelvbernier:
@cjh1 -- thoughts as to the failed check?

cjh1:
@joelvbernier I am working on it see #85 for details, its unrelated to this PR ....

joelvbernier:
ah shit -- do we need to pin a version?

cjh1:
@joelvbernier Not sure, what is causing it at the moment, will let you know.

cjh1:
@joelvbernier I think the latest errors are related to the distortion changes? Looks like the same issue @psavery reported?

psavery:
@cjh1 @joelvbernier Yeah, this looks like the same error I mentioned [here](https://github.com/HEXRD/hexrdgui/pull/490#issuecomment-685594592).

Here is the CI error:
```
        if self.distortion is not None:
            dist_d = dict(
                function_name=self.distortion.maptype,
>               parameters=self.distortion.params.tolist()
            )
E           AttributeError: 'list' object has no attribute 'tolist'
```

Not sure how the params is becoming a list instead of a numpy array.

joelvbernier (approved):

Issue #80, opened 2020-08-25
Inconsistencies in how 'working_dir' is applied in HEDM config


joelvbernier:
The 'working_dir' option in the HEDM config is a strange beast that is not oft employed.  It defaults to the [current working directory](https://github.com/HEXRD/hexrd/blob/4d488e4a4d9e73e20570913f26780a926835c0a4/hexrd/config/root.py#L128).  The inconsistencies arise in how various files are loaded.

- analysis dir -- prepends working_dir
- materials file -- prepends working_dir
- imageseries -- prepends working_dir
- instrument -- _doesn't_ prepend working_dir

PR #79 fixed the routing of the `find-orientations` output to prepend the working_dir.  It would appear that the instrument should be made to behave like the others; however regarding inputs, it may be more intuitive to specify paths relative the the pwd?  Need to decide one way or another.

PR #87, opened 2020-09-01
updated config to use working_dir consistently

donald-e-boyce:
To address this, I added a static method in the base config class that checks filename "correctness." If the filename is an absolute path, that is what gets used. If it is a relative path, then the working_dir is prepended. Finally, if a file exists for that path, then that is returned. Otherwise, IOError is raised.

I applied this to instrument config file and to the imageseries files loaded in the image_dict.

### DIFF
--- a/hexrd/config/config.py
+++ b/hexrd/config/config.py
@@ -1,6 +1,7 @@
 """Base Config class"""
 
 import logging
+import os
 
 from .utils import null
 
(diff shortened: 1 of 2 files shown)


donald-e-boyce:
Fixed pep 8 error. Also note this pull request refers to Issue #80 .

joelvbernier:
@donald-e-boyce and @cjh1 -- want to look into why the checks failed on this?

joelvbernier:
@donald-e-boyce , looks ok, @cjh1 is on the cause of the failed checks.  Regarding that, @cjh1 what is the meaning of "canceled" vs. "failing" checks?

joelvbernier:
As the failed tests are unrelated, going to get this merged.

cjh1:
@joelvbernier The distinction between "canceled" and "failing" checks is the following. The checks on Windows "failed", because of the numpy/setuptools issue, this causes the other platforms to be "cancelled", as the CI is setup so everything should pass, so if something fails it cancels everything ...

PR #89, opened 2020-09-02
Work around issue with setuptools and numpy

cjh1:
See: https://github.com/numpy/numpy/issues/17216


cjh1:
@joelvbernier I am going to use my powers to merge this, so we can move on!

PR #92, opened 2020-09-03
Add helper functions and a few fixes for Material class

psavery:
This adds helper functions to save/load materials HDF5 files.

It also adds `dmin` properties so that the dependent values get
re-computed when the dmin is modified, both on the unit cell
and on the material.

It also uses valWUnit for unit conversions (Material uses angstrom,
while UnitCell uses nanometers).

These bugs were fixed:

1. "rhombohedral" should be "trigonal" for crystal system
2. The dmin that gets passed to getHKLs is now converted to the correct units

This also increased the default dmin to 1 angstrom, since less than this generates
a lot of values, and slows down the start of hexrdgui.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -103,7 +103,7 @@ class Material(object):
     the dmin parameter is used to figure out the maximum sampling for g-vectors
     this parameter is in angstroms
     '''
-    DFLT_DMIN = _angstroms(0.2)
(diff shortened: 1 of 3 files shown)


psavery:
@cjh1 @saransh13

cjh1 (approved):
LGTM

cjh1:
I am going to go ahead and merge this.

PR #93, opened 2020-09-03
Add a few more updates for setters

psavery:
When a material's `sgnum` gets set, also set the unitcell's `sgnum` if it has been created.

When the unitcell's `sgnum` is set, update the point group and Laue group.


cjh1 (approved):
LGTM

PR #97, opened 2020-09-11
Fix U type check and add update_structure_factor() to Material

psavery:
The commit messages are pasted below

1. Use isinstance(var, np.floating) for numpy floats

`np.float` is an alias to Python's `float` type, thus
`type(var) == np.float` does not return True for numpy float types.
However, `isinstance(var, np.floating)` will work. For instance:

```python
>>> a = np.float64(1)
>>> type(a) == np.float
False
>>> isinstance(a, np.floating)
True
```

If we want to also check for the Python float in addition to numpy float,
we can do:

```python
isinstance(var, (np.floating, float))
```

This change fixes an issue where U was being interpreted as a tensor, when
it was actually a single float.

2. Add Material function to update structure factor

We will call this when certain unit cell parameters are modified.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -170,14 +170,7 @@ class Material(object):
             self._sgsetting)
 
         self._newPdata()
-        hkls = self.planeData.getHKLs(allHKLs=True)
-        sf = numpy.zeros([hkls.shape[0],])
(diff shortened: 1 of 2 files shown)


cjh1 (approved):
LGTM

PR #100, opened 2020-09-12
Wppf feature

saransh13:
Multiple different type of input is now supported for initializing the `LeBail` class.

1. Passing spectrum data as `numpy.ndarray`, as a text file name or as a `Spectrum` class.
2. Passing params as a `dict`, a yaml file or a `Parameters` class.
3. Passing phases as a `dict`, a yaml file or a `LeBail_phases` class.

All options were tested with the example ceria dataset uploaded to the HEXRD google drive.


saransh13:
@cjh1 @psavery this pull request should fix the data passing issues with LeBail class.

joelvbernier (approved):

PR #102, opened 2020-09-14
Calibration fixes

joelvbernier:
added a new fuction for converting bragg angles for diffraction center offset.


joelvbernier (commented):
Looks good -- calibration routines are not ready for primetime yet though.

PR #103, opened 2020-09-14
WPPF feature

### LINKED ISSUE #101: Initialize WPPF phases and materials from Material object
#100 is a nice improvement, and we can now initialize most of the WPPF classes with data rather than file names.

Although we can now initialize the phases with a phase object directly (as is done [here](https://github.com/HEXRD/hexrd/commit/3aa726ab40f13c18b0621784464cca1fa33fb706#diff-c142f9d2dde3f2f20b5a293b8380cf5fR1495-R1499)), these phase objects and their material objects require a material file in order for them to be created (such as [here](https://github.com/HEXRD/hexrd/blob/3aa726ab40f13c18b0621784464cca1fa33fb706/hexrd/WPPF.py#L1179-L1181)).

In hexrdgui, we already have the [Material](https://github.com/HEXRD/hexrd/blob/3aa726ab40f13c18b0621784464cca1fa33fb706/hexrd/material.py#L76) object loaded in memory. It would be nice if we could instantiate the WPPF phases and material objects with this material object.

Instead, for now, it looks like we will need to write out the material that is loaded in memory to a file in a temporary directory, and then provide the file path to the WPPF phases classes so it can read it back into memory.

Note: we cannot use the original material file for this because the user may have edited the material settings without writing out the new material.

saransh13:
`WPPF.Material_LeBail` can now be initialized by passing a list or an instance of `material.Material` class. This should resolve #101.

Additionally, the `bkgmethod` argument to the `LeBail` class is now a dictionary. If `spline` is chosen, the value is `None`, if `chebyshev` is chosen, then the value is the polynomial degree. This degree needs to be an integer.

### DIFF
--- a/hexrd/WPPF.py
+++ b/hexrd/WPPF.py
@@ -519,7 +519,10 @@ class Material_LeBail:
         self.dmin = material_obj.dmin.value
         self.sgnum = material_obj.unitcell.sgnum
         self.sgsetting = material_obj.sgsetting
-        self.lparms = [x.value for x in material_obj.latticeParameters]
+        if(material_obj.latticeParameters[0].unit == 'nm'):
+            self.lparms = [x.value for x in material_obj.latticeParameters]
+        elif(material_obj.latticeParameters[0].unit == 'angstrom'):
+            self.lparms = [x.value/10.0 for x in material_obj.latticeParameters]
 
         self.dmt = material_obj.unitcell.dmt
         self.rmt = material_obj.unitcell.rmt
@@ -1569,7 +1572,7 @@ class LeBail:
     def chebyshevfit(self):
         degree = self.bkgmethod['chebyshev']
         p = np.polynomial.Chebyshev.fit(
-            self.tth_list, self.spectrum_expt._y, degree, w=self.weights)
+            self.tth_list, self.spectrum_expt._y, degree, w=self.weights**4)
         self.background = Spectrum(x=self.tth_list, y=p(self.tth_list))
 
     def selectpoints(self):
@@ -1592,6 +1595,8 @@ class LeBail:
         '''
         >> @AUTHOR:     Saransh Singh, Lawrence Livermore National Lab, saransh1@llnl.gov
         >> @DATE:       06/08/2020 SS 1.0 original
+                        09/11/2020 SS 1.1 multiple different ways to initialize phases
+                        09/14/2020 SS 1.2 added phase initialization from material.Material class
         >> @DETAILS:    load the phases for the LeBail fits
         '''
 
--- a/hexrd/WPPF.py
+++ b/hexrd/WPPF.py
@@ -516,9 +516,10 @@ class Material_LeBail:
         from an instance of the material.Material class. this option is
         provided for easy integration of the hexrdgui with WPPF.
         '''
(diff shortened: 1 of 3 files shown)


joelvbernier:
PEP8 hates you @saransh13 🤣

Do you have a linter in whatever editor you are using that understands PEP8 source rules?  Spyder works well for this, but I am pretty sure Atom has plugins too.  Not sure about that Sublime garbage...

joelvbernier:
Good idea to fix the source, if nothing else, for consistency.

saransh13:
@joelvbernier come on, Sublime has all this. Will push fixes today.

hexrd/WPPF.py
```diff
@@ -1243,11 +1248,11 @@ def _nm(x):
...
-        expt_spectrum=None,
-        params=None,
-        phases=None,
-        wavelength={'kalpha1':_nm(0.15406),'kalpha2':_nm(0.154443)},
-        bkgmethod='spline'):
+                 expt_spectrum=None,
+                 params=None,
+                 phases=None,
+                 wavelength={'kalpha1':_nm(0.15406),'kalpha2':_nm(0.154443)},
+                 bkgmethod={'spline':None}):
```
psavery: FYI, setting dicts as default arguments falls under the category of using "mutable default arguments", which can lead to difficult to track bugs, because python stores copies of the default arguments. See [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) for an example.

We would typically do something like this instead:
```python
def my_func(wavelength=None, bkgmethod=None):
    if wavelength is None:
        wavelength = {'kalpha1':_nm(0.15406),'kalpha2':_nm(0.154443)}
    if bkgmethod is None:
        bkgmethod = {'spline':None}
```

psavery (approved):
This works for me in a small test I did.

Are we going to add `Rietveld` to the GUI as well? If so, we might need to modify the phase and material in the same way for `Rietveld` also.

hexrd/WPPF.py
saransh13: Got it. I can fix these.

saransh13:
`Rietveld` class will eventually be added tot he GUI too. But I don't think it's a feature we want to show at the demo.
I can slowly start to make those changes.

hexrd/WPPF.py
```diff
@@ -463,46 +494,80 @@ def _readHDF(self, fhdf, xtal):
...
+        self.lparms = list(gid.get('LatticeParameters'))
         fid.close()
 
+    def _init_from_materials(self, material_obj):
+        '''
+        this function is used to initialize the materials_lebail class
+        from an instance of the material.Material class. this option is
+        provided for easy integration of the hexrdgui with WPPF.
+        '''
+        self.dmin = material_obj.dmin.value
```
psavery, line 519: Do we need to make sure this is in nanometers? We are using angstroms in the GUI, and setting `_angstroms()` on the material's `dmin`.

saransh13: yep, nice catch.

cjh1:
@saransh13 Is this ready to merge?

saransh13:
Yes, please go ahead.

PR #104, opened 2020-09-15
Fix signal.fft() call to signal.fftconvolve()

johnkit:
It appears there was a bug injected as part of commit 55a797c0 (31-Jul-2020) for pep8 compliance. The PR changes the fast_snip1d() method back to calling signal.fftconvolve().


cjh1 (approved):
Good catch, thanks @johnkit

Issue #99, opened 2020-09-11
Add "lmfit" to list of requirements


psavery:
```python
Traceback (most recent call last):
  File "example_LeBail.py", line 1, in <module>
    from hexrd.WPPF import Spectrum, LeBail, Rietveld#, LeBail_Asym, Rietveld_Asym, LeBail_FPA
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/WPPF.py", line 11, in <module>
    import lmfit
ModuleNotFoundError: No module named 'lmfit'
```

saransh13:
one thing I want to note here that `lmfit1.0.1` has some strange behavior where the least square minimization routines don't work as expected. `lmfit1.0.0` does not have those issues and should be preferred.

PR #105, opened 2020-09-16
Add lmfit to list of requirements for hexrd

psavery:
It is required to use WPPF.

The version is set at 1.0.0 for reasons mentioned in hexrd/hexrd#99.

Fixes: #99

I think I might need to add it to the [meta.yaml](https://github.com/HEXRD/hexrd/blob/master/conda.recipe/meta.yaml) for conda as well.

### DIFF
--- a/.github/workflows/container_build.sh
+++ b/.github/workflows/container_build.sh
@@ -9,4 +9,4 @@ ${HOME}/miniconda3/bin/activate hexrd
 ${HOME}/miniconda3/bin/conda activate hexrd
 ${HOME}/miniconda3/bin/conda install conda-build
 mkdir output
-${HOME}/miniconda3/bin/conda build --output-folder output/ conda.recipe/
+${HOME}/miniconda3/bin/conda build -c defaults -c conda-forge --output-folder output/ conda.recipe/
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -57,7 +57,7 @@ jobs:
       run: |
           conda activate hexrd
           mkdir output
-          conda build --output-folder output/ conda.recipe/
+          conda build -c defaults -c conda-forge --output-folder output/ conda.recipe/
       # This is need to ensure ~/.profile or ~/.bashrc are used so the activate
       # command works.
       shell: bash -l {0}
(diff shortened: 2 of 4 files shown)


cjh1:
Yes lets add it to the conda recipe as well.

cjh1 (approved):

psavery:
Looks like it broke the conda package :slightly_frowning_face:

psavery:
This works in the command line, though:

```bash
conda install -c conda-forge lmfit==1.0.0
```

joelvbernier:
Hmmm.  We can probably find a replacement if we have to, right @saransh13 ?  Is this straight up Levenberg-Marquadt?

psavery:
@cjh1 Do we need to just get it to check `conda-forge` for the runtime dependencies as well?

cjh1:
I think [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html) has Levenberg-Marquadt if that is all that is needed. Might be nice to use it if its not too much trouble, as its already a dep. Otherwise, I can take a look at the packaging issue.

saransh13:
`lmfit` just wraps the `scipy` routines. However, theres a parameter class that `lmfit` introduces which is currently being used.

cjh1:
@saransh13 Is [this](https://github.com/lmfit/lmfit-py/blob/master/lmfit/parameter.py) the class you are talking about? Would be possible to call scipy directly?

saransh13:
@cjh1 yep thats the class. It should be possible, but will probably require a little time to update everything.
I've also have a PR which should eliminate the requirement for `1.0.0` and we can safely use the latest `1.0.1`.
This should resolve the broken conda package for the time being.

cjh1:
@saransh13 I think we need to make come changes to the packaging for this, I will take a look.

psavery:
@cjh1 Should we put the anaconda channel first to prefer anaconda over conda-forge?

The mac build failed because of something to do with libopenblas. But I think preferring anaconda will eliminate libopenblas usage anyways.

cjh1:
@psavery I think you might be right, I am looking at it now ...

cjh1:
Finally packaged! :fireworks:

cjh1 (approved):

psavery:
Once #108 gets merged, we can potentially remove the version number from lmfit.

cjh1:
@psavery I updated it to relax the version ...

PR #108, opened 2020-09-17
Lmfit fix

saransh13:
There was a change in one of the keywords to the `lmfit.Minimizer` class in release `1.0.1`. 
Fixing the keyword seems to produce identical results to `v1.0.0`. 
This should resolve the conda package being broken in PR #105

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -24,7 +24,7 @@ requirements:
   run:
     - python {{PY_VER}}
     - h5py
-    - lmfit ==1.0.0
+    - lmfit
     - numba
     - numpy
     - psutil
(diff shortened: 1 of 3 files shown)


cjh1:
@saransh13 I don't this this will help. However, I can look at fixed the packaging in #105

joelvbernier (commented):
Didn't you commit the material and symbols changes in #107 ?

cjh1 (approved):
This looks good to me, @saransh13 Looks like @joelvbernier has a question?

saransh13:
@joelvbernier Yep, I think so. This change just made `lmfit1.0.1`work with WPPF.

cjh1:
@saransh13 Let merge this  :+1:

PR #115, opened 2020-09-24
catch nans, zeros and negative values in WPPF

saransh13:
This catches the `nan` values, and `<= 0.0` inside WPPF and sets the statistical weights for these points to `0`.
This fixes some non-convergence issues in `WPPF` with the Tardis and ceria dataset.


cjh1 (approved):

psavery (approved):
I didn't check to see if the result was correct for the tardis example, but this at least fixes the error for me.

Issue #120, opened 2020-09-25
scattering factors for ions


saransh13:
scattering factors for ions are not present in `hexrd`

joelvbernier:
Just checking on this.  Where you at @saransh13 ?

joelvbernier:
@saransh13 I think this is closed, yeah?

PR #121, opened 2020-09-25
Ionic form factors to compute "f0"

saransh13:
adding form factor to compute `f0` for ionic elements. The way in which the full scattering calculation is done now relies on the atomic number to element symbol mapping which does not convey the information about the charge. Therefore it'll be hard to implement scattering from ionic species in the way things are done right now. That'll require a more substantial change will be undertaken sometime later in the future.

For now at least putting in the form factors. This will partially resolve issue #120.


joelvbernier:
Should this be a draft PR for now?  We'll have to modify the CIF key munging to detect ionic species, yes?

saransh13:
This just adds form factors for ions in addition to the data already present. Shouldn't mess with anything really.
Nothing is done with the data yet.

joelvbernier (approved):

PR #123, opened 2020-09-28
Don't modify angles for Angstrom -> nm conversion

psavery:
Previously, the angles were being divided by 10 in addition
to the lengths. The angles should not be modified for this conversion.


cjh1 (approved):
LGTM

PR #126, opened 2020-09-29
Use pathlib to get suffix

cjh1:
The logic was searching for the first dot. This does not work if the path has a dot in it for example a version number in the package.


cjh1:
@saransh13 This is needed to get things working in the package.

psavery (approved):
LGTM

PR #127, opened 2020-10-02
Add extensions directory

cjh1:
This fixes an issue with develop install on Windows, as the sglite
directory already exists.


cjh1:
@joelvbernier This should resolve the dev install issue.

joelvbernier (changes requested):
The idea of the extensions folder was to have a single place for all the compiled extenstions to go.  Also, sglite is going away, as the new materials package replaces it (and we will no longer need to build the sglite package).  Lastly, the transforms extension has no relation to sglite, and if anything can go into the transforms package folder.

cjh1:
OK, I will rework

joelvbernier (approved):

PR #131, opened 2020-10-10
Ported "applySym" function in symmetry module from python2 hexrd

rachelelim:
Needed to calculate Schmid tensors.


joelvbernier:
This method was moved to hexrd.rotations to fix a circular import.

PR #132, opened 2020-10-21
Check for empty presults before fitting attempt

psavery:
In some cases, presults may be empty for a specific detector. This would
lead to a numpy exception later when a vstack was attempted with an empty
list.

Check if presults is empty, and if so, set the culled results to empty as
well and continue.

@joelvbernier Is it reasonable for the `presults` variable to be empty? It looks like it is empty
sometimes for our 8-detector example, and that was causing the exception in https://github.com/HEXRD/hexrdgui/issues/643.

If it is not reasonable for `presults` to be empty, we may need to find a different fix.

Fixes: https://github.com/HEXRD/hexrdgui/issues/643

### DIFF
--- a/hexrd/fitgrains.py
+++ b/hexrd/fitgrains.py
@@ -228,6 +228,10 @@ def fit_grain_FF_reduced(grain_id):
         for det_key in culled_results_r:
             presults = culled_results[det_key]
 
+            if not presults:
+                culled_results_r[det_key] = []
+                continue
+
             ims = imgser_dict[det_key]
             ome_step = sum(np.r_[-1, 1]*ims.metadata['omega'][0, :])
 


cjh1:
@saransh13 Any chance you can review this and answer @psavery question. It would be good to get this in.

saransh13:
I haven't worked too much on the fit grains program so I would defer to @joelvbernier on this.

psavery:
The filesystem is being used here currently: https://github.com/HEXRD/hexrd/blob/2ee9f3bcb42ed4328b930dd68acf65da7e127125/hexrd/fitgrains.py#L155-L161

joelvbernier:
Yeah @psavery -- that is in the case of an overlap table only (currently not implemented in the standard workflow).  Looks like the `presults` are coming from the `results` dict in memory, **not** the filesystem, so I think your fix is appropriate (_i.e._, empty really means empty!): https://github.com/HEXRD/hexrd/blob/2ee9f3bcb42ed4328b930dd68acf65da7e127125/hexrd/fitgrains.py#L100-L111

joelvbernier (approved):

psavery:
Sounds great, thanks Joel!

Issue #133, opened 2020-10-30
Compilation error


saransh13:
There is some error in the `hexrd` installation. Not sure where this is originating. The installation goes fine on my Mac.

`gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -I/usr/local/opt/sqlite/include -I/usr/local/opt/sqlite/include -I/Users/runner/hostedtoolcache/Python/3.8.6/x64/include/python3.8 -c src/lib/lex.yy.c -o build/temp.macosx-10.14-x86_64-3.8/src/lib/lex.yy.o
133
    star.l:111:28: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
134
                     while ( p = strchr( p, '\''))
135
                             ~~^~~~~~~~~~~~~~~~~~
136
    star.l:111:28: note: place parentheses around the assignment to silence this warning
137
                     while ( p = strchr( p, '\''))
138
                               ^
139
                             (                   )
140
    star.l:111:28: note: use '==' to turn this assignment into an equality comparison
141
                     while ( p = strchr( p, '\''))
142
                               ^
143
                               ==
144
    star.l:136:28: warning: using the result of an assignment as a condition without parentheses [-Wparentheses]
145
                     while ( p = strchr( p, '\"'))
146
                             ~~^~~~~~~~~~~~~~~~~~
147
    star.l:136:28: note: place parentheses around the assignment to silence this warning
148
                     while ( p = strchr( p, '\"'))
149
                               ^
150
                             (                   )
151
    star.l:136:28: note: use '==' to turn this assignment into an equality comparison
152
                     while ( p = strchr( p, '\"'))
153
                               ^
154
                               ==
155
    star.l:117:24: warning: comparison of integers of different signs: 'int' and 'yy_size_t' (aka 'unsigned long') [-Wsign-compare]
156
                           yyless( i );
157
                           ^~~~~~~~~~~
158
    lex.yy.c:205:9: note: expanded from macro 'yyless'
159
            YY_LESS_LINENO(yyless_macro_arg);\
160
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
161
    lex.yy.c:187:36: note: expanded from macro 'YY_LESS_LINENO'
162
                    for ( yyl = n; yyl < yyleng; ++yyl )\
163
                                   ~~~ ^ ~~~~~~
164
[...]

PR #134, opened 2020-11-04
Switch to XCode 11.7

cjh1:
pycifrw will not build with the latest version of XCode so downgrade to 11.7.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,12 @@ jobs:
       run:
         shell: bash
     steps:
+    - name: Use XCode 11.7 so pycifrw will build
+      if: ${{ matrix.config.name == 'MacOSX' }}
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: '11.7'
+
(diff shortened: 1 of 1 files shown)


cjh1:
@saransh13 This fixes #133. The problem was caused by [this upgrade](https://github.com/actions/virtual-environments/issues/1712). There is update upstream that fixes the package to build with the latest XCode, but its not in a release yet. I have requested the maintainer pushes a new release. For now this will get the CI back passing.

.github/workflows/test.yml
```diff
@@ -31,6 +31,12 @@ jobs:
       run:
         shell: bash
     steps:
+    - name: Use XCode 12.0 so pycifrw will build
```
psavery, line 34: Should this be "11.7" instead of "12.0"?

cjh1: Indeed, I tried 12.0 first :-)

cjh1: Fixed

psavery (approved):
LGTM

saransh13:
@psavery Thanks! Merging this.

PR #137, opened 2020-11-11
Density computation

saransh13:
Fix density computation and make it a property such that it is recalculated when space group or lattice parameters are changed.


hexrd/unitcell.py
```diff
@@ -1258,6 +1314,8 @@ def a(self, val):
         self.ik = 1
         self.il = 1
         self.CalcMaxGIndex()
+        if(hasattr(self, 'numat')):
```
psavery, line 1317: You don't actually need these outer parentheses in `if` statements in python. Most people leave them out.

psavery (approved):
LGTM

PR #140, opened 2020-12-10
IPF fcolor

saransh13:
This PR adds the functionality to color direction vectors based on either the point or the laue group of the crystal. Since x-rays can't really tell these apart, the laue group coloring is switched on by default. A few new files are being added and a couple of new functions have been added to `unitcell.py`. 

the colors are returned as an array or `nx3`, where n is the number of orientations to color and the second dimension has the rgb color information. The call is as follows
`unitcell.color_orientations(rmats, ref_dir)`
`rmats` is the `nx3x3` array of rotation matrices and `ref_dir` is the reference direction set to `[0., 0., 1.]` by default.

NOTE: the current implementation is not set up for very efficient computations and there is most likely going to be some improvements made in the future.
NOTE: there is a known bug with point group `mm2` which returns incorrect colors. Since this is a very rare case, it'll be patched later.


cjh1:
@saransh13 Looks like we have a few conflicts to resolve

saransh13:
@cjh1 I think these should be resolved now.

saransh13:
Looks like some issue with the conda package on linux though.

cjh1:
@saransh13 I will take a look at the CI

cjh1:
@saransh13 I merged in master as the CI is now fixed there.

saransh13:
Thanks! Just need @joelvbernier to review and merge now.

hexrd/constants.py
```diff
@@ -71,6 +71,35 @@
...
+# this is a constant which defines the sign of the 
+# cross-product in the quaternion multiplication rule
+# we will set it to 1 for the standard rule always
+pjik  = 1
+
+# sqrt 2 - 1
+tp_8  = 0.41421356237309504880168872420969807856967187537694807317667973799073248
+
+# 2 - sqrt(3)
+tp_12 = 0.26794919243112270647255365849412763305719474618961937194419302054806698
```
psavery, line 101: So, the python float type can only store ~16 decimal places of precision. Thus, I don't think there need to be this many decimal places in the file.

It might also be easier to read to just store these as:
```python
tp_8 = np.sqrt(2) - 1
tp_12 = 2 - np.sqrt(3)
```

You can maybe use `numpy.float128` if you want more decimal places as well, or the `decimal` module.

hexrd/ipfcolor/colorspace.py
```diff
@@ -0,0 +1,165 @@
...
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the IMPLIED WARRANTY OF MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE. See the terms and conditions of the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this program (see file LICENSE); if not, write to
+# the Free Software Foundation, Inc., 59 Temple Place, Suite 330,
+# Boston, MA 02111-1307 USA or visit <http://www.gnu.org/licenses/>.
+# =============================================================================
```
psavery, line 27: FYI, I think hexrd is under a 3-clause BSD license now, instead of LGPL 2.1, although I do see that this header is in some of the other files too...

Is this right, @cjh1 ?

cjh1: That is correct: https://github.com/HEXRD/hexrd/blob/master/LICENSE

hexrd/constants.py
joelvbernier: Yes -- let's fix this :-)

hexrd/ipfcolor/colorspace.py
joelvbernier: Actually, yes it is.  I have been meaning to implement some sort of 'headerizing' function, because LLNL wants all the files to have a header.  It does need updating.

psavery:
I think it looks good to me after https://github.com/HEXRD/hexrd/pull/140#discussion_r543730899 is resolved.

joelvbernier (approved):

joelvbernier (approved):

joelvbernier (approved):

PR #141, opened 2020-12-14
Store and retrieve compliance tensor in TPa⁻¹

psavery:
The stiffness is in GPa. Previously, the compliance was in GPa⁻¹.
Now, store and retrieve the compliance in TPa⁻¹ instead.

This can be done in hexrdgui instead, if you prefer to leave
compliance in GPa⁻¹ in hexrd.

Please double-check my math...


saransh13 (approved):
Math looks good to me.

cjh1 (approved):

PR #142, opened 2020-12-14
Move package container to CentOS 7

cjh1:
CentOS 6 has reached the end of its life!


psavery (approved):
LGTM, if the tests pass

Issue #136, opened 2020-11-09
Save and load stiffness tensor


psavery:
When [loading a material from a file](https://github.com/HEXRD/hexrd/blob/0c3c2a76d2a72126273430eb0db76837bdebd8c4/hexrd/material.py#L411-L465), check for and read in the stiffness tensor.

When [writing a material to a file](https://github.com/HEXRD/hexrd/blob/0c3c2a76d2a72126273430eb0db76837bdebd8c4/hexrd/material.py#L467-L492), save the stiffness tensor.

These are needed for hexrd/hexrdgui#655

joelvbernier:
@saransh13 -- this looks to be a fairly high priority; can you knock this out by the end of the week?

saransh13:
Ok, on it. Will make a PR soon.

psavery:
@saransh13 Is this ready to close?

saransh13:
Yep, this can be closed.

PR #144, opened 2021-01-07
I/O of stiffness matrix

saransh13:
PR for issue #136 

I/O of stifness or compliance matrix from materials file. Its a conditional read depending on the presence or absence of the field. If the material file does not have this field, then it is filled with zeros.

The `Material.dump_material` function always writes a stiffness field now, even if it's all zeros.

The default material file with values of stiffness and compliance will be updated in a separate PR.


psavery (approved):
I skimmed through the code, and gave it a try, and it works for me.

LGTM

joelvbernier (approved):

PR #148, opened 2021-01-12
updates to HEDM CLI material handling

joelvbernier:
The HEDM CLI was still using the old material class (which was serialized using pickles); this has been changed such that the new material.Material class is used.  In the process, some additional options are available in the material block of the config:
- dmin: hook to specify the minimum d-spacing to generate the material's reflection list up to
- tth_width: hook to specify the tth tolerance (in degrees) on the active plane_data object (_i.e._, `PlaneData.tThWIdth`)
- min_sfac_ratio: hook to specify the minimum |F|^2 ratio to set exclusions in `PlaneData`
Also included a fix to the wavelength conversion in `crystallography.PlaneData` that was causing a loss of precision.


joelvbernier:
CI tests need fixing but tested by hand successfully.

PR #150, opened 2021-01-13
update hkl list properly once space group number is updated.

saransh13:
Found a bug that space group needed to be updated twice to update hkl list. this should fix that bug.


hexrd/material.py
```diff
@@ -525,6 +525,9 @@ def _set_sgnum(self, v):
         # Update the unit cell if there is one
         if hasattr(self, 'unitcell'):
             self.unitcell.sgnum = v
+            self.unitcell.getHKLs(self.dmin.getVal('nm'))
+            self._newPdata()
```
psavery, line 529: It looks like `self._newPdata()` calls `self.unitcell.getHKLs(self._dmin.getVal('nm'))` as well.

Do we need the `getHKLs()` here?

saransh13: Oh right, yes good catch. We can remove that.

joelvbernier (approved):

PR #152, opened 2021-01-19
Looks like the conda-build issue is resolve in 3.21.4

cjh1:
So move back to latest. See https://github.com/conda/conda-build/issues/4182 for more details.


psavery (approved):
LGTM

PR #153, opened 2021-01-19
Wppf h5 dump

saransh13:
write out all the info to hdf5 file. The groups created are `Parameters`, `Phases` and `Spectrum`. This just adds the write capabilities. The initialization of the `LeBail` class from this hdf5 will be part of a later commit.


joelvbernier:
You've tested this, yes?

joelvbernier (approved):

PR #157, opened 2021-01-22
add distortion function for dexela 2923

joelvbernier:
Simple in-plane distortion function for the dexela detectors.  Includes a simple unit test, but will need to integrate it in CI later.


joelvbernier:
Tests are failing because I still haven't fixed the materials HKL selection discrepancies, but I have tested this.

PR #158, opened 2021-01-22
Powder simulator

saransh13:
Added powder simulator to instrument class. Hooks to `WPPF.LeBail` are used to simulate. 
Required argument is list of `Material` objects. 
The parameters for the simulation can be changed using the optional `params` keyword. If this is not provided, then some sane default values are chosen.
The background is specified by an array or a filename with two rows of 2theta and background intensity. Finally, optional noise can be added via the `noise` keyword. Options to add `poisson`, `gaussian`, `salt and pepper` noise etc. have been added. Tested with MEC configuration and Zirconium crystal

### DIFF
--- a/hexrd/WPPF.py
+++ b/hexrd/WPPF.py
@@ -114,7 +114,8 @@ class Parameters:
             fid = file
 
         else:
-            raise RuntimeError('Parameters: dump_hdf5 Pass in a filename string or h5py.File object')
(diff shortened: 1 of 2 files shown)


saransh13:
One of the fit_grains test is failing. Something to do with the materials I think.

joelvbernier (approved):

PR #174, opened 2021-02-09
Set numba cache dir on Windows to avoid freeze

psavery:
If the following are true, set the numba cache directory:

1. We are using numba
2. We are on Windows
3. We don't have write permissions to constants.py (implies
   we won't have write permissions to __pycache__ dirs as well).
4. The numba cache dir environment variable isn't already set

This avoids a bug on Windows where, if numba tries to create a tempfile
in a directory where it doesn't have permission, the application will freeze.

This bug is documented in these two places:

https://stackoverflow.com/questions/55109076/python-tempfile-temporaryfile-hangs-on-windows-when-no-write-privilege

https://bugs.python.org/issue22107


psavery:
Only whitespace was removed from the lines pep8speaks is complaining about. I think we can ignore it.

cjh1 (approved):
LGTM, thanks for getting to bottom of this.

psavery:
@cjh1 Looks like this is causing some kind of infinite import error on Mac.

cjh1:
@psavery Lucky we have the CI!

PR #175, opened 2021-02-09
set exclusion from hkl list in material HDF5 file.

### LINKED ISSUE #172: Need mechanism for saving PlaneData exclusions to new material
In analyzing HEDM data from the most recent APS run, I am running into the need to save the exclusions (or "valid hkls" if you will) for the reflections lists on a material.  Currently, the objects are instantiated from the data stored in the hdf5 file upon load, and there is no mechanism for specifying a user-specified list of reflections to use.  For the most generality, we can probably store a list of {hkl} that are selected as valid, and have the material associate those with its master list upon generation.  That way we can handle the case if parameters like dmin are changed, which would change the total number of reflections generated.  In the HEDM workflow, the structure factor ratio and tThMax are still applied to mask the reflection list on the PlaneData object; the saved exclusions would be applied before these.

saransh13:
@joelvbernier This should fix #172. The mechanism is exactly the same as you commented. The initialization reads the hkl list from HDF5 file and sets exclusions based on which reflections are present there.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -108,7 +108,7 @@ class Material(object):
     the dmin parameter is used to figure out the maximum sampling for g-vectors
     this parameter is in angstroms
     '''
-    DFLT_DMIN = _angstroms(0.5)
+    DFLT_DMIN = _angstroms(0.75)
 
     '''
     some materials have more than one space group setting. for ex
@@ -466,8 +466,8 @@ class Material(object):
 
         gid = fid.get(xtal)
 
-        sgnum = numpy.asscalar(numpy.array(gid.get('SpaceGroupNumber'),
-                                           dtype=numpy.int32))
+        sgnum = numpy.array(gid.get('SpaceGroupNumber'),
+                dtype=numpy.int32).item()
         """
             IMPORTANT NOTE:
             note that the latice parameters is nm by default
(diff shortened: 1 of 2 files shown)


hexrd/material.py
```diff
@@ -475,6 +495,16 @@ def _readHDFxtal(self, fhdf=DFLT_NAME, xtal=DFLT_NAME):
         else:
             self.stiffness = numpy.zeros([6, 6])
 
+        if('dmin' in gid):
+            # if dmin is present in the HDF5 file, then use that
+            dmin = numpy.asscalar(numpy.array(gid.get('dmin'), \
+                                    dtype = numpy.float64))
```
psavery: `numpy.asscalar()` is [deprecated since version 1.16](https://numpy.org/doc/stable/reference/generated/numpy.asscalar.html). Looks like `.item()` is recommended instead:
```python
numpy.array(gid['dmin'], dtype=numpy.float64).item()
```

hexrd/material.py
```diff
@@ -108,7 +108,7 @@ class Material(object):
     the dmin parameter is used to figure out the maximum sampling for g-vectors
     this parameter is in angstroms
     '''
-    DFLT_DMIN = _angstroms(1.0)
+    DFLT_DMIN = _angstroms(0.5)
```
psavery, line 111: Remember that in hexrdgui, we are currently initializing all of the materials' plane data at program start-up. This is tracked in hexrd/hexrdgui#497.

Reducing the default dmin to 0.5 might slow start-up time.

psavery (commented):
This worked for me in my quick test of it in the GUI. But start-up time for the GUI is now slower due to the default dmin being smaller.

psavery:
Apparently, I get an email notification if I am requested to review something, but no email notification if that request is removed.

saransh13:
Oh sorry about that patrick. Must have accidentally unchecked you.

Joel and I talked about reducing the dmin because we had to do that later
in the GUI for most cases. But maybe the startup time is a good reason why
we shouldn’t do that. Maybe a value of 0.75 might be a happy medium.

Also will switch to items instead of asscalar. Thanks!

psavery:
Okay, we can also prioritize HEXRD/hexrdgui#497 if it is now more important.

psavery (approved):
LGTM

joelvbernier (commented):
looks all good to me.

PR #177, opened 2021-02-12
nan handling in astropy convolve

joelvbernier:
necessary for properly working with the GUI's version of PolarView that includes NaN padding around the detectors.


psavery (approved):
LGTM

Issue #180, opened 2021-02-18
Consider adding pyproject.toml file


psavery:
We can use pyproject.toml to specify which dependencies need to be installed before `setup.py` is ran. It could look like this, for instance:

```toml
[build-system]
requires = ["setuptools", "wheel", "numpy"]
```

This would reduce the development install instructions from this:
```python
pip install numpy
pip install -e hexrd
```

to this:
```python
pip install -e hexrd
```

Thoughts? @cjh1 @joelvbernier

joelvbernier:
I am game for _ANYTHING_ that helps users get going!

joelvbernier:
Can this be done in a platform-specific manner?  We *_NEED_* this

psavery:
@joelvbernier Yep! I think this should work for all platforms. I'll put together a PR...

PR #181, opened 2021-02-19
Add pyproject.toml for easier development installs

psavery:
The pyproject.toml file can be used to specify dependencies needed
for pip to run setup.py. Pip will create a virtual environment and
install these dependencies, then run setup.py.

This simplifies the development setup process so that instead of:
```python
pip install numpy
pip install -e hexrd
```

The developer can simply do:
```python
pip install -e hexrd
```

We need to include numpy in "install_reqs" now as well because
dependencies specified in pyproject.toml are only installed in the
virtual environment used to run setup.py, and numpy also needs to
be installed for use at runtime.

Fixes: #180

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,7 +59,6 @@ jobs:
 
     - name: Install HEXRD
       run: |
-          pip install numpy
           pip install .
       working-directory: hexrd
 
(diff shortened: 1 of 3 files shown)


cjh1 (approved):

joelvbernier:
Can one of you (@psavery or @cjh1) update the README install instructions on the HEXRD repo?

cjh1:
I will take care of that.

PR #185, opened 2021-02-25
adding default stifness. All zeros.

saransh13:
As explained in title.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -113,7 +113,7 @@ class Material(object):
     '''
     default stiffness tensor in voight notation
     '''
-    DFLT_STIFFNESS = numpy.zeros([6,6])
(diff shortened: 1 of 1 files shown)


joelvbernier (approved):

hexrd/material.py
```diff
@@ -175,7 +180,10 @@ def __init__(self, name=None,
             self._dmin.getVal('nm'), self._beamEnergy.value,
             self._sgsetting)
 
-        self.unitcell.stiffness = self.stiffness
+        if hasattr(self, 'stifness'):
```
psavery, line 183: `stifness` here should be `stiffness`

psavery: @saransh13

saransh13: I always spell it wrong :-/

PR #187, opened 2021-02-26
added eta_ome filtering

joelvbernier:
Added map filtering to address artifacts I was seeing in recent data (details to follow).

### DIFF
--- a/hexrd/config/findorientations.py
+++ b/hexrd/config/findorientations.py
@@ -112,7 +112,7 @@ class OmegaConfig(Config):
 
     @property
     def period(self):
-        # ??? maybe should get from image_series like before in v0.3.x
+        # FIXME: this is deprecated and now set from the imageseries
         key = 'find_orientations:omega:period'
         temp = self._cfg.get(key, [-180., 180])
         range = np.abs(temp[1]-temp[0])
@@ -235,3 +235,10 @@ class OrientationMapsConfig(Config):
     @property
     def threshold(self):
         return self._cfg.get('find_orientations:orientation_maps:threshold')
+
+    @property
+    def filter_maps(self):
+        return self._cfg.get('find_orientations:orientation_maps:filter_maps',
+                             default=False)
+
+
--- a/hexrd/findorientations.py
+++ b/hexrd/findorientations.py
@@ -12,6 +12,7 @@ import scipy.cluster as cluster
 from scipy import ndimage
 
 from skimage.feature import blob_dog, blob_log
+from skimage.exposure import rescale_intensity
 
 from hexrd import constants as const
 from hexrd import matrixutil as mutil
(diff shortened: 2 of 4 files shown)


hexrd/findorientations.py
```diff
@@ -552,6 +591,47 @@ def generate_eta_ome_maps(cfg, hkls=None, save=True):
...
+
+        # filter
+        # !!! False/None: only row median subtraction
+        #     True: use default
+        #     scalar: stdev for filter in pixels
+        # ??? simplify this behavior
+        if filter_stdev:
+            if isinstance(filter_stdev, bool):
+                filter_stdev = filter_stdev_DFLT
+            pf = -gl_filter(pf, filter_stdev)
```
psavery, line 628: It doesn't matter very much, but you might be able to modify `pf` in place via
`pf[:] = -gl_filter(pf, filter_stdev)`

hexrd/config/findorientations.py
```diff
@@ -235,3 +235,10 @@ def file(self):
     @property
     def threshold(self):
         return self._cfg.get('find_orientations:orientation_maps:threshold')
+
+    @property
+    def filter_maps(self):
+        return self._cfg.get('find_orientations:orientation_maps:filter_maps',
+                             default=False)
```
psavery, line 242: We should add this as an option in hexdgui I am guessing?

hexrd/findorientations.py
```diff
@@ -57,35 +59,40 @@ def _process_omegas(omegaimageseries_dict):
 
 
 def label_spots(this_map, method, method_kwargs):
-    # !!! need to remove NaNs from map in case of eta gaps?
-    this_map[np.isnan(this_map)] = 0.
+    # !!! need to remove NaNs from map in case of eta gaps
+    # !!! doing offset and truncation by median value now
+    nan_mask = np.isnan(this_map)
+    med_val = np.median(this_map[~nan_mask])
+    this_map[nan_mask] = med_val
```
psavery, line 66: Do you know if we ever confirmed that `this_map` here is a deep copy, so these modifications aren't getting pushed on to the indexing?

hexrd/config/findorientations.py
joelvbernier: yeah, we should update the widget.  In the filtering there are two options:

1. row-wise median filtering to get rid of the streaking artifacts; and
2. the above plus filtering with `gauss_laplace` with a user-specified stdev

hexrd/findorientations.py
joelvbernier: I want the changes to persist to the indexing.

hexrd/findorientations.py
joelvbernier: We can probably use the `output` kwarg in `gauss_laplace` 

https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.filters.gaussian_laplace.html#scipy.ndimage.filters.gaussian_laplace

joelvbernier: have you tried the `[:]` syntax?

psavery: > have you tried the `[:]` syntax?

Yeah, it works.

psavery (approved):
LGTM

PR #189, opened 2021-02-26
Near field code with MPI

cjh1:


hexrd/xrdutil/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .utils import *
+
+# For now import these private members!
+from .utils import _project_on_detector_plane
+from .utils import _fetch_hkls_from_planedata
+from .utils import _filter_hkls_eta_ome
+from .utils import _fetch_hkls_from_planedata
```
psavery, line 7: Looks like `_fetch_hkls_from_planedata` is imported twice!

hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
```diff
@@ -0,0 +1,1132 @@
...
+                                              'grand-loop-experiment.gz'),
+                                 compress=True)
+            id_state = joblib.dump(state[:-1],
+                                   os.path.join(tmp_dir, 'grand-loop-data'))
+            pool = multiprocessing.Pool(ncpus, worker_init,
+                                        (id_state[0], id_exp[0]))
+            yield pool
+        finally:
+            logging.info('Deleting "%s".', tmp_dir)
+            shutil.rmtree(tmp_dir)
```
psavery, line 1020: @joelvbernier It looks to me that starting in python 3.8, they deemed the fork method unsafe on mac. It now defaults to using spawn, which might be why only `--force-spawn-multiprocessing` works for you. You may have to call [`multiprocessing.set_start_method('fork')`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method) to get Mac to fork, but even then, since it is considered unsafe, it might not work.

From [here](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods): "Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess."

hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
```diff
@@ -0,0 +1,1132 @@
...
+# This part is critical for the performance of simulate diffractions. It
+# basically "renders" the "pixels". It takes the coordinates, quantizes to an
+# image coordinate and writes to the appropriate image in the stack. Note
+# that it also performs clipping based on inv_deltas and clip_vals.
+#
+# Note: This could be easily modified so that instead of using an array of
+#       booleans, an array of uint8 could be used so the image is stored
+#       with a bit per pixel.
+
+@numba.njit
```
psavery, line 660: For these numba functions, we might be able to get a little bit of a speed-up from using `cache=True`, like [here](https://github.com/HEXRD/hexrd/blob/0a0886e84bee8094e0a544e181fbaae25a928f76/hexrd/matrixutil.py#L855).

The issues with caching on Windows in hexrd should all be resolved. We can benchmark and test it though.

hexrd/xrdutil/__init__.py
cjh1: Fixed up

hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
cjh1: I am guess at the end of the day if we are targeting HPC, Linux will be the target platform.

hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
joelvbernier: good point

psavery:
I just merged in #199 into this branch. Here are the notes I left on that PR:

This appears to be running very effectively.

On the cluster we are testing this on, we have access to 4 nodes, each of which have 2 sockets, and each socket has 14 physical cores and 28 logical cores (so each node has 28 physical cores and 56 logical cores).

Timings below were done after ensuring numba caches were generated.

For running multiprocessing on a single node (using both sockets), the grand loop takes approximately 10.7 seconds to complete with the test.

For running one MPI process per socket across the nodes (8 sockets in total) where each MPI process uses 28 CPUS (utilizing the hyperthreading), the grand loop takes approximately 2.6 seconds to complete with the test. The amount of time the MPI communication takes at the end varies a lot, but it usually is somewhere between 0.05 seconds and 1 second. Utilizing all 28 logical cores per socket was found to be faster than utilizing only the 14 physical cores per socket (3 seconds instead of 2.6 seconds).

Thus, the MPI on 4 nodes is almost exactly 4x faster than multiprocessing on one node.

conda.recipe/meta.yaml
```diff
@@ -34,6 +34,7 @@ requirements:
     - scikit-image
     - scikit-learn
     - scipy
+    - progressbar
```
psavery, line 37: The progressbar doesn't have very pretty output when running this on the cluster.

I made it optional in the near field script. Maybe we should axe the dependency, and it will only be used if it is installed?

joelvbernier (approved):

georgejoe018:
I do not want to interrupt the flow that you all have created. My name is Joseph, and I am a student at the Colorado School of Mines. I am trying to run the hexrd program that you all have worked on. I was wondering if you could give me some guidance as I am having issues reading files of data and having the options to select multiple detectors. Hopefully you can help me at all and thank you for allowing my intrusion.

joelvbernier:
@georgejoe018 -- I need more information about what workflow you are trying to run before I can help.  This thread is for the new NF workflow, which is still a work in progress.

georgejoe018:
@joelvbernier -- I am working with the linux version for FF workflow. NF analysis is not necessarily needed at the moment. I can move this conversation to a different thread for issues with FF workflow if need be. The data I am trying to upload is .ge3 and .h5 files. I can get one detector graph to load for .h5 files, but not both. I cannot load in any .ge3 files. I am running hexrd 0.8 as well. Hopefully this information is helpful.

joelvbernier:
@georgejoe018 -- yes, this is not the proper thread for your questions.  I can say that we are regularly using ge and hdf5 files, so without more information, I won't be able to help you.  Please start a [new issue](https://github.com/HEXRD/hexrd/issues/new/choose) describing your problem in detail and I will try to help you out.

PR #190, opened 2021-03-02
Move code to filter maps if requested to function

psavery:
I'm mostly just doing this so we can call the function in hexrdgui...


psavery:
@joelvbernier I just moved the filtering outside of the `load_eta_ome_maps()` functions as well, so that we can
load the eta omega maps in the GUI without filtering, and then apply filtering later when the user selects the
options.

Is this fine?

joelvbernier (approved):
Looks fine to me.

PR #191, opened 2021-03-03
WIP: Naive MPI implementation using mpi4py.futures

cjh1:
This is a super simple naive implementation using `mpi4py.futures`. We can basically use `MPIPoolExecutor` as drop in replacement for the `multiprocessing` pool, then we have our worker pool distributed over our ranks. The is obviously some communication overhead, so my test on a single node run slower than the `multiprocessing` version, but it would be good to test on a real cluster and there are definitely improvements we can make, but it super simple. It uses MPI's dynamic process management features, introduced in MPI-2, works with the MPICH version I have locally, it also possible to run in a legacy mode if we don't have these features:

```
mpiexec -n 1 -usize 20 python nf-HEDM_test.py 
```

### DIFF
--- a/hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
+++ b/hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
@@ -31,6 +31,17 @@ from skimage.morphology import dilation as ski_dilation
 
 from progressbar import ProgressBar, Percentage, Bar
 
+USE_MPI = False
+try:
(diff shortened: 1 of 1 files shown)


cjh1:
@psavery Here is my experiment with `mpi4py.futures`, very few changes, which it pretty nice.

psavery:
I'm closing this in favor of #199, which seems to be very effective at this point.

PR #194, opened 2021-03-04
return structure factors filtered by exclusions.

saransh13:
The structure factors were not getting filtered by exclusions after the exclusions were updated. This fixes it.
The `planeData.get_structFact()` subroutine now returns `planeData.__structFact[~planeData.exclusions]` in place of just `planeData.__structFact` .

### DIFF
--- a/hexrd/crystallography.py
+++ b/hexrd/crystallography.py
@@ -787,7 +787,7 @@ class PlaneData(object):
     wavelength = property(get_wavelength, set_wavelength, None)
 
     def get_structFact(self):
-        return self.__structFact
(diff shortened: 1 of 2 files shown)


psavery (approved):
LGTM

psavery:
I think I was actually trying to bypass this bug [here](https://github.com/HEXRD/hexrdgui/blob/c4b906171a7cafdea6c9786ac8d963921a5f275f/hexrd/ui/materials_table.py#L116-L121). That might not be necessary anymore.

joelvbernier (approved):

PR #199, opened 2021-03-05
WIP: Simple MPI implementation using gather

cjh1:
This is a less naive implementation. That doesn't ship around a bunch of data. It basically partitions the coordinates across the ranks and then uses multiprocessing on each rank ( rather than multiple MPI processes per rank ). The image stack is loading and dilated on each rank to avoid have to ship it around. By making the MPI calls ourselves we can also take advantage of direct memory access to numpy arrays, so removes the overhead of pickling.

Note: If we go down this route we still need to work out what todo with the controller, currently progress monitoring is a little confusing with multiple ranks!

### DIFF
--- a/hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
+++ b/hexrd/xrdutil/nf-hedm/nf-HEDM_test.py
@@ -387,12 +387,12 @@ def mockup_experiment():
 # =============================================================================
 
 # Some basic 3d algebra =======================================================
-@numba.njit
+@numba.njit(nogil=True, cache=True)
 def _v3_dot(a, b):
     return a[0]*b[0] + a[1]*b[1] + a[2]*b[2]
 
 
-@numba.njit
+@numba.njit(nogil=True, cache=True)
 def _m33_v3_multiply(m, v, dst):
     v0 = v[0]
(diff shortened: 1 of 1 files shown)


psavery (approved):
This appears to be running very effectively.

On the cluster we are testing this on, we have access to 4 nodes, each of which have 2 sockets, and each socket has 14 physical cores and 28 logical cores (so each node has 28 physical cores and 56 logical cores).

Timings below were done after ensuring numba caches were generated.

For running multiprocessing on a single node (using both sockets), the grand loop takes approximately 10.7 seconds to complete with the test.

For running one MPI process per socket across the nodes (8 sockets in total) where each MPI process uses 28 CPUS (utilizing the hyperthreading), the grand loop takes approximately 2.6 seconds to complete with the test. The amount of time the MPI communication takes at the end varies a lot, but it usually is somewhere between 0.05 seconds and 1 second. Utilizing all 28 logical cores per socket was found to be faster than utilizing only the 14 physical cores per socket (3 seconds instead of 2.6 seconds).

Thus, the MPI on 4 nodes is almost exactly 4x faster than multiprocessing on one node. 

I have also confirmed that the output is correct.

I think this is ready to merge into the near-field branch.

Issue #205, opened 2021-03-10
Modifying Material sgnum doesn't update affected lattice parameters


psavery:
This was mentioned first in https://github.com/HEXRD/hexrdgui/issues/767#issue-815874490

Note the following test:
```python
from hexrd.material import Material
  
mat = Material()

def print_params():
    print('lattice Params:', *[x.value for x in mat.latticeParameters])

print(f'{mat.sgnum=}')
print_params()

mat.sgnum = 165

print(f'{mat.sgnum=}')
print_params()
```

Output:
```
mat.sgnum=225
lattice Params: 3.61 3.61 3.61 90.0 90.0 90.0
mat.sgnum=165
lattice Params: 3.61 3.61 3.61 90.0 90.0 90.0
```

You can see that even though the space group number was changed to 165 (a trigonal lattice type), the
angles all still remain 90, when the third angle should be 120. It looks like lattice parameters are not
updating when the space group number changes, and it is also affecting later calculations.

I guess we need to add some update to the lattice parameters in the [sgnum setter?](https://github.com/HEXRD/hexrd/blob/8917d038aaeef9e24c5d3d335e458bd7cf30b4f1/hexrd/material.py#L588-L599).

saransh13:
Ok, I'm going to push a fix for this. Thanks for catching this @psavery

joelvbernier:
I thought we had an issue for this already @saransh13?

joelvbernier:
Yes, it is on hexrdgui: https://github.com/HEXRD/hexrdgui/issues/767

joelvbernier:
Is there a way to link issues across repos @psavery or @cjh1?

joelvbernier:
Adding a ref to this issue in the assoc hexrdgui one for now (comments).

cjh1:
Yes, that is the closes you can do.

joelvbernier:
added comments to HEXRD/hexrdgui#767

saransh13:
Ok I think I've figured this one out. PR coming soon.

psavery:
This was fixed by #207

PR #207, opened 2021-03-11
Update lattice parameter when space group is updated.

saransh13:
fixing the update of lattice parameter to correspond to update of space group number. 
Fix for issue #205

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -605,7 +605,11 @@ class Material(object):
 
     @property
     def vol(self):
-        return self.unitcell.vol
+        return self.unitcell.vol*1e3
+
+    @property
+    def vol_per_atom(self):
+        return self.unitcell.vol_per_atom
 
     # property:  sgnum
 
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -38,6 +38,7 @@ from hexrd.crystallography import PlaneData as PData
 from hexrd.valunits import valWUnit
 from hexrd import unitcell
 from hexrd.constants import ptable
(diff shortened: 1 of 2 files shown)


psavery:
@saransh13 So I guess it was easier to generate a new unit cell from scratch rather than updating the lattice parameters on the unit cell?

hexrd/material.py
```diff
@@ -199,6 +191,34 @@ def __str__(self):
...
+        ltype = symmetry.latticeType(self.sgnum)
+        lparms = unitcell._rqpDict[ltype][1](lparms)
+        lparms_vu = []
+        for i in range(6):
+            if(i < 3):
+                lparms_vu.append(_angstroms(lparms[i]))
+            else:
+                lparms_vu.append(_degrees(lparms[i]))
+
+        self._lparms = lparms_vu
```
psavery, line 211: It might be nice to put this section into a separate function for clarity, such as `_reset_lparms()` or `_sgnum_changed()`.

joelvbernier: I second this.

saransh13: Gotcha!

saransh13: I can just update the lattice parameter and then set the sgnum since that is a property in unitcell and updates everything. But I think it'll be the same computational time as initializing a new `unitcell` since each function in the `__init__` is pretty much called again. Need to check that though.

psavery (approved):
LGTM

saransh13:
Thanks! Merging.

Issue #212, opened 2021-03-15
Ruby "c" is getting reset


psavery:
Using [materials.h5 from hexrdgui](https://github.com/HEXRD/hexrdgui/blob/cc3bb73c0f3a8a96d675da97a6cc53b1da5ff647/hexrd/ui/resources/materials/materials.h5).

```python
from hexrd.material import Material
  
mat = Material('ruby', 'materials.h5')
print([x.value for x in mat.latticeParameters])

# What it used to print: [4.7608, 4.7608, 12.99568, 90.0, 90.0, 120.0]
# What it now prints:    [4.7608, 4.7608, 4.7608, 90, 90, 120]
```

psavery:
@saransh13 Looks like it is getting modified [here](https://github.com/HEXRD/hexrd/blob/ba84bf71c888d021504db828e23be034440d06e5/hexrd/material.py#L202).

Here is an example:
```python
lparms = [4.7608, 4.7608, 12.99568, 90.0, 90.0, 120.0]
unitcell._rqpDict['trigonal'][1](lparms)
# (4.7608, 4.7608, 4.7608, 90, 90, 120)
```

PR #213, opened 2021-03-15
Material fix

saransh13:
fixing error introduced in last push to materials. A full 6 parameter list was being passed to `~rqpDict[self.unitcell.latticeType][1]` instead of the reduced set. This is a fix for issue #212 

Fixes: #212

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -780,11 +780,33 @@ The values have units attached, i.e. they are valWunit instances.
         else:
             print("Improper syntax, array must be n x 4")
 
+        self.update_structure_factor()
         return
 
     atominfo = property(
         _get_atominfo, _set_atominfo, None,
         "Information about atomic positions and electron number")
+
+    # # property: "atominfo"
+    # def _get_atomtype(self):
+    #     """Set method for name"""
+    #     return self._atomtype
+
+    # def _set_atomtype(self, v):
+    #     """Set method for atomtype"""
+    #     if v.shape[1] == 4:
+    #         self._atominfo = v
+    #         if hasattr(self, 'unitcell'):
+    #             self.unitcell.atom_pos = v
+    #     else:
+    #         print("Improper syntax, array must be n x 4")
+
+    #     self.update_structure_factor()
+    #     return
+
+    # atominfo = property(
+    #     _get_atomtype, _set_atomtype, None,
+    #     "Information about atomic positions and electron number")
 
     #
     #  ========== Methods
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -775,6 +775,8 @@ The values have units attached, i.e. they are valWunit instances.
         """Set method for name"""
         if v.shape[1] == 4:
             self._atominfo = v
+            if hasattr(self, 'unitcell'):
+                self.unitcell.atom_pos = v
         else:
             print("Improper syntax, array must be n x 4")
 
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -200,6 +200,7 @@ class Material(object):
         """
         lparms = [x.value for x in self._lparms]
         ltype = symmetry.latticeType(self.sgnum)
+        lparms = [lparms[i] for i in unitcell._rqpDict[ltype][0]]
         lparms = unitcell._rqpDict[ltype][1](lparms)
         lparms_vu = []
         for i in range(6):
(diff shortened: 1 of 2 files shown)


hexrd/unitcell.py
```diff
@@ -1480,9 +1479,10 @@ def Required_C(self, C):
 
     def MakeStiffnessMatrix(self, inp_Cvals):
         if(len(inp_Cvals) != len(_StiffnessDict[self._laueGroup][0])):
-            raise IOError('number of constants entered is not correct.\
-            need a total of ', len(_StiffnessDict[self._laueGroup][0]),
-                          'independent constants')
+            x = len(_StiffnessDict[self._laueGroup][0])
+            msg = f"number of constants entered is not correct."
+            "need a total of {x} independent constants"
```
psavery, line 1484: I think multi-line strings need to be in parentheses. And each f-string needs to start with "f", like:

```python
            msg = (f"number of constants entered is not correct. "
                   f"Need a total of {x} independent constants")
```

hexrd/unitcell.py
```diff
@@ -1982,17 +1982,42 @@ def atom_pos(self):
...
+        updating asymmetric positions after 
+        updating atominfo
+        fixing 
+        """
+        if hasattr(self, 'atom_type'):
+            if self.atom_ntype != val.shape[0]:
+                raise ValueError(f"incorrect number of atom positions."
+                                 "number of atom type = {self.atom_ntype} "
+                                 "and number of"
+                                 "atom positions = {val.shape[0]}.")
```
psavery, line 1996: Each string needs to start with "f" for it to be an f-string. So:

```python
                raise ValueError(f"incorrect number of atom positions. "
                                 f"number of atom type = {self.atom_ntype} "
                                 f"and number of "
                                 f"atom positions = {val.shape[0]}.")
```

Be sure to also add spaces at the end of each string so it won't come out like `"positions.number of"`

hexrd/unitcell.py
```diff
@@ -1480,8 +1480,8 @@ def Required_C(self, C):
     def MakeStiffnessMatrix(self, inp_Cvals):
         if(len(inp_Cvals) != len(_StiffnessDict[self._laueGroup][0])):
             x = len(_StiffnessDict[self._laueGroup][0])
-            msg = f"number of constants entered is not correct."
-            "need a total of {x} independent constants"
+            msg = (f"number of constants entered is not correct."
+                   "need a total of {x} independent constants")
```
psavery, line 1484: This string needs to start with `f`.

hexrd/unitcell.py
```diff
@@ -1990,10 +1990,11 @@ def atom_pos(self, val):
...
         if hasattr(self, 'atom_type'):
             if self.atom_ntype != val.shape[0]:
-                raise ValueError(f"incorrect number of atom positions."
-                                 "number of atom type = {self.atom_ntype} "
-                                 "and number of"
-                                 "atom positions = {val.shape[0]}.")
+                msg = (f"incorrect number of atom positions."
+                        "number of atom type = {self.atom_ntype} "
+                        "and number of"
+                        "atom positions = {val.shape[0]}.")
```
psavery, line 1996: These strings also need to start with `f`

hexrd/material.py
```diff
@@ -200,6 +200,7 @@ def _reset_lparms(self):
         """
         lparms = [x.value for x in self._lparms]
         ltype = symmetry.latticeType(self.sgnum)
+        lparms = [lparms[i] for i in unitcell._rqpDict[ltype][0]]
```
psavery, line 203: This is the important part of the PR, and it looks good to me.

hexrd/unitcell.py
```diff
@@ -1480,9 +1479,10 @@ def Required_C(self, C):
 
     def MakeStiffnessMatrix(self, inp_Cvals):
         if(len(inp_Cvals) != len(_StiffnessDict[self._laueGroup][0])):
-            raise IOError('number of constants entered is not correct.\
-            need a total of ', len(_StiffnessDict[self._laueGroup][0]),
-                          'independent constants')
+            x = len(_StiffnessDict[self._laueGroup][0])
+            msg = (f"number of constants entered is not correct."
```
psavery, line 1483: Technically, there is not going to be a space after this period when this error message gets printed out. But I guess it's not that big of a deal.

hexrd/unitcell.py
```diff
@@ -1982,17 +1982,43 @@ def atom_pos(self):
...
+        SS 03/08/2021 fixing some issues with
+        updating asymmetric positions after 
+        updating atominfo
+        fixing 
+        """
+        if hasattr(self, 'atom_type'):
+            if self.atom_ntype != val.shape[0]:
+                msg = (f"incorrect number of atom positions."
+                       f"number of atom type = {self.atom_ntype} "
+                       f"and number of"
```
psavery, line 1995: There won't be a space after "positions." or "number of" here either.

psavery (approved):
LGTM

PR #216, opened 2021-03-16
Perform multiprocessing on pixel_solid_angles

psavery:
Threading was attempted, but it appeared that all the threads would
either be fighting for the GIL or fighting for the compiler lock in
numba. Threading actually resulted in a large slowdown.

Multiprocessing seems to work well, since there is not a lot of
data to transmit back and forth. For 2048 rows by 2048 columns
and pixel sizes of 0.2, I was able to go from about 46 seconds
to about 6.8 seconds for running this property.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -1991,7 +1991,7 @@ class PlanarDetector(object):
 
     @property
     def row_edge_vec(self):
-        return self.pixel_size_row*(0.5*self.rows-np.arange(self.rows+1))
+        return _row_edge_vec(self.rows, self.pixel_size_row)
 
     @property
     def col_pixel_vec(self):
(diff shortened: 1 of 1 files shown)


psavery:
I gave this a try on a Windows computer with 4 cores, and it went from 42 seconds to 17.5 seconds.

psavery:
Looks like for 2048 x 2048 pixels, and 0.2 pixel sizes, this uses up about 0.4 GB of RAM per process on the computers I have tried it on: 1.6 GB of RAM on the 4-core Windows computer, and ~9 GB of RAM on the 24-core Linux computer. It did not overwhelm the memory in either one.

psavery:
This is ready for review.

joelvbernier (approved):

Issue #214, opened 2021-03-15
Unit testing of materials package


joelvbernier:
Given the critical functionality of the materials package, it is time that we add some basic unit tests to help prevent bad bugs from being merged into master.  For starters I suggest the following:

- [ ] for the reference materials, compare the data in the HDF5 to what gets instantiated in moth the `unitcell` and `PlaneData` classes, including
- cross-check of lattice parameters
- cross check of, say, the first 10 hkls (probably need some additional static data like and npz in `material.testing` for this)
- [ ] check for consistency when switching the default material to a different space group (for each of the other 6 non-cubic lattice types)

psavery:
If we add a script to the [tests directory](https://github.com/HEXRD/hexrd/tree/master/tests), it should be automatically included when github CI performs the testing. I can help you guys set that up if you'd like.

psavery:
So, if I'm recalling things correctly, pytest will import the scripts in that directory and automatically look for functions that start with `test_`. It assumes those are test functions, and it tries to run them.

There are some other things such as pytest fixtures that are used to make things easier.

psavery:
We are testing a whole lot of things currently, and I think most of this is covered. I think we can look to the code coverage to see things that are missing.

PR #220, opened 2021-03-16
Add some basic testing for the Material class

psavery:
This is just adding a little framework for testing the Material class,
through which we can add more tests.

For: #214

### DIFF
--- a/tests/test_material.py
+++ b/tests/test_material.py
@@ -1,0 +1,57 @@
+import pytest
+
+from hexrd.material import Material
+
+# Tolerance for comparing floats
+FLOAT_TOL = 1.e-8
+
+# Use consistent units to simplify testing
+DEFAULT_LENGTH_UNIT = 'angstrom'
+DEFAULT_ANGLE_UNIT = 'degrees'
+
+
+@pytest.fixture
+def default_material():
+    return Material()
+
+
+def normalize_unit(v):
+    if hasattr(v, 'unit'):
+        # Assume it's a val unit
+        if v.unit in ('radians', 'degrees'):
+            # Assume it's an angle
+            return v.getVal(DEFAULT_ANGLE_UNIT)
+        else:
+            # Assume it's a length
+            return v.getVal(DEFAULT_LENGTH_UNIT)
+
+    return v
+
+
+def are_close(v1, v2, tol=FLOAT_TOL):
+    return abs(normalize_unit(v1) - normalize_unit(v2)) < tol
+
+
+def lparms_are_close(lparms, indices, tol=FLOAT_TOL):
+    first = lparms[indices[0]]
+    return all(are_close(first, lparms[i], tol) for i in indices[1:])
+
+
+def test_sgnum_setter(default_material):
+    # Just in case we change the default...
+    default_material.sgnum = 225
+
+    # Should have a == b == c and alpha == beta == gamma == 90
+    lparms = default_material.latticeParameters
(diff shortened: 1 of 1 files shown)


tests/test_material.py
```diff
@@ -0,0 +1,57 @@
...
+def are_close(v1, v2, tol=FLOAT_TOL):
+    return abs(normalize_unit(v1) - normalize_unit(v2)) < tol
+
+
+def lparms_are_close(lparms, indices, tol=FLOAT_TOL):
+    first = lparms[indices[0]]
+    return all(are_close(first, lparms[i], tol) for i in indices[1:])
+
+
+def test_sgnum_setter(default_material):
```
psavery, line 40: Just to explain what is going on here, pytest does a few things automatically based upon how you name things.

If you start a function name with `test_`, pytest assumes it is a test and tries to run it.

Arguments that you put in the `test_` function can be pytest fixtures. In this case, the argument is named `default_material`, which, via name matching, pytest automatically associates with [the fixture function defined above named `default_material`](https://github.com/HEXRD/hexrd/pull/220/files#diff-428844a4010d0b1e799aacd172b9ac423226cd6ce830c4c75858b434be6df8b5R13-R15). It automatically calls that function and passes the returned value as the argument to this test function.

tests/test_material.py
```diff
@@ -0,0 +1,57 @@
...
+    assert are_close(lparms[5], 90)
+
+    default_material.sgnum = 165
+    lparms = default_material.latticeParameters
+
+    # Gamma should be 120, the other angles should be 90
+    assert not lparms_are_close(lparms, [3, 4, 5])
+    assert lparms_are_close(lparms, [3, 4])
+    assert are_close(lparms[3], 90)
+    assert are_close(lparms[5], 120)
```
psavery, line 57: So, this is what is actually being tested here right now. You can see it being tested [here](https://github.com/HEXRD/hexrd/runs/2122961429?check_suite_focus=true#step:9:201).

tests/test_material.py
```diff
@@ -0,0 +1,83 @@
...
+        # Check that the values for ruby match
+        ruby = materials['ruby']
+        params = f['ruby']['LatticeParameters'][()]
+
+        # Convert to angstroms...
+        for i in range(3):
+            params[i] *= 10
+
+        for i in range(6):
+            assert are_close(params[i], ruby.latticeParameters[i])
```
psavery, line 83: And here is loading a materials file and checking that the lattice parameters for ruby match.

psavery:
If you want to run the tests locally, you just need to make sure you have also cloned the [examples repo](https://github.com/hexrd/examples), and set this environment variable to the path on your file system:
```bash
export HEXRD_EXAMPLE_REPO_PATH=/path/to/examples
```

Then, in the hexrd repo, you can pip install the requirements and run the tests like
```python
pip install -r tests/requirements-dev.txt
pytest -s tests
```

Or, specially only run the materials test:
```python
pytest -s tests/test_material.py
```

That will skip the fit grains and find orientations tests, and only run the materials tests.

joelvbernier:
I'll add a bit to this then merge.

joelvbernier:
I need to pass along some commits for this... or should we merge this and add additional testing in a later commit?

psavery:
@joelvbernier I'm good with merging this and adding extra testing later if that's good with you.

saransh13 (approved):
I agree. This is a good start. We can add more tests later.
LGTM.

PR #222, opened 2021-03-16
Small changes to latticeParameters setter...

psavery:
_newPdata() now doesn't always generate a new PlaneData - only when
the hkls have been changed. So we have to set the lattice parameters
on the PlaneData object anyways, just in case a new PlaneData object
doesn't get generated.

Can you look over this, @saransh13? I also made a property to make it easier to get reduced lattice parameters...


saransh13 (approved):
Aah good idea to make it a property! LGTM.

PR #223, opened 2021-03-16
Remove duplicate code.

saransh13:
removing duplicate code in multiple spots for finding allowed hkls for a particular space group.
The one main routine is now in `spacegroup.py` and `WPPF` and `unitcell` now just call it.


psavery (approved):
Just did a quick read-through, but LGTM

psavery:
@saransh13 Although, it looks like there is a failed test we should look into.

saransh13:
Yeah just noticed. I’ll look at it.

PR #226, opened 2021-03-17
Fix ProcessedImageSeries __iter__

psavery:
Before, it was just iterating over the un-processed images
underneath. We need it to perform the processing, though.

Fix the iterator to perform the processing.

### DIFF
--- a/hexrd/imageseries/process.py
+++ b/hexrd/imageseries/process.py
@@ -46,10 +46,7 @@ class ProcessedImageSeries(ImageSeries):
         return len(self._frames) if self._hasframelist else len(self._imser)
 
     def __iter__(self):
-        if self._hasframelist:
-            return (self._imser[i] for i in self._frames)
-
-        return self._imser.__iter__()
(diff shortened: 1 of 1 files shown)


hexrd/imageseries/process.py
```diff
@@ -46,10 +46,7 @@ def __len__(self):
         return len(self._frames) if self._hasframelist else len(self._imser)
 
     def __iter__(self):
-        if self._hasframelist:
-            return (self._imser[i] for i in self._frames)
-
-        return self._imser.__iter__()
+        return (self[i] for i in range(len(self)))
```
psavery, line 49: I believe this should work for the `__iter__` function.

It is a generator expression, so it only processes and returns one frame at a time.

Additionally, `len(self)` is already defined by `__len__` to take into account whether it is using a frame list or not. And, `__getitem__` also takes into account whether a frame list is being used.

joelvbernier (approved):

PR #227, opened 2021-03-17
added handling for missed threshold kwarg

joelvbernier:


joelvbernier:
Test is likely failing since the reference data (maps) were inadvertently formed without applying a threshold; will confirm tomorrow.

hexrd/instrument.py
```diff
@@ -935,8 +936,15 @@ def extract_polar_maps(self, plane_data, imgser_dict,
                             )
                             pass
                         pass
+
+                    # handle threshold if specified
+                    this_det_image = np.array(imgser_dict[det_key])
+                    if threshold is not None:
+                        # !!! NaNs get preserved
+                        this_det_image[this_det_image < threshold] = 0.
```
psavery, line 944: For a `ProcessedImageSeries`, this won't modify the data in place (since it makes and returns a copy). I wonder if it might modify other image series in place, though?

joelvbernier: I checked and I _think_ np.array() always makes a new copy.  I could also just use copy.deepcopy() to be more explicit?  Thoughts?

joelvbernier:
```
a = np.random.randn(5, 5)

a
Out[28]: 
array([[ 0.4201636 ,  0.74112256, -0.41287613,  0.77579234, -2.02663135],
       [-0.11125852, -1.79585744,  0.25656169, -0.84654235, -1.05847483],
       [ 0.27707109,  0.89522628,  0.55110632, -0.65242993, -0.2066158 ],
       [ 0.32781415, -0.28699208,  0.75758832,  0.29122605,  0.96421459],
       [-1.82266946,  0.3363352 , -0.19634523,  0.69217495, -1.54374077]])

b = np.array(a)

a is b
Out[30]: False

a == b
Out[31]: 
array([[ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True],
       [ True,  True,  True,  True,  True]])
```

psavery: Yeah, that looks right. From [the documentation](https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array), it looks like there is a `copy` argument that defaults to `True`. So I think this looks right.

psavery (commented):
LGTM

PR #229, opened 2021-03-17
Unitcell U setter

saransh13:
Improvement for the `U.setter` function in `unitcell` class.
Added `U.setter` to material class. RIght now everything is passed as 1-D or 2-D arrays depending on the isotropic and anisotropic cases resp.

### DIFF
--- a/hexrd/unitcell.py
+++ b/hexrd/unitcell.py
@@ -36,60 +36,12 @@ class unitcell:
 
         self.atom_type = atomtypes
         self.atom_pos = atominfo
-        self.U = U
 
         self._dmin = dmin
 
-        # set some default values for
-        # a,b,c,alpha,beta,gamma
-        self._a = 1.
-        self._b = 1.
-        self._c = 1.
-        self._alpha = 90.
-        self._beta = 90.
-        self._gamma = 90.
+        self.lparms = lp
 
-        if(lp[0].unit == 'angstrom'):
-            self.a = lp[0].value * 0.1
-        elif(lp[0].unit == 'nm'):
-            self.a = lp[0].value
-        else:
-            raise ValueError('unknown unit in lattice parameter')
-
-        if(lp[1].unit == 'angstrom'):
-            self.b = lp[1].value * 0.1
-        elif(lp[1].unit == 'nm'):
-            self.b = lp[1].value
-        else:
-            raise ValueError('unknown unit in lattice parameter')
-
-        if(lp[2].unit == 'angstrom'):
-            self.c = lp[2].value * 0.1
-        elif(lp[2].unit == 'nm'):
-            self.c = lp[2].value
-        else:
-            raise ValueError('unknown unit in lattice parameter')
-
-        if(lp[3].unit == 'degrees'):
-            self.alpha = lp[3].value
-        elif(lp[3].unit == 'radians'):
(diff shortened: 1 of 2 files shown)


hexrd/unitcell.py
```diff
@@ -1246,6 +1209,50 @@ def color_orientations(self,
...
+        is editable. this depends on the space group number or the
+        lattice class
+        """
+        res = False
+        _lpnamelist = list(_lpname)
+        index = _lpnamelist.index(lp_name)
+        editable_fields = _rqpDict[self.latticeType][0]
+        if(index in editable_fields):
+            res = True
+        return res
```
psavery, line 1226: This could be simplified to just:
```python
return index in editable_fields
```

And that removes the `res` variable altogether.

hexrd/unitcell.py
```diff
@@ -1262,94 +1269,130 @@ def compliance(self, v):
...
+    def a(self, val):
+        if self.is_editable("a"):
+            lp = self.lparms
+            lp_valunit = self.fill_correct_lp_vals(
+                lp, val, "a")
+            self.lparms = lp_valunit
+        else:
+            msg = (f"not an editable field"
+                   f" for this space group")
+            print(msg)
```
psavery, line 1315: It might be better if these are exceptions rather than just a print statement.

hexrd/unitcell.py
```diff
@@ -1246,6 +1209,50 @@ def color_orientations(self,
...
+    def is_editable(self, lp_name):
+        """
+        @author Saransh Singh, Lawrence Livermore National Lab
+        @date 03/17/2021 SS 1.0 original
+        @details check if a certain field in the lattice parameter 
+        is editable. this depends on the space group number or the
+        lattice class
+        """
+        res = False
+        _lpnamelist = list(_lpname)
```
psavery, line 1221: Where is this `_lpname` variable defined?

hexrd/unitcell.py
```diff
@@ -1246,6 +1209,50 @@ def color_orientations(self,
...
+                lp_valunit.append(
+                    valWUnit('lp', 'angle',  lp[i], 'degrees'))
+
+        return lp_valunit
+
+    def fill_correct_lp_vals(self, lp, val, lp_name):
+        """
+        added 03/17/2021 SS
+        """
+        index = list(_lpname).index(lp_name)
```
psavery, line 1248: Where is this `_lpname` variable defined?

hexrd/unitcell.py
saransh13: It's defined in `unitcell` module [here](https://github.com/HEXRD/hexrd/blob/669c2334ab39c9bc069e7818e63ff1530f8a32aa/hexrd/unitcell.py#L1592)

hexrd/unitcell.py
saransh13: Same as above.

hexrd/unitcell.py
saransh13: ok, got it.

hexrd/unitcell.py
saransh13: ok modified to a `RuntimeError`

joelvbernier (approved):

PR #230, opened 2021-03-18
Update PlaneData laue and lparms simultaneously

psavery:
When the Laue group changes, the lattice parameters may be incompatible,
and cause an error in PlaneData.__calc().

This commit adds a function to allow us to update both the Laue group
and lattice parameters simultaneously to avoid this issue.

Fixes: hexrd/hexrdgui#821


saransh13 (approved):
I haven't pulled the branch and tested the code but logic looks correct to me.

PR #231, opened 2021-03-18
Use numba, if possible, in hexrd.symmetry.isnew

psavery:
Profiling hexrdgui indicated that the most time consuming function
on hexrdgui startup is hexrd.symmetry.isnew. Adding a numba
decorator provides a 20-40% speed-up for starting hexrdgui when
using numba.

This commit also adds and uses a decorator in hexrd.utils.decorators
called "numba_njit_if_available". Basically, if we are using numba,
it forwards the decorator arguments to numba.njit. And if we are not
using numba, it does nothing.

If we want to improve hexrdgui start-up time in the future, profiling
indicates that these are the next most time consuming functions (in
order):

```python
hexrd.unitcell.unitcell.CalcLength
hexrd.matrixutil.findDuplicateVectors
hexrd.unitcell.unitcell.CalcStar
```

### DIFF
--- a/hexrd/symmetry.py
+++ b/hexrd/symmetry.py
@@ -373,7 +373,7 @@ def MakeGenerators(genstr, setting):
         mat = SYM_fillgen(t)
         genmat = np.concatenate((genmat, mat))
         centrosymmetric = True
-        
+
     n = int(genstr[1])
     if(n > 0):
         for i in range(n):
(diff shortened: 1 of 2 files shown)


saransh13:
I think vectorization of the code is probably also possible. It's written pretty non-efficiently at the moment. Do you think it'll help further?

psavery:
@saransh13 You can give it a try if you'd like, and we can time it to see.

On my computer, without numba, this function had a cumulative time of ~4 seconds when starting up hexrdgui.

With numba, it has a cumulative time of ~0.1 seconds when starting up hexrdgui. It might not be worth extra effort for efficiency improvements.

saransh13:
Aah nice. How about the `CalcLength` and `CalcStar` routines. How much time are they taking? They're operating on single g-vectors right now. I can probably write one which operates on a whole list and stores it so you don't have to call it one at a time every time these numbers are required.

saransh13 (approved):
LGTM

psavery:
@saransh13 Here are those cumulative timings:
`CalcLength`: 0.959 seconds
`CalcStar`: 1.234 seconds

Note that these timings include functions that they call as well. Since `CalcStar` calls `CalcLength`, I'm sure there's a little overlap there.

The top 30 of two different sortings are listed below. "cumulative_time.txt" is sorted by cumulative time, which is the
time spent inside that function *including* the time spent in functions that function calls.

"total_time.txt" is sorted by total time, which is only the total time spent inside that function alone (does not include
time spent in functions that function calls).

[cumulative_time.txt](https://github.com/HEXRD/hexrd/files/6178396/cumulative_time.txt)
[total_time.txt](https://github.com/HEXRD/hexrd/files/6178397/total_time.txt)

PR #234, opened 2021-03-23
Material dmin fix

saransh13:
name is a property.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -139,7 +139,7 @@ class Material(object):
         or hdf5
         """
         if name is None:
-            self.name = Material.DFLT_XTAL
+            self._name = Material.DFLT_XTAL
 
         self.description = ''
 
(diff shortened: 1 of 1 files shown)


joelvbernier (approved):

hexrd/material.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, name=None,
         or hdf5
         """
         if name is None:
-            self.name = Material.DFLT_XTAL
+            self._name = Material.DFLT_XTAL
```
psavery, line 142: I might recommend we change the logic here to be something like:

```python
if name is None:
    name = Material.DFLT_XTAL

self._name = name
```

We want a guarantee that there are no possible code paths where `self._name` will not be set, or else we might run into an `AttributeError`.

In its current state, I could see it potentially hitting an `AttributeError` if `name is not None` and we are loading from a cif file. There may be other possible code paths in the future with changes to this class. Having some kind of assurance that `self._name` will be set I think is very helpful.

psavery: Okay, I think #236 fixes this...

Issue #155, opened 2021-01-21
Warning running find-orientations and fit-grains on Windows


joelvbernier:
I am getting the following during both the fiber generation, the orientation scoring, and grain fitting -- all using `map`:
```
C:\Users\berni\miniconda3\envs\hexrd_0.8\lib\runpy.py:127: RuntimeWarning: 'hexrd.cli.main' found in sys.modules after import of package 'hexrd.cli', but prior to execution of 'hexrd.cli.main'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
```
Thoughts @psavery and/or @cjh1?

cjh1:
I think removing [this](https://github.com/HEXRD/hexrd/blob/master/hexrd/cli/__init__.py#L11) line will resolve it.

cjh1:
I don't see this warning when running the hexrd cli on Windows from a conda install. When are you seeing this?

joelvbernier:
conda package install on Windows 10 as well.  @donald-e-boyce, do you think that we can safely remove [this](https://github.com/HEXRD/hexrd/blob/master/hexrd/cli/__init__.py#L11) line?

donald-e-boyce:
I think you can safely remove it.

cjh1:
@joelvbernier I am still not see this even with hexrd installed with pip. What is the hexrd invocation you are using?

joelvbernier:
@cjh1 -- running the CLI from a local install in dev mode on Windows 10 (Anaconda Powershell).  Each process echos back the warning.

cjh1:
So you get it if you run just `hexrd --help`?

joelvbernier:
I think this has been resolved.

PR #242, opened 2021-03-29
Update entrypoint for conda

psavery:
Since [this import was removed](https://github.com/HEXRD/hexrd/blob/c19aaaf5e0ad6660475b13aad2b8f98424faf64a/hexrd/cli/__init__.py#L11), the entrypoint needed updating.

This now matches the [entrypoint in setup.py](https://github.com/HEXRD/hexrd/blob/c19aaaf5e0ad6660475b13aad2b8f98424faf64a/setup.py#L81), so should hopefully work.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -10,7 +10,7 @@ build:
   detect_binary_files_with_prefix: true
   osx_is_app: yes
   entry_points:
-    - hexrd = hexrd.cli:main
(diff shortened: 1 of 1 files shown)


joelvbernier:
Ah shit, yeah -- this was the one causing the weird warning on Windows in #155.  Guess we'll just live with that.

joelvbernier (approved):

PR #246, opened 2021-04-01
added pixel angle gradients

joelvbernier:
Added the following methods to `instrument.PlanarDetector`:

```
pixel_tth_gradient(origin=[0., 0., 0.])
pixel_eta_gradient(origin=[0., 0., 0.])
```
which return |&nabla;2&theta;| and |&nabla;&eta;|, respectively.  The `origin` kwarg is handed to `PlanarDetector.pixel_angles()`.  Logic for handling the &eta; branch cut is included.  Results for the default instrument shown below
image

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -2417,6 +2417,37 @@ class PlanarDetector(object):
         eta = angs[1].reshape(self.rows, self.cols)
         return tth, eta
 
+    def pixel_tth_gradient(self, origin=ct.zeros_3):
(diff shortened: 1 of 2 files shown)


joelvbernier:
@psavery : maybe replace the default polar pixel sizes with
```
tth_ps_DFLT = np.power(10, np.round(np.log10(100*np.median(panel.pixel_tth_gradient()))))
eta_ps_DFLT = np.power(10, np.round(np.log10(100*np.median(panel.pixel_eta_gradient()))))
```
(_i.e._, the max of these quantities over all the detectors in the instrument).

PR #247, opened 2021-04-02
Use numba.njit(cache=True, nogil=True) for all

psavery:
Using "njit" instead of "jit" means to use the "nopython" argument,
which can be faster. But it may not be allowed for some function
argument types.

Caching will get numba to cache the functions it compiles so that it
does not have to compile them again. This can provide a big speed-up
after the functions have been called once and saved.

Note that if you edit the file that contains the function, the cache
will automatically be re-generated the next time it is ran.

nogil tells numba to release the GIL. This is helpful in case the
function is used in any python multi-threading.

Caching is also helpful if the function is used in any python
multi-threading because numba can only compile on one thread at a time...

In any case, these changes all have the potential to be good, but we should
definitely test them all, and make sure that everything still works, gives
the same results, and takes at most the same amount of time that it did
previously.

### DIFF
--- a/hexrd/distortion/dexela_2923.py
+++ b/hexrd/distortion/dexela_2923.py
@@ -70,7 +70,7 @@ def _find_quadrant(xy_in):
 
 
 if USE_NUMBA:
-    @numba.njit
+    @numba.njit(nogil=True, cache=True)
     def _dexela_2923_distortion(out_, in_, params):
         for el in range(len(in_)):
             xi, yi = in_[el, :]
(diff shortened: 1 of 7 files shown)


joelvbernier:
LMK when this looks ready @psavery

psavery:
@joelvbernier I've played around with a few things in the GUI, and it looks okay. It also passes the HEDM workflow hexrd tests. Is there anything else you think we should test?

joelvbernier:
@psavery it should be ready if it passed :-)  The CI test exercises all that code.

joelvbernier (approved):

PR #252, opened 2021-04-09
Structure setters

saransh13:
Make sure numbers are updated before updating everything.


saransh13:
@psavery @joelvbernier any idea why this is passing for linux and windows but failing for mac?

psavery:
It looked to me that it just didn't run at all. I'll try to restart the job.

joelvbernier (approved):

PR #253, opened 2021-04-09
make sure exclusions are set to default values when properties are up…

saransh13:
…dated. Was leading to hkl list artificially getting cutoff.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -283,11 +283,10 @@ class Material(object):
             exclusions = numpy.ones(hkls.shape[1], dtype=bool)
             exclusions[new_indices] = old_pdata.exclusions[old_indices]
 
-            if numpy.all(exclusions):
-                # If they are all excluded, just set the default exclusions
-                self.set_default_exclusions()
-            else:
-                self._pData.exclusions = exclusions
+            # always set it back to default values
+            # @SS 04/09/2021
+            self.set_default_exclusions()
+
         else:
             # Make the PlaneData object from scratch...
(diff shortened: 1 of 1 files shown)


hexrd/material.py
```diff
@@ -283,11 +283,10 @@ def _newPdata(self):
...
             exclusions[new_indices] = old_pdata.exclusions[old_indices]
 
-            if numpy.all(exclusions):
-                # If they are all excluded, just set the default exclusions
-                self.set_default_exclusions()
-            else:
-                self._pData.exclusions = exclusions
+            # always set it back to default values
+            # @SS 04/09/2021
+            self.set_default_exclusions()
```
joelvbernier, line 288: `plane_data.exclusions = None` resets all.  Just use that?

saransh13: Ok, that works too.

joelvbernier: Is this ready/tested @saransh13 ?

joelvbernier (approved):

hexrd/material.py
psavery: @saransh13 @joelvbernier FYI, the code block above this (lines 280 to 284) is written to try to preserve the exclusions, so that in hexrdgui, when the user modifies things such as lattice parameters, dmin, etc., the exclusions do not get modified.

Now, the exclusions are always getting modified when the user makes changes to material properties. Does it make sense for us to always reset the exclusions when properties on the material get changed?

PR #255, opened 2021-04-16
moved 2d peak detection to imageutil

joelvbernier:
a little reorganization -- the util routine for peak finding was hidden inside `hexrd.findorientations`


psavery (approved):
LGTM. I put up hexrd/hexrdgui#893 to reflect these changes in the GUI.

PR #256, opened 2021-04-20
added table of X-ray energies

joelvbernier:
The new HDF5 file has characteristic X-ray energies for all elements with available data (sourced from the [CXRO X-ray data booklet](https://xdb.lbl.gov/)).  The top-level groups are either atomic number or element symbol.  Under each element group are two subgroups:
- `absorption_edges`
- `emission_data`
entries are self-explanatory.
image
https://github.com/HEXRD/hexrdgui/issues/894 on hexrdgui depends on this PR.


saransh13 (approved):
File looks good to me.

PR #257, opened 2021-04-20
Improve performance of eta omega map generation

psavery:
This generates ring parameters one time in its own loop, and then
swaps the previous ordering of the loops so that looping over the
images is on the outside and looping over the rings is on the inside.

This results in each image only being loaded once, rather
than once for each ring, as it was before. Since loading the images is
time consuming, this results in a nice speedup.

For the GE example in the examples repository, my performance improved
from 17 seconds to 5 seconds, and the same result was obtained.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -889,93 +889,44 @@ class HEDMInstrument(object):
                 # initialize maps and assing by row (omega/frame)
                 nrows_ome = len(omegas)
 
-                ring_maps = []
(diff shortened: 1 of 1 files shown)


joelvbernier (approved):

joelvbernier:
Awesome @psavery -- The other thing I wanted to check (if you think it would be appropriate) would be to replace the `ThreadPoolExecutor` with a `ProcessPoolExecutor`.  I think we might see better performance, although it might be marginal.

PR #258, opened 2021-04-21
Eta omega map generation speedup

psavery:
Before, the multi-threading was a little too fine-grained, and the main thread was running a lot of code that the threads could have been running. Now, it's a little more coarse-grained, and a fair amount of time is being spent in the threads other than the main thread.

This resulted in nearly a 2x speed-up for the GE example.

Multiprocessing was attempted as well, but pickling and unpickling image series seems to be expensive, and makes it slow. Perhaps it's something we can pursue if we find a faster way to pickle/unpickle image series data.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -33,7 +33,7 @@ Created on Fri Dec  9 13:05:27 2016
 import copy
 import os
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
-import functools
+from functools import partial
 
 import yaml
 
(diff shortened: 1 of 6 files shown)


hexrd/instrument.py
```diff
@@ -867,98 +868,56 @@ def extract_polar_maps(self, plane_data, imgser_dict,
         delta_eta = eta_edges[1] - eta_edges[0]
         ncols_eta = len(eta_edges) - 1
 
+        max_workers = os.cpu_count()
```
joelvbernier, line 871: @psavery -- this should be an input arg; the HEDM config object has a `multiprocessing` attribute that specifies this number.  I have found that on intel processors with hyperthreading, where the `cpu_count` is 2x the number of physical cores, asking for all of the cpus can lead to bad performance.  Now that is for multiprocessing instead of multithreading...  maybe it is ok to as for all available for the latter?

joelvbernier (approved):

PR #259, opened 2021-04-21
removing a space in "W"

saransh13:
there was a space in the periodic table dictionary for W leading to key error. fixed with this.


hexrd/material.py
```diff
@@ -684,9 +684,9 @@ def vol_per_atom(self):
     def atom_pos(self):
         return self.unitcell.atom_pos
 
-    @property
-    def atom_type(self):
-        return self.unitcell.atom_type
+    # @property
+    # def atom_type(self):
+    #     return self.unitcell.atom_type
```
joelvbernier, line 689: @saransh13 -- does removing this property break anything?

saransh13: Ran fine for me. I noticed it was doubly defined. There is an `atomtypes`
property on `material.Material` which is actually doing the updating.

joelvbernier (approved):

PR #260, opened 2021-04-30
Memoize pixel_solid_angle results

psavery:
Since this is an expensive computation, memoizing can help a lot.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -2103,14 +2103,6 @@ class PlanarDetector(object):
 
     @property
     def pixel_solid_angles(self):
-        # connectivity array for pixels
-        conn = cellConnectivity(self.rows, self.cols)
-
-        # result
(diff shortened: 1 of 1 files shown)


psavery:
@joelvbernier If there are only a few different detector shapes around out there, I wonder if it might also be a good idea to pre-compute the pixel solid angles for the main detectors we'd be using, and store them in a data file somewhere (like in hexrd). Then hexrd could check first to see if the detector shape has already been computed and stored in the cache file, and if not, then it can compute it.

That could make it so that for most detectors, users wouldn't need to compute the pixel solid angles at all.

joelvbernier:
@psavery -- the solid angles are not just a function of detector shape and intrinsic pixel size, but also the affine transform parameters.  Therefore, it isn't possible to calculate a generally applicable set.

joelvbernier (approved):

PR #267, opened 2021-05-19
introduced atomdata setter for simultaneous setting of Z, positions+occupancy, U and charge. 

saransh13:
This avoids shape conflicts. This will fix https://github.com/HEXRD/hexrdgui/issues/933


saransh13:
@psavery Any idea why the windows check failed? The error looks alien to me.

psavery:
> @psavery Any idea why the windows check failed? The error looks alien to me.

It's been failing during the conda install of `scikit-learn` :slightly_frowning_face:. This is happening on other PRs too, such as #189.

It's not something we introduced, though.

hexrd/material.py
```diff
@@ -707,15 +707,17 @@ def U(self):
...
+        # self.unitcell.U = Uarr
+        # if self.unitcell.U.shape == Uarr.shape:
+        #     if not numpy.allclose(self.unitcell.U, Uarr):
+        #         self.unitcell.U = Uarr
+        #         self.update_structure_factor()
+        #     else:
+        #         return
+        # else:
+        #     self.unitcell.U = Uarr
+        #     self.update_structure_factor()
```
psavery, line 720: So we are removing these updates since we should only modify the data with `_set_atomdata` now?

hexrd/material.py
```diff
@@ -905,32 +909,86 @@ def _set_atomtype(self, v):
...
+        #     self._atomtype = numpy.array(v)
+            #self._newUnitcell()
+            #self.update_structure_factor()
+        self._atomtype = v
 
     atomtype = property(
         _get_atomtype, _set_atomtype, None,
         "Information about atomic types")
 
+    def _set_atomdata(self, atomdata):
```
psavery, line 939: I think this would feel a little more natural if the `atomdata` argument were split into three separate arguments (atomtype, atominfo, and U), rather than passing a list of all three.

psavery (approved):
LGTM

saransh13:
Ok, makes sense. Should be an easy change.

In the future, once we have the option to input the charge state of the
atoms in the gui, we can also pass that information as the fourth option.

Right now the function defaults to neutral atoms.

saransh13:
That’s the best option I could come up with. But certainly open to other
ideas.

joelvbernier (approved):

joelvbernier:
Not sure why the Windows conda package failed here; re-running.

joelvbernier:
Looks like scikit-learn is causing a problem building the windows conda package?  Failed again...  @psavery?

psavery:
> Looks like scikit-learn is causing a problem building the windows conda package? Failed again... @psavery?

Yeah, that's right. Specifically, it is failing when it tries to run `conda install anaconda conda-build`. 

It's failing when it tries to install conda-build. It hasn't even reached the step where it tries to install hexrd.

Pretty lame. No idea why.

psavery:
@saransh13 I rebased this on master to fix the CI. If you want to checkout this branch locally, you'll need to delete your local branch and fetch this one first.

hexrd/material.py
joelvbernier: @saransh13 -- can we resolve this today?

hexrd/material.py
saransh13: Ok this was part of the latest push.

hexrd/material.py
saransh13: This is resolved.

joelvbernier (approved):

PR #268, opened 2021-05-26
Fixing long standing bug in multipeak fitting non-linear backgrounds.

darrencpagan:
Small simple fix to peak fitting. Also remembering how to submit fixes before attempting larger update to peak fitting module.


joelvbernier (approved):

darrencpagan:
What did this fail? I was looking for this fix in the code today.

psavery:
@darrencpagan We have been running into an issue installing scikit learn 0.24.2 on Windows in our CI. Looks like that was the failure here. We downgraded it to 0.24.1 in #269, which is in master. So I think things should be fine.

psavery:
Looks like the master branch CI is passing, which is the most important :slightly_smiling_face:

PR #270, opened 2021-05-27
filter copy of maps instead of main object

joelvbernier:
There was an oversight where the call to `clean_maps` to enable feature detection was modifying the underlying object and removing NaNs, which are critical for indexing cases where the detectors have gaps in eta coverage.


joelvbernier:
@psavery -- use this branch to test the GUI/CLI inconsistency

psavery (approved):
LGTM.

joelvbernier (commented):
LGTM

PR #272, opened 2021-05-27
Update meta.yaml

joelvbernier:
Just to be safe, pin fabio version to >=0.11 (on conda-forge)

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -21,7 +21,7 @@ requirements:
 
   run:
     - appdirs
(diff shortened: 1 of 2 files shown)


joelvbernier:
@psavery -- should we do anything in setup.py as well?

psavery:
@joelvbernier We can do the same kind of thing if you'd like.

joelvbernier (commented):
I think this is ready.

joelvbernier:
I'll merge this upon receiving your blessing @psavery

PR #277, opened 2021-05-28
[WIP] DBG check conda-forge sklearn 0.24.2 package

glemaitre:
Related to https://github.com/scikit-learn/scikit-learn/issues/20140

Checking if the conda-forge binary is failing

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -46,7 +46,8 @@ jobs:
     - name: Install build requirements
       run: |
           conda activate hexrd
-          conda install anaconda conda-build scikit-learn==0.24.1
+          conda install anaconda conda-build
+          conda install -c conda-forge scikit-learn
 
       # This is need to ensure ~/.profile or ~/.bashrc are used so the activate
       # command works.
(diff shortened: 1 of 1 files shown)


glemaitre:
@psavery we can attend to only install scikit-learn at first and see if it fails. Otherwise, it means that this is an interaction with the `anaconda` distribution as well.

psavery:
@glemaitre It looks like installing from `conda-forge` worked. So it is indeed a problem specific to the `anaconda` distribution.

glemaitre:
Yep. I am interested to know if it is related to install `anaconda` and `scikit-learn` or just `scikit-learn`. It would be easier to provide a bit more inside to anaconda.

glemaitre:
So `conda install scikit-learn` is enough to reproduce the problem. I am unsure where we can report this thought :). Maybe @ogrisel has an idea.

psavery:
@glemaitre Thank you for looking into this. We changed our project to now use dependencies from `conda-forge` instead of `anaconda`. That fixed the `scikit-learn` issues we were having. So I think we can close this PR.

PR #286, opened 2021-06-15
Lorentz Polarization factor for 2D detectors

saransh13:
computes the lorentz polarization factor for any 2D planar detector given the fraction of horizontal and vertical polarizations of the incident x-ray. 

Note that the factors **aren't** properties of the `PlanarDetector` class since it requires the inputs of the polarization fractions. 

The computed factors for a typical MEC configuration gives reasonable results shown below.

Screen Shot 2021-06-14 at 7 37 13 PM


psavery (approved):
LGTM. Nice use of memoize!

PR #290, opened 2021-06-22
added eta patch specification option

joelvbernier:
Added the ability to specify (rather than autogenerate) eta bin centers in `HEDMInstrument.extract_line_positions`.  The involved a pass-thru for a similar arg in `PlanarDetector.make_powder_rings`.  There are likely more changes to this function coming, as we should use it for interpolating Laue spots (for Laue cal among other things).


psavery (approved):
LGTM

Issue #225, opened 2021-03-17
Consider modifying the number of default exclusions for materials


psavery:
Should we consider increasing the number of default exclusions for materials?

For instance, for ruby from hexrdgui:
```python
>>> from hexrd.material import Material
>>> mat = Material('ruby', 'materials.h5')
>>> mat.planeData.exclusions
array([False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False, False, False, False, False, False, False, False,
       False, False])
```

Results in:
![image](https://user-images.githubusercontent.com/9558430/111412141-d66d1480-86a9-11eb-87dc-a1d506f8612b.png)

in hexrdgui, at program start-up (if ruby powder overlays are visible).

This is particularly painful if you try to run something like generating eta omega maps, and you forget to modify the exclusions (because it will take a very long time, and many more hkls will be used than typically needed).

Should we modify the [default exclusions function](https://github.com/HEXRD/hexrd/blob/4736bcbbb9f0fc3cdeccc00528e734de6407cc03/hexrd/material.py#L280) so that fewer hkls are included?

joelvbernier:
I'd say if the default should be all excluded but the first 5 reflections.  The exclusions set technically _can_ be preserved unless the space group is changed (in general).

Issue #273, opened 2021-05-27
KeyError in CalcXRFormFactor


psavery:
On Linux, at least, for this attached [example.zip](https://github.com/HEXRD/hexrd/files/6556066/example.zip), if I run
`hexrd find-orientations mruby_config.yml`, I get a `KeyError` (attached below).

It looks like this is being caused by the `charge` variable being bytes instead of a string.
The check is essentially doing this:

```python
if b'0' == '0':
    ...
else:
    ...
```

At least on my system, `b'0' == '0'` returns `False`.

I guess a potentially good solution to this is to make sure that wherever the `charge` variable
gets set, that it is a string instead of bytes. Something like this:

```python
if isinstance(charge, (bytes, bytearray)):
    charge = charge.decode()
```

Full error is below:

```python
21-05-27 13:39:04 - === begin find-orientations ===
21-05-27 13:39:04 - "working_dir" not specified, defaulting to "/data/patrick/hexrd/hydra/example"
Traceback (most recent call last):
  File "/home/patrick/virtualenvs/hexrd/bin/hexrd", line 33, in <module>
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/cli/main.py", line 73, in main
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/cli/find_orientations.py", line 123, in execute
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/config/root.py", line 68, in analysis_id
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/config/root.py", line 57, in material
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/config/material.py", line 88, in beam_energy
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/config/material.py", line 39, in materials
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 975, in load_materials_hdf5
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 976, in <dictcomp>
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 201, in __init__
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/material.py", line 341, in update_structure_factor
[...]

PR #293, opened 2021-06-24
Enable the first five hkls by default

psavery:
Rather than enabling all hkls with two theta less than 90 degrees
by default, we are now enabling the first five hkls by default.

This adds two utility functions, though, for automatically enabling
hkls: one for enabling all hkls below an index, and one for enabling
all hkls below a two theta value (which is copied from the previous
`set_default_exclusions()` function).

The first utility function gets called by default, which enables
only the first five hkls.

Fixes: #225

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -309,10 +309,8 @@ class Material(object):
 
     def enable_hkls_below_index(self, index=5):
         # Enable hkls with indices less than @index
-        exclusions = self._pData.exclusions
-        for i in range(len(exclusions)):
-            exclusions[i] = i >= index
-
+        exclusions = np.ones_like(self._pData.exclusions, dtype=bool)
+        exclusions[:index] = False
         self._pData.exclusions = exclusions
 
     def enable_hkls_below_tth(self, tth_threshold=90.0):
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -304,10 +304,23 @@ class Material(object):
             self.set_default_exclusions()
 
     def set_default_exclusions(self):
+        # Enable only the first 5 by default
+        self.enable_hkls_below_index(5)
+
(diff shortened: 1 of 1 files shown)


hexrd/material.py
```diff
@@ -302,10 +302,23 @@ def _newPdata(self):
...
 
     def set_default_exclusions(self):
+        # Enable only the first 5 by default
+        self.enable_hkls_below_index(5)
+
+    def enable_hkls_below_index(self, index=5):
+        # Enable hkls with indices less than @index
+        exclusions = self._pData.exclusions
+        for i in range(len(exclusions)):
+            exclusions[i] = i >= index
```
joelvbernier, line 312: Typically I prefer the vectorized approach:
```
exclusions = np.ones_like(self._pData.exclusions, dtype=bool)
try:
    exclusions[:index] = False
except(IndexError):
    exclusions = np.zeros_like(self._pData.exclusions, dtype=bool)
```

joelvbernier (commented):
commented in line.

joelvbernier:
Ah -- I see this is failing our CI `find-orientations` test due to the fact that it uses the first 6 reflections;  this should be related to the larger issue of the ability to save and restore exclusions in material instantiation.  Must fix that, or will have to scrap this approach.

psavery:
> Ah -- I see this is failing our CI `find-orientations` test due to the fact that it uses the first 6 reflections; this should be related to the larger issue of the ability to save and restore exclusions in material instantiation, which is actualy a very useful feature. This was the case in the older material class. We must fix that, or will have to scrap this approach.

I think there is code for that. I'll fix it.

joelvbernier:
Need to get this fixed ASAP.

psavery:
Tests are now failing due to h5py issues that are mentioned in #273

joelvbernier (approved):

Issue #288, opened 2021-06-16
Find-orientations depreciation warning


darrencpagan:
We are currently receiving this numba depreciation warning when using find-orientations with hexrd 0.8.5. We are currently using a list of thresholds in the find-orientations.threshold config option.
```
warnings.warn(NumbaPendingDeprecationWarning(msg, loc=loc))
/tri-s-1/s2/dcp5303/sw/miniconda3/envs/20210530_hexrdgui/lib/python3.8/site-packages/numba/core/ir_utils.py:2031: NumbaPendingDeprecationWarning:
Encountered the use of a type that is scheduled for deprecation: type 'reflected list' found for argument 'threshold' of function '_filter_and_count_hits'.

For more information visit https://numba.pydata.org/numba-doc/latest/reference/deprecation.html#deprecation-of-reflection-for-list-and-set-types

File "../../../dcp5303/sw/miniconda3/envs/20210530_hexrdgui/lib/python3.8/site-packages/hexrd/indexer.py", line 605:
    @numba.njit(nogil=True, cache=True)
    def _filter_and_count_hits(angs_0, angs_1, symHKLs_ix, etaEdges,
```

joelvbernier:
Thanks @darrencpagan -- I'll look into strategies for replacing the offending code before it causes problems.

psavery:
@joelvbernier Maybe we just need to make sure the `threshold` argument is a numpy array before setting it [here](https://github.com/HEXRD/hexrd/blob/2b6f1445b25db30ab6007a111d0aa1bdbd023742/hexrd/indexer.py#L259)?

PR #296, opened 2021-06-30
adjust output of line position extraction

joelvbernier:
Now `collapse_eta` returns properly summed intensities.

Fixes: #288


psavery (approved):
LGTM. This, along with hexrd/hexrdgui#971, fixes the auto powder issue for me.

PR #298, opened 2021-07-23
updated legacy instrument config conversion

joelvbernier:
The legacy conversion script to take old hexrd 0.5.x-style instrument config files to the current (>=v0.6) format was out of date.  A user (@hsheng94) had an example of a legacy instrument file from CHESS (closes HEXRD/hexrdgui#987).  This adds the new conversion script, which includes a CLI.


psavery (approved):
I didn't try running it, but LGTM

PR #301, opened 2021-07-27
pass charge value to _set_atomdata.

saransh13:
ability to charge as part of _set_atomdata function in Material class. This will be utilized for PR https://github.com/HEXRD/hexrdgui/pull/973 in the GUI.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -964,7 +964,7 @@ class Material(object):
         _get_atomtype, _set_atomtype, None,
         "Information about atomic types")
 
-    def _set_atomdata(self, atomtype, atominfo, U):
+    def _set_atomdata(self, atomtype, atominfo, U, charge):
         """
         sometimes the number of atom types and their
         positions are changed when creating a material.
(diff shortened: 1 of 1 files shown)


psavery (approved):
LGTM

psavery:
When we merge this, we need to also merge hexrd/hexrdgui#973 right away, so that the GUI doesn't break...

joelvbernier:
Well merge this one when we are ready to stage a new GUI release including hexrd/hexrdgui#513, yeah?  Or are things broken in the gui without this @saransh13?

saransh13:
The current GUI works. But if we merge this one, then the GUI code would
need to be updated. I think Patrick already has a PR for that.

psavery:
Yeah, we currently need to merge this one before we merge hexrd/hexrdgui#973. That can wait until after the release if you prefer, though, @joelvbernier

joelvbernier:
Question: it seems like if I merge this, then I should tag at 0.8.10 and point the corresponding hexrdgui (post merging hexrd/hexrdgui#973) to require hexrd >= 0.8.10, yes?

PR #305, opened 2021-08-02
Fix python=3.8 for Linux packaging

psavery:
Lately, the Linux packaging has been using python 3.9. Python 3.9 may
work, but hexrdgui is currently fixed to python 3.8. Fix python to 3.8
for the Linux packaging as well until we decide to upgrade python in
both hexrd and hexrdgui simultaneously.


cjh1 (approved):
Excellent, thank you

PR #307, opened 2021-08-05
Use conda-forge instead of anaconda channels

psavery:
This is an attempt to use the `conda-forge` channel instead of the
`defaults` and `anaconda` channels, for all packages.
    
This will hopefully fix the issue where we had to downgrade scikit-learn
because it would fail to install on Github Actions on Windows. This will
hopefully also fix a current issue for the Windows packaging where
the llvmlite dll cannot be loaded by numba. It makes more sense for us
to install these packages from conda-forge since the respective project
developers actually maintain the packages on conda-forge (whereas their
packages on the anaconda/defaults channels are maintained by anaconda).
    
We will first need to make sure the packaging and tests pass, and then
move on to test the packages manually.
    
If we run into crashing issues on Mac with openblas, as we have in the
past, we may be able to resolve it by setting the environment variable
`OPENBLAS_NUM_THREADS=1`.

### DIFF
--- a/.github/workflows/container_build.sh
+++ b/.github/workflows/container_build.sh
@@ -7,6 +7,6 @@ chmod a+x Miniconda3-latest-Linux-x86_64.sh
 ${HOME}/miniconda3/bin/conda create -y -n hexrd python=3.8
 ${HOME}/miniconda3/bin/activate hexrd
 ${HOME}/miniconda3/bin/conda activate hexrd
-${HOME}/miniconda3/bin/conda install conda-build -y
+${HOME}/miniconda3/bin/conda install --override-channels -c conda-forge conda-build -y
 mkdir output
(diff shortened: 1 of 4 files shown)


psavery:
I tried the tests locally with all of the packages from conda-forge for Windows, Mac, and Linux, and they all passed. I think it would be nice if this one and hexrd/hexrdgui#1006 are merged simultaneously, though. So I think this one will be ready once that one is...

psavery:
`python.app` has been causing issues with packaging in HEXRD/hexrdgui#1006 (due to `conda-pack`, see conda/conda-pack#68).

In the previous commit, I removed `osx_is_app: yes`, which removes the need to install `python.app`. I tested the resulting conda package on Mac and it passed the tests, leading me to believe that we do not need that setting.

Removing that setting allows us to skip installing `python.app` and thus avoid the errors with `conda-pack` in hexrdgui...

We'll keep an eye out for any issues this may cause.

cjh1 (approved):

Issue #306, opened 2021-08-04
median_iter() has slow performance for higher numbers of chunks


psavery:
[Attached is an example](https://github.com/HEXRD/hexrd/files/6928055/median_time_test.zip), for which you need to download the `RUBY_4537.ge` and `RUBY_4538.ge` files from [here](https://drive.google.com/drive/u/0/folders/1CgGzNEFeaB06I-Q3tbB0lf6nVOjN93fr) to run it.

For example, on my computer, I just ran the script and received the following output:
```bash
Running median()...
median() took 12.79 seconds
Running median_iter() with nchunks=1...
median_iter() with nchunks=1 took 13.02 seconds
Running median_iter() with nchunks=60...
median_iter() with nchunks=60 took 72.32 seconds
```

You can see that that with `nchunks=60`, it took ~5.5x as long to run it compared to `nchunks=1` (and a Windows user reported it taking ~5 minutes for the same data). Since we are using `median_iter()` for progress updates, `nchunks=60` is a reasonable number for us to use in the GUI as well, because it allows us to show an update to the progress bar every 1 or 2 percent.

I believe that the issue is that each frame in the image series is getting expanded for each chunk, which, for our example above, leads to 120 frames * 60 chunks = 7200 frame expansions. This is because each frame is being expanded [here](https://github.com/HEXRD/hexrd/blob/2fef12289aba6adc5212505d83972fd92f2a5b23/hexrd/imageseries/stats.py#L182-L183), and the `_toarray()` function is being called for each chunk [here](https://github.com/HEXRD/hexrd/blob/2fef12289aba6adc5212505d83972fd92f2a5b23/hexrd/imageseries/stats.py#L135).

I understand that part of the issue is that to obtain every median value `ij` in the resulting image, you need `ij` of every frame, and currently, to obtain any single value from a frame, you must expand the whole frame.

If I understand the code correctly, I believe `median()` (which calls [`percentile()`](https://github.com/HEXRD/hexrd/blob/2fef12289aba6adc5212505d83972fd92f2a5b23/hexrd/imageseries/stats.py#L120-L123)) expands the entire image series and keeps it in memory for the `percentile()` call. Perhaps one solution for the performance issue with `percentile_iter()` is to keep in memory as many frames as it can, to reduce the number of times each frame needs to be expanded. It may be able to keep in memory all of the frames if there is enough memory on the system, and thus match the performance of `median()` / `percentile()`.

donald-e-boyce:
[...]

PR #309, opened 2021-08-12
Imageseries stats

donald-e-boyce:
This addresses Issue #306. For percentile_iter and median_iter, a buffer is created to store as many frames as possible so as to avoid repeated imageseries frame copies in stats._toarray(). The size of the buffer is half of available memory when the stats module is loaded. 

Additionally,
* This fixes a bug in the percentile iterator that duplicates a row between chunks. It did not affect the result, but performed unnecessary computations.
* Unit tests were improved: old ones strengthened and new ones added.

### DIFF
--- a/hexrd/imageseries/stats.py
+++ b/hexrd/imageseries/stats.py
@@ -15,12 +15,16 @@ img = stats.average(ims)
 for img in stats.average_iter(ims, 10):
     # update progress bar
     pass
+
+NOTE:
+* Perhaps we should rename min -> minimum and max -> maximum to avoid
+  conflicting with the python built-ins
 """
 import numpy as np
 
 from psutil import virtual_memory
 
-# Default Buffer: half of available memory
+# Default Buffer Size: half of available memory
 vmem = virtual_memory()
 STATS_BUFFER = int(0.5*vmem.available)
 del vmem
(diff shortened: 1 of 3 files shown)


psavery (approved):
Here are my recent timings:

Master branch:
```
median() took 12.65 seconds
median_iter() with nchunks=1 took 12.16 seconds
median_iter() with nchunks=60 took 65.77 seconds
```

This branch:
```
median() took 12.69 seconds
median_iter() with nchunks=1 took 12.42 seconds
median_iter() with nchunks=60 took 8.24 seconds
```

It's certainly much faster, which is great! I wonder why more chunks is running faster?

Anyways, assuming the tests confirm it is producing the correct answers, LGTM.

hexrd/imageseries/stats.py
```diff
@@ -118,31 +122,42 @@ def average_iter(ims, nchunk, nframes=0):
...
     for s in stops:
         r1 = s + 1
-        img[r0:r1] = np.percentile(_toarray(ims, nf, (r0, r1)), pctl, axis=0)
-        r0 = s
+        img[r0:r1] = np.percentile(
+            _toarray(ims, nf, rows=(r0, r1), buffer=buffer),
+            pctl, axis=0
+        )
+        r0 = r1
         yield img.astype(np.float32)
```
psavery, line 150: If we are going to always convert the image to `np.float32`, should we just set the dtype as `np.float32` in the call to `np.zeros`?

donald-e-boyce:
I fixed the nbf = 0 problem. That was simple.

The suggestion to set the image dtype to be np.float32 inside percentile is more complicated. It affects the code in several other places. 

* In `alloc_buffer`, the buffer is set to be the dtype of the imageseries, so we would need to change that to float32.
* In `_toarray`, each frame of the imageseries would have to be converted to float32 every time it's called. Also, I don't like losing generality of this function or alloc_buffer.
* In `percentile` we would also drop the `astype` in the return statement.
* I would also have to update the unit tests since they compare the results to numpy function on that dtype , then converted to float32, which may different from converting to float32 and then applying the numpy function.

If it were a 2-line fix, it would be fine, but here I am tempted to leave well enough alone here.

PR #315, opened 2021-09-16
added kwarg for distortion application in cart<->ang transforms

joelvbernier:
Finally added the option to apply distortion (or inverse) in the transformation methods taking (2θ, η) → (x, y) and (x, y) → (2θ, η); namely the `PlanarDetector.angles_to_cart()` and `PlanarDetector.cart_to_angles()` methods.  The previous behavior was to assume that (x, y) were always in the "ideal" (read: undistorted) frame.  This is preserved in the default value of the new kwarg: `apply_distortion=False`.  Of course if `self.distortion is not None` is `False`, then `apply_distortion=True` has no effect.

Documentation for the affected methods has been updated as well.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -2527,9 +2527,38 @@ class PlanarDetector(object):
                 on_panel = np.logical_and(on_panel_x, on_panel_y)
         return xy[on_panel, :], on_panel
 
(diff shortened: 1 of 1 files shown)


psavery (approved):
LGTM

Issue #299, opened 2021-07-26
We should remove all contents from hexrd.__init__.py


cjh1:
We are using `hexrd` as a [namespace package](http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages) for hexrd and hexrdgui. This may appear to work in dev but these files will not be installed with a system packaging tool ...

joelvbernier:
Seems ok by me.

joelvbernier:
@cjh1 and @psavery -- I don't think we did anything here yet.  Should we proceed?

cjh1:
Yes, we should proceed. Its low risk.

joelvbernier:
Blanking this causes problems firing up the GUI in develop mode, as I pointed out in #317

cjh1:
Resolved by: #322

PR #317, opened 2021-09-21
Remove contents of hexrd/__init__.py

cjh1:
As we are using a namespace package for hexrd and hexrdgui we can't have any code in this file. It will not be installed. It seems to be dead code anyway.

Fixes: #299


joelvbernier:
@cjh1 -- blanking this breaks (at least) trying to run in development mode:
```
  File ".\Documents\GitHub\hexrdgui\hexrd\ui\main.py", line 8, in <module>
    from hexrd.ui import resource_loader
ModuleNotFoundError: No module named 'hexrd.ui'
```
Restoring it allowed me to fire up the GUI again.

cjh1:
@joelvbernier I will take a look

PR #319, opened 2021-09-22
Fix cases when the occupancy or the isotropic Debye waller factors are undefined

saransh13:
Check for cases where occupancy or the Debye Waller factors are present but undefined, most notably by a "?".
The update looks for strings and populates them with default values of 1.0 and 1/2sqrt(2)pi.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -519,8 +519,8 @@ class Material(object):
                 else:
                     occ.append(p)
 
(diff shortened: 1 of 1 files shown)


joelvbernier:
@saransh13 -- did this work with the TATB CIF?

saransh13:
@joelvbernier yep, worked fine with the unedited file.

joelvbernier (approved):

PR #326, opened 2021-09-29
modify expression for Lorentz factor

saransh13:
modify the form of lp correction based on the form in "Two dimensional x-ray diffraaction" by Bob He. Also introduces the ability to pass unpolarized as an argument. This modifies the computation of the polarization factor.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -2214,7 +2214,7 @@ class PlanarDetector(object):
     # METHODS
     # =========================================================================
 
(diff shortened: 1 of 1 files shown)


joelvbernier:
just to be sure -- merging this will break the application in `hexrdgui` until HEXRD/hexrdgui#1053 is merged, yes?

joelvbernier (approved):

saransh13:
No, this will not break anything.

PR #331, opened 2021-10-09
Simulated peak

saransh13:
The function is present in `material.Material.compute_powder_overlay`. The inputs are `ttharray` which has the two theta coordinates from the polar view, `fwhm` which is the full width half maxima of the powder peaks in degrees and `scale` which scales the peaks by a constant factor. Important to note that there could be a DC offset which can make the plot not look great. That can ideally also be estimated if the azimuthal integration is passed. This will fix https://github.com/HEXRD/hexrdgui/issues/1002

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -379,7 +379,7 @@ class Material(object):
         self.planeData.set_structFact(sf)
 
     def compute_powder_overlay(self,
-                               ttharray=numpy.linspace(0,80,2000),
+                               ttharray=numpy.linspace(0, 80, 2000),
                                fwhm=0.25,
                                scale=1.0):
         """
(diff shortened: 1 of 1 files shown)


joelvbernier:
@saransh13 -- have you tested this?  If it works, can you open an issue on hexrdgui to get GUI integration started?

joelvbernier (approved):

hexrd/material.py
```diff
@@ -377,6 +378,26 @@ def update_structure_factor(self):
 
         self.planeData.set_structFact(sf)
 
+    def compute_powder_overlay(self,
+                               ttharray=numpy.linspace(0, 80, 2000),
```
psavery, line 382: It looks like this isn't causing an issue here, but for future reference, it's best practice to not have any mutable default arguments, since it can cause hard to track down bugs. Integers, floats, strings, etc. are all immutable, but a numpy array is mutable. Here's an example:

```python
import numpy as np
  

def example(a=np.linspace(1, 2, 2)):
    return a

out = example()
print(f'{out=}')  # out=array([1., 2.])
out[:] = 0
out2 = example()
print(f'{out2=}')  # out2=array([0., 0.])
```

You can see that since the numpy array was modified, the default argument was also modified.

The better way to do this is to default `ttharray=None`, and then in the function:
```python
if ttharray is None:
    ttharray = numpy.linspace(0, 80, 2000)
```

PR #344, opened 2021-10-20
Use np.nanmax() for getting max powderI

psavery:
Previously, `powderI.max()` was being used, which would return `nan` if
any of the `powderI` values were `nan`. This would result in the rescaled
`powderI` values all being `nan`.

Use `np.nanmax()` instead, which ignores `nan` values when computing the
max. This results in a real value max.

Fixes: hexrd/hexrdgui#1068


psavery:
`tth` is `nan` if it is greater than 180 degrees, I believe by design (right @joelvbernier?). So there will be some nans in the two theta list if some of them are greater than 180 degrees.

PR #353, opened 2021-11-05
Fixing minor typo in periodic table dictionary.

saransh13:
periodic table had incorrect atomic numbers for Cr and Ge (both offset by 1).


psavery (approved):
LGTM

PR #361, opened 2021-12-03
Lebail2d rebased

joelvbernier:


hexrd/instrument.py
```diff
@@ -1298,13 +1298,6 @@ def simulate_powder_pattern(self,
                             self.simulated_spectrum.x,
                             self.simulated_spectrum.y + self.background.y)
 
-            # normalize everything to 0-1
-            mi = img.min()
-            ma = img.max()
-
-            if(ma > mi):
-                img = (img - mi) / (ma - mi)
-
```
psavery, line 1307: We should make sure this won't have any unexpected side effects.

psavery (approved):

hexrd/constants.py
```diff
@@ -743,44 +742,40 @@ def is_writable_file(path):
...
+    'Ru': 44, 'Rh': 45, 'Pd': 46, 'Ag': 47, 'Cd': 48, 'In': 49, 'Sn': 50,
+    'Sb': 51, 'Te': 52, 'I': 53, 'Xe': 54, 'Cs': 55, 'Ba': 56, 'La': 57,
+    'Ce': 58, 'Pr': 59, 'Nd': 60, 'Pm': 61, 'Sm': 62, 'Eu': 63, 'Gd': 64,
+    'Tb': 65, 'Dy': 66, 'Ho': 67, 'Er': 68, 'Tm': 69, 'Yb': 70, 'Lu': 71,
+    'Hf': 72, 'Ta': 73, 'W': 74, 'Re': 75, 'Os': 76, 'Ir': 77, 'Pt': 78,
+    'Au': 79, 'Hg': 80, 'Tl': 81, 'Pb': 82, 'Bi': 83, 'Po': 84, 'At': 85,
+    'Rn': 86, 'Fr': 87, 'Ra': 88, 'Ac': 89, 'Th': 90, 'Pa': 91, 'U': 92,
+    'Np': 93, 'Pu': 94, 'Am': 95, 'Cm': 96, 'Bk': 97, 'Cf': 98, 'Es': 99,
+    'Fm': 100, 'Md': 101, 'No': 102, 'Lr': 103, 'Rf': 104, 'Db': 105,
+    'Sg': 106, 'Bh': 107, 'Hs': 108, 'Mt': 109
```
psavery, line 761: I think it'd be nice if we put some big dictionaries like this in their own file. But maybe another time.

hexrd/instrument.py
joelvbernier: @saransh13 -- I asked you about this and it seemed ok, but let's verify?

hexrd/constants.py
joelvbernier: @psavery: AGREE!

PR #364, opened 2021-12-17
Bump to Python 3.9

cjh1:


joelvbernier:
Hmmm -- looks like the Mac OS package build failed.  Also, will these changes work for python >=3.8?

cjh1:
@joelvbernier Yep, looking at it. If we still want 3.8 support, we will have to build for that version as well ...

psavery:
@cjh1 It looks like it is trying to install specifically `numpy==1.6.6`, which is older and does not have binaries on conda-forge for python 3.9.

psavery:
Note the message:
```
No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.16
WARNING:conda_build.metadata:No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.16
```

joelvbernier:
Maybe make the dependency be `numpy >=1.19`?

joelvbernier:
Must be the last time the c api changed

psavery:
Looks like we'll need to force it to use a newer version of numpy.

cjh1:
@joelvbernier @psavery We are now building packages for 3.8 and 3.9.

.github/workflows/container_build.sh
```diff
@@ -4,7 +4,7 @@ yum install -y wget git gcc
 wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
 chmod a+x Miniconda3-latest-Linux-x86_64.sh
 ./Miniconda3-latest-Linux-x86_64.sh -b
-${HOME}/miniconda3/bin/conda create -y -n hexrd python=3.8
+${HOME}/miniconda3/bin/conda create -y -n hexrd python=3.9
```
psavery, line 7: I'm wondering if we should add `--override-channels -c conda-forge` here to ensure we are using python from conda forge. However, that doesn't need to be done in this PR.

conda.recipe/meta.yaml
```diff
@@ -25,10 +25,10 @@ requirements:
     - h5py
     - lmfit
     - numba
-    - numpy
+    - numpy>=1.20
     - psutil
     - pycifrw
-    - python=3.8
+    - python
```
psavery, line 31: I guess this will just use the same python from the `build` step?

psavery (approved):
LGTM. Note that for python 3.8, we will likely need to upload and use our own python for Mac.

conda.recipe/meta.yaml
cjh1: Correct

.github/workflows/container_build.sh
cjh1: I can add that in to be safe.

cjh1:
> LGTM. Note that for python 3.8, we will likely need to upload and use our own python for Mac.

I think for the GUI we should just build with 3.9, so we don't have to continue support our own build.

psavery:
> > LGTM. Note that for python 3.8, we will likely need to upload and use our own python for Mac.
> 
> I think for the GUI we should just build with 3.9, so we don't have to continue support our own build.

That sounds perfect.

PR #367, opened 2021-12-22
Added list sort to deal with glob

joelvbernier:
List is perhaps long overdue; not sure why we decided to punt on it earlier, but it seems to be the only way to reliably allow the use of glob strings (which is useful!) in the `image-files` adapter for `hexrd.imageseries.load`.

### DIFF
--- a/hexrd/imageseries/load/imagefiles.py
+++ b/hexrd/imageseries/load/imagefiles.py
@@ -2,9 +2,9 @@
 """
 
 
-import sys
+# import sys
 import os
-import logging
+# import logging
 import glob
 
 # # Put this before fabio import and reset level if you
@@ -37,7 +37,7 @@ class ImageFilesImageSeriesAdapter(ImageSeriesAdapter):
         self._load_yml()
         self._process_files()
 
-    #@memoize
+    # @memoize
     def __len__(self):
         if self._maxframes_tot > 0:
             return min(self._nframes, self._maxframes_tot)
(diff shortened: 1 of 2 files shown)


joelvbernier:
I also snuck a change in here that slightly modifies the was that the `PolarView` class calculates the number of bins for 2θ and η.  Specifically, it adds a `round` to make sure that the difference in the max angle is never more than half a pixel size from the requested value.

hexrd/imageseries/load/imagefiles.py
```diff
@@ -95,14 +95,19 @@ def _load_yml(self):
         self._files = []
         for g in fglob.split():
             self._files += glob.glob(os.path.join(dname, g))
-
+        # !!! must sort due to the non-determinate nature of glob on various
+        #     filesystems.  See https://github.com/HEXRD/hexrd/issues/263
+        self._files = self._files.sort()
```
psavery, line 100:
```suggestion 
        self._files.sort()
```

Hi Joel! Sorry I am delayed in reviewing this, I was out on break.

I see my above suggested change was implemented in #369, so that's good.

Also, the sort here changes the output slightly compared to before. This now always sorts the list of files no matter what. So if the user had this, for instance:
```
  files:
    IMAGE-D
    IMAGE-A
```

This would be sorted internally to be `IMAGE-A, IMAGE-D`. Not sure if this matters. If we only wanted to sort the glob rather than the whole list, we could just put `sorted` around the `glob` call like `sorted(glob.glob(...))`.

Otherwise, this looks good!

PR #369, opened 2021-12-23
Update imagefiles.py

joelvbernier:
mistakenly tried to assign the result of `list.sort()` (instead of assigning `sorted(list)`).


psavery (commented):
LGTM

PR #370, opened 2022-01-03
Pin numpy to version 1.20

psavery:
The latest version of numba (0.54.1) requires numpy to be < 1.21. The
latest version of numpy is 1.22, however.

Thus, for both the build environment and the run environment, we need
to downgrade numpy to 1.20, so that it is compatible with numba. Fixing
the version is needed for pip dev environments, because pip is not smart
enough to downgrade numpy for us. Conda, however, does downgrade the numpy
version automatically, which is why we aren't running into the issue with
conda environments.

This should fix the errors occurring on hexrdgui master.

Note: it looks like numba 0.55 is supposed to come out soon, which will
require numpy < 1.22.


cjh1 (approved):
LGTM

PR #372, opened 2022-01-05
Update fitpeak.py

darrencpagan:
Fix long-standing issues with hard coded estimates in mpeak function estimating. Also now can handle vectors for bounding guesses.


joelvbernier (approved):
I'll clean up the PEP8 later.

PR #374, opened 2022-01-05
Use skimage for center of mass instead of scipy

psavery:
It appear's that skimage's `regionprops` followed by `weighted_centroid`
is much faster than scipy's `center_of_mass` for `pull_spots()`.

In one example I had, using `skimage` brought the total time for
`pull_spots()` down from 55.2 seconds to 31.6 seconds (where
skimage took 1.4 seconds compared to scipy's 23.8 seconds).

We should make sure that our results are identical before merging, and
that skimage can handle all cases the same way that scipy did.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -45,6 +45,7 @@ from io import IOBase
 
 from scipy import ndimage
 from scipy.linalg.matfuncs import logm
+from skimage.measure import regionprops
 
 from hexrd import constants
 from hexrd.gridutil import cellConnectivity, cellIndices, make_tolerance_grid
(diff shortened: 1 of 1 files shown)


joelvbernier:
Maybe we should replace the call to `ndimage.label` with the `skimage` equivalent as well:
https://github.com/HEXRD/hexrd/blob/627548b14ff77c4203fe03605923f328d70076e5/hexrd/instrument.py#L1719-L1723

joelvbernier (approved):

psavery:
> Maybe we should replace the call to `ndimage.label` with the `skimage` equivalent as well:
> 
> https://github.com/HEXRD/hexrd/blob/627548b14ff77c4203fe03605923f328d70076e5/hexrd/instrument.py#L1719-L1723

This sounds good. We'll look at it in another PR.

PR #377, opened 2022-01-07
Rotations features

joelvbernier:
Added utility function `expMapOfQuat` as well as some enhanced docstrings.


joelvbernier:
Ah, I forgot that I also added the *sample* frame change of basis matrices -- both (3, 3) and (4, 4) for squats -- that takes the components in the Fable-style sample frame to the hexrd sample frame.  You have to be careful if converting orientations that the same *crystal* frame is used as well (I believe IceNine and Hexomap both use the same one as hexrd, but Fable and Midas use a different one).

PR #388, opened 2022-01-17
Raw image imageseries

donald-e-boyce:
This creates an imageseries adapter for custom file formats (`raw-image` keyword). 

You can create an instance like this:

```ims = imageseries.open(filename, "raw-image")```

The input file is a YAML file describing the scalar data format and basic properties of the images. Here is an example:

```
#
# YAML example for raw image
#
# For scalar definition:
#   "type": i -> int, f -> float, d -> double, b -> bool
#    "bytes" and "signed" are only for int types
#    "bytes": 1, 2, 4, or 8
#    "signed": true | false
#    "endian": use sys.byteorder to determine value for local system
#
filename: RUBY_4537.ge
shape: 2048 2048
skip: 8192
scalar:
  type: i
  bytes: 2
  signed: false
  endian: little

```

### DIFF
--- a/hexrd/imageseries/load/__init__.py
+++ b/hexrd/imageseries/load/__init__.py
@@ -18,7 +18,7 @@ class ImageSeriesAdapter(ImageSeriesABC, metaclass=_RegisterAdapterClass):
 
 # import all adapter modules
 
-from . import array, framecache, hdf5, imagefiles, metadata, trivial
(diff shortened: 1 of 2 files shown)


donald-e-boyce:
I updated the load/__init__.py file correcting PEP 8 issues, but the import after class definitions is reuqired as those modules need the registry class to be defined.

joelvbernier (approved):

PR #389, opened 2022-01-18
Fitting fixes jvb

joelvbernier:
Adding a new 1-d spectrum fitting module for calibration that uses `lmfit`; this was really precipitated by the need to get the DCS pink beam function working.  Constraints are essential.  That use case is passing basic testing, but need to do several things:

- [x] add unit tests for DCS case (already included) and plain CeO2 (maybe using gaussian and pseudo-Voigt)
- [x] make interface class for generic spectrum fitting with 1 or more peaks and choice of background function
- [x] steal/enhance parameter estimator from `fitpeak`

@saransh13 and @darrencpagan -- I am open to any ideas you have regarding enhancing the initial parameter estimation.  What we currently have is _ok_, but might not be robust if the initial guess is a bit further off.  I am aware that there very well may not be a better generic rubric, but I figured I would plant the seed.

### DIFF
--- a/hexrd/fitting/calibration.py
+++ b/hexrd/fitting/calibration.py
@@ -31,11 +31,16 @@ grain_flags_DFLT = np.array(
     dtype=bool
 )
 
+nfields_powder_data = 8
+
+
 # =============================================================================
 # %% POWDER CALIBRATION
 # =============================================================================
 
-nfields_powder_data = 8
+
+def _normalized_ssqr(resd):
+    return np.sum(resd*resd)/len(resd)
 
 
 class PowderCalibrator(object):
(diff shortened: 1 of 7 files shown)


hexrd/fitting/calibration.py
```diff
@@ -4,14 +4,14 @@
 
 from scipy.optimize import leastsq, least_squares
 
+from tqdm import tqdm
```
psavery, line 7: @joelvbernier Should we add `tqdm` as a dependency in the `setup.py` and conda recipe?

psavery (approved):
LGTM after we add tqdm as a dependency...

hexrd/fitting/calibration.py
psavery: This was added in #401.

joelvbernier:
Still needs unit tests, but I've tested it with an array of data including DCS.  Also tested the updated "auto" powder calibration on a Pilatus.

PR #395, opened 2022-01-31
Fix default material tThWidth for HDF5

psavery:
If the `tThWidth` wasn't in the HDF5 file, then the default was being
set to `np.radians(Material.DFLT_TTH)`. However, `Material.DFLT_TTH`
was already in radians.

This would result in the default two theta widths in HexrdGui being
0.004° instead of 0.25°.

Fix this so that the conversion to radians only occurs if the two theta
width is loaded from the HDF5 file.


saransh13 (approved):
LGTM

joelvbernier (approved):

PR #396, opened 2022-01-31
Fixing write error in material

saransh13:
check if tThWidth is present in planeData before trying to convert to degrees. This was leading to an error during state write.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -732,7 +732,7 @@ class Material(object):
         AtomInfo['hkls'] = self.planeData.getHKLs()
         AtomInfo['dmin'] = self.unitcell.dmin
         if self.planeData.tThWidth is None:
-            AtomInfo['tThWidth'] = Material.DFLT_TTH
+            AtomInfo['tThWidth'] = numpy.degrees(Material.DFLT_TTH)
         else:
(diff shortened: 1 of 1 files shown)


hexrd/material.py
```diff
@@ -731,7 +731,10 @@ def dump_material(self, file, path=None):
         AtomInfo['stiffness'] = self.unitcell.stiffness
         AtomInfo['hkls'] = self.planeData.getHKLs()
         AtomInfo['dmin'] = self.unitcell.dmin
-        AtomInfo['tThWidth'] = numpy.degrees(self.planeData.tThWidth)
+        if self.planeData.tThWidth is None:
+            AtomInfo['tThWidth'] = Material.DFLT_TTH
```
psavery, line 735: Isn't `Material.DFLT_TTH` in radians? I think you need to convert it to degrees here.

joelvbernier: @saransh13 -- have you verified?

saransh13:
Yea, tested with GUI master.

hexrd/material.py
psavery: In the two lines below this, the plane data `tThWidth` is being converted to degrees. The `Material.DFLT_TTH` must be converted to degrees as well.

joelvbernier (approved):

PR #397, opened 2022-02-02
Bump numpy version

psavery:
1. Bump maximum numpy version on PyPI to 1.21
    
The latest numba, 0.55, which was released a few weeks ago, is
compatible with numpy 1.21, but not numpy 1.22.
    
Bump the maximum numpy version to 1.21 to reflect this.

2. Add numba to conda build reqs, unfix numpy version

This is to ensure that the numpy version used to build hexrd is
compatible with numba.

We should test this to be sure it works...

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -13,7 +13,9 @@ build:
 
 requirements:
   build:
-    - numpy=1.20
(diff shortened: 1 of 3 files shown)


psavery:
I tested this out locally with one of the examples and it seems to be working.

joelvbernier (approved):

PR #398, opened 2022-02-11
Fix multiprocessing context for ProcessPoolExecutor

joelvbernier:
We need to force the multiprocessing context to 'fork' on platforms (not Windows) where it is available.


joelvbernier:
Fixes #399.

joelvbernier:
I agree we may have to revisit this later -- but this solves my immediate problem.

PR #400, opened 2022-02-11
Check shape of hkl array before write

saransh13:
only write hkl if shape is not (0,). 

Fixes: hexrd/hexrdgui#1154


psavery (approved):
LGTM. I also tested it and this fixed the issue.

PR #401, opened 2022-02-14
add tqdm package dependency

joelvbernier:
The updated calibration package uses this for a status bar, and we should probably use `tqdm` across the code base to report status in loops.  @psavery -- would be great to figure out how to use it to talk to status bars in the GUI.


psavery (approved):
LGTM

PR #403, opened 2022-02-16
catch None in hdf5 write as TypeError

joelvbernier:
This address https://github.com/HEXRD/hexrdgui/issues/1161; now if a `panel_buffer` attribute is `None`, no exception is encoutnered.  However the HDF5 format (.hexrd) config gets written with `panel_buffer = [0., 0.]` instead of omitting it.  I think we should alter that behavior to just omit, which gets caught in the `HEDMInstrument.__init__`.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -3581,8 +3581,11 @@ def unwrap_dict_to_h5(grp, d, asattr=False):
                 try:
                     grp.create_dataset(key, data=np.atleast_1d(item))
                 except(TypeError):
-                    # probably a string badness
-                    grp.create_dataset(key, data=item)
+                    if item is None:
+                        continue
+                    else:
+                        # probably a string badness
+                        grp.create_dataset(key, data=item)
 
 
 def unwrap_h5_to_dict(f, d):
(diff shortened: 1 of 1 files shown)


hexrd/instrument.py
```diff
@@ -3576,13 +3580,20 @@ def unwrap_dict_to_h5(grp, d, asattr=False):
             unwrap_dict_to_h5(subgrp, item, asattr=asattr)
         else:
             if asattr:
-                grp.attrs.create(key, item)
+                try:
+                    grp.attrs.create(key, item)
+                except(TypeError):
+                    if item is None:
+                        continue
```
psavery, line 3587: Could we either check if the item is `None` like the following:
```suggestion
                if item is None:
                    continue
                grp.attrs.create(key, item)
```

Or add an `else: raise` to the end of this (see below)? 

```suggestion
                try:
                    grp.attrs.create(key, item)
                except(TypeError):
                    if item is None:
                        continue
                    else:
                        raise
```

If the `item` is not `None`, then we probably want to propagate the exception so that we see it. If we don't, writing the item will silently fail, and might produce an invalid HDF5 file without us knowing.

Adding a `raise` in exception handling will propagate the exception.

joelvbernier: yes -- I agree with the second and applied it.

psavery (approved):

PR #404, opened 2022-02-16
attribute for picks and logging

joelvbernier:
This is another step in trying to finalize the design for the calibrator classes.  Added attributes to act as a container for calibration points, so you can set them.  Lots of tasks:

- [ ] define container class for calibration point data
- [ ] add consistency checks for calibration data (_i.e._, in the setter)
- [ ] add option to fit interpolated data in the auto-picker vs. always performing interpolation to raw
- [ ] add awareness to any masks
- [x] handle case of flat spectrum data to fit (`ValueError`)

### DIFF
--- a/hexrd/constants.py
+++ b/hexrd/constants.py
@@ -36,6 +36,8 @@ from scipy import constants as sc
 # !!! for stetting mp start method
 if 'Windows' in platform.system():
     mp_context = mp.get_context("spawn")
+elif 'Darwin' in platform.system():
+    mp_context = mp.get_context("spawn")
 else:
     mp_context = mp.get_context("fork")
 
--- a/hexrd/fitting/calibration.py
+++ b/hexrd/fitting/calibration.py
@@ -14,7 +14,9 @@ from hexrd.transforms import xfcapi
 from . import grains as grainutil
 from . import spectrum
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger()
+logger.setLevel('INFO')
+
 
 # =============================================================================
 # %% PARAMETERS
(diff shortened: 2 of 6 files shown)


joelvbernier:
@psavery -- as per our conversation, here is a task list for improving the auto picking (separate from the above):
- [ ] split out the actually peak fitting from `PowderCalibrator._extract_powder_lines` at the level of [fitting individual spectra](https://github.com/HEXRD/hexrd/blob/e88e6df0055663ad470890b94503a74561197788/hexrd/fitting/calibration.py#L272) and make it a standalone function in `utils`
- [ ] condense the calls to `PowderCalibrator._interpolate_images` and `PowderCalibrator._extract_powder_lines` by adding the fitting as an option inside `HEDMInstrument.extract_line_positions`.  This avoided doubly looping over the ring sets and vastly simplifies the call stack in the calibrator.  The fitting can be performed in-line after the interpolation [here](https://github.com/HEXRD/hexrd/blob/e88e6df0055663ad470890b94503a74561197788/hexrd/instrument.py#L1143); this is the case where we have summed the 2-d intensity array interpolated onto the (2θ, η) patch into a 1-d intensity vs. 2θ spectrum.  We can call the imported spectrum fitting function from above and tack on results here.  This means that we may need to give some though to the output data structure as well, I realize...
- [ ] exploit concurrency at the level of the loop over ring sets; [these iterations](https://github.com/HEXRD/hexrd/blob/e88e6df0055663ad470890b94503a74561197788/hexrd/instrument.py#L1073) are all independent and can be performed concurrently.

psavery:
I'm sure there is more we can do to improve the performance of the `HEDMInstrument.extract_line_positions()` function, but what I have so far has improved the performance by ~5x on my computer. Maybe we can take it the way it is and check for more performance improvements in the future if it becomes a high priority?

psavery:
@joelvbernier This is rebased on master and ready for review.

We'll add more improvements to this work in future PRs.

PR #406, opened 2022-02-17
Catch for flat patch data

joelvbernier:
The fitting will fail as currently implemented if the patch/spectrum data is all zeros or otherwise has no discernible peak.  Can you test this for your failing cases @psavery?  Satisfies one of the tasks in #404.

### DIFF
--- a/hexrd/fitting/calibration.py
+++ b/hexrd/fitting/calibration.py
@@ -265,6 +265,8 @@ class PowderCalibrator(object):
                         #     np.array(intensities).squeeze(),
                         #     axis=0
                         # )
+                        if len(intensities) == 0:
+                            continue
                         spec_data = np.vstack(
                             [np.degrees(angs[0]),
                              intensities[0]]
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -1117,6 +1117,14 @@ class HEDMInstrument(object):
                         else:
                             tmp = image[ijs[0], ijs[1]]*area_fac
 
+                        # catch flat spectrum data, which will cause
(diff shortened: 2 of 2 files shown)


psavery:
I gave this a try on my examples, but I get this error now when I run it:

```python
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/async_worker.py", line 73, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/patrick/virtualenvs/hexrd/src/hexrdgui/hexrd/ui/calibration/auto/powder_runner.py", line 95, in run_extract_powder_lines
    self.data_dict = self.ic.extract_points(**kwargs)[0]
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/fitting/calibration.py", line 546, in extract_points
    calib_class._extract_powder_lines(
  File "/home/patrick/virtualenvs/hexrd/src/hexrd/hexrd/fitting/calibration.py", line 270, in _extract_powder_lines
    intensities[0]]
IndexError: list index out of range
```

I guess this might make that intensities list empty if the ring was skipped?

joelvbernier:
Yes, we should catch that.  I'll note that a completely empty ring is *extremely* unlikely, but not impossible for limited/messy data.  Not that you'd try to do a calibration with such data :-)

joelvbernier:
Try that @psavery -- btw, what are you testing with?

psavery (approved):
Works for me! I can run it now without issues.

It was with the GE Ruby example (the one that has spots, even though we aren't supposed to use that kind of data for auto fitting).

PR #415, opened 2022-03-22
Make HDF5 imageseries pickleable

psavery:
This adds the `__getstate__()` and `__setstate__()` functions to the
HDF5 imageseries adapter so that it can be pickled.

This is needed, for instance, when running fit-grains with more than
one grain on a system using "spawn" multiprocessing (such as Windows).
Otherwise, an error will occur.

The logic here mostly follows [the official python example](https://docs.python.org/3/library/pickle.html#handling-stateful-objects).

### DIFF
--- a/hexrd/imageseries/load/hdf5.py
+++ b/hexrd/imageseries/load/hdf5.py
@@ -29,8 +29,7 @@ class HDF5ImageSeriesAdapter(ImageSeriesAdapter):
         self.__path = kwargs['path']
         self.__dataname = kwargs.pop('dataname', 'images')
         self.__images = '/'.join([self.__path, self.__dataname])
-        self.__image_dataset = self.__h5file[self.__images]
-        self.__data_group = self.__h5file[self.__path]
+        self._load_data()
         self._meta = self._getmeta()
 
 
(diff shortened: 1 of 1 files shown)


psavery:
> Hello @psavery! Thanks for opening this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> * In the file [`hexrd/imageseries/load/hdf5.py`](https://github.com/HEXRD/hexrd/blob/4daaa50f7333771897f8e4cacb40b040d8bc3be2/hexrd/imageseries/load/hdf5.py):
> 
> > [Line 68:5](https://github.com/HEXRD/hexrd/blob/4daaa50f7333771897f8e4cacb40b040d8bc3be2/hexrd/imageseries/load/hdf5.py#L68): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 90:5](https://github.com/HEXRD/hexrd/blob/4daaa50f7333771897f8e4cacb40b040d8bc3be2/hexrd/imageseries/load/hdf5.py#L90): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)
> > [Line 96:5](https://github.com/HEXRD/hexrd/blob/4daaa50f7333771897f8e4cacb40b040d8bc3be2/hexrd/imageseries/load/hdf5.py#L96): [E303](https://duckduckgo.com/?q=pep8%20E303) too many blank lines (2)

The spacing I added follows the spacing used in the rest of the file, so I think we can ignore this.

PR #418, opened 2022-03-25
added logic to handle omega gaps

joelvbernier:
Currently, the η-ω map generation could not handle multi-wedge imageseries.  This PR addresses that shortcoming.  There are a few tacit assumptions about the imageseries, however:
1. the wedges must be non-overlapping
2. the wedges must not exceed 360º
3. the wedges, and thus frame order, must be monotonically increasing in ω
I will attach a test case form a DAC experiment.

### DIFF
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -3659,6 +3659,8 @@ class GenerateEtaOmeMaps(object):
         image_series must be OmegaImageSeries class
         instrument_params must be a dict (loaded from yaml spec)
         active_hkls must be a list (required for now)
+
+        FIXME: get rid of omega period; should get it from imageseries
         """
 
         self._planeData = plane_data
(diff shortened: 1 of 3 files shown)


hexrd/instrument.py
```diff
@@ -3672,18 +3674,66 @@ def __init__(self, image_series_dict, instrument, plane_data,
             self._iHKLList = active_hkls
             n_rings = len(active_hkls)
 
+        # grab a det key and corresponding imageseries (first will do)
+        # !!! assuming that the imageseries for all panels
+        #     have the same length and omegas
+        det_key = list(eta_mapping.keys())[0]
```
psavery, line 3680: The `eta_mapping` variable gets created via `instrument.extract_polar_maps()` below. Maybe you need to move that call to before this line?

hexrd/instrument.py
```diff
@@ -3672,18 +3674,66 @@ def __init__(self, image_series_dict, instrument, plane_data,
...
+        this_det_ims = image_series_dict[det_key]
+
+        # handle omegas
+        # !!! for multi wedge, enforncing monotonicity
+        # !!! wedges also cannot overlap or span more than 360
+        omegas_array = this_det_ims.metadata['omega']  # !!! DEGREES
+        frame_mask = None
+        ome_period = omegas_array[0, 0] + 360.  # !!! be careful
```
psavery, line 3688: Is the official way to get the period via the `hexrd.findorientations._process_omegas()` function being called on the imageseries dict? Maybe we can do that?

hexrd/instrument.py
joelvbernier: yep.

hexrd/rotations.py
```diff
@@ -790,7 +795,7 @@ def angleAxisOfRotMat(R):
     #  *   near pi   -- binary rotation; need to find axis
     #  *   neither   -- general case; can use skew part
     #
-    tol = cnst.sqrt_epsf
+    tol = cnst.epsf
```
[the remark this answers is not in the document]
joelvbernier: I clobbered it with master's

PR #420, opened 2022-03-28
add roi handling

### LINKED ISSUE #419: Add ROI attribute to detector class to address multi-element detector analysis.
By enabling an ROI to be specified as a property on each detector class, we can both avoid splitting data for multi-element detectors (_e.g._, Pilatus), as well as enable the handling of novel acquisition schemes.  in this implementation, I think it is easiest to 

1. Enable `None` and/or `'all'` as `panel` specifier in the `image_series` block of the configuration; this would require that we address how the image data is passed to routines that are expecting a dictionary of imageseries over the detector keys

2. Add checks inside detector loops to see `if detector.roi is None`, `None` being the default value in the detector init.

3. Will have to modify the panel dimensions if ROI is set

joelvbernier:
Finally adding the ability to specify ROI in a detector.  The usage is that any imageseries with an unspecified detector panel will be treated as "shared", and the detectors not represented in an image dictionary input will use the ROI to grab its pixel data.  The obvious application is for multi element detectors.  Closes #419. 
- [x] handle case where there is at most 1 shared imageseries
- [ ] generalize to multiple shared imageseries
  - requires explicitly passing a list of detectors over which to share

### DIFF
--- a/hexrd/config/imageseries.py
+++ b/hexrd/config/imageseries.py
@@ -6,6 +6,7 @@ class ImageSeries(Config):
     """
     """
     BASEKEY = 'image_series'
+    share_ims_key = 'mulitple'
 
     def __init__(self, cfg):
         super(ImageSeries, self).__init__(cfg)
@@ -17,7 +18,7 @@ class ImageSeries(Config):
 
     @property
     def imageseries(self):
-        """Return the imageseries dictionar.y"""
+        """Return the imageseries dictionary"""
         if self._image_dict is None:
             self._image_dict = dict()
             fmt = self.format
@@ -26,7 +27,12 @@ class ImageSeries(Config):
                 args = ispec['args']
                 ims = imageseries.open(fname, fmt, **args)
                 oms = imageseries.omega.OmegaImageSeries(ims)
-                panel = oms.metadata['panel']
+                # handle special case for single IMS
+                # for use with ROI
+                try:
+                    panel = oms.metadata['panel']
+                except(KeyError):
+                    panel = 'share_ims_key'
                 self._image_dict[panel] = oms
 
         return self._image_dict
--- a/hexrd/config/imageseries.py
+++ b/hexrd/config/imageseries.py
@@ -1,12 +1,11 @@
 from .config import Config
 from hexrd import imageseries
 
+from hexrd.constants import shared_ims_key
+
 
 class ImageSeries(Config):
-    """
-    """
     BASEKEY = 'image_series'
-    share_ims_key = 'mulitple'
 
     def __init__(self, cfg):
         super(ImageSeries, self).__init__(cfg)
@@ -31,8 +30,12 @@ class ImageSeries(Config):
                 # for use with ROI
                 try:
                     panel = oms.metadata['panel']
+                    if isinstance(panel, (tuple, list)):
+                        panel = '_'.join(panel)
+                    elif panel is None:
+                        panel = shared_ims_key
                 except(KeyError):
-                    panel = 'share_ims_key'
+                    panel = shared_ims_key
                 self._image_dict[panel] = oms
 
         return self._image_dict
--- a/hexrd/fitgrains.py
+++ b/hexrd/fitgrains.py
@@ -238,7 +238,7 @@ def fit_grain_FF_reduced(grain_id):
                 culled_results_r[det_key] = []
                 continue
 
-            ims = imgser_dict[det_key]
+            ims = next(iter(imgser_dict.values()))  # grab first for the omes
             ome_step = sum(np.r_[-1, 1]*ims.metadata['omega'][0, :])
 
             xyo_det = np.atleast_2d(
--- a/hexrd/instrument.py
+++ b/hexrd/instrument.py
@@ -330,8 +330,10 @@ def _parse_imgser_dict(imgser_dict, det_key, roi=None):
                 raise RuntimeError(
                     f"multiple entries found for '{det_key}'"
                 )
-            img_keys = list(imgser_dict.keys())  # same order
-            matched_det_key = img_keys[matched_det_keys]
+            # use boolean array to index the proper key
+            # !!! these should be in the same order
+            img_keys = img_keys = np.asarray(list(imgser_dict.keys()))
+            matched_det_key = img_keys[matched_det_keys][0]  # !!! only one
             images_in = imgser_dict[matched_det_key]
         else:
             raise RuntimeError(
@@ -2558,10 +2560,13 @@ class PlanarDetector(object):
         # assign local vars; listify if necessary
         tilt = self.tilt
         translation = self.tvec
+        roi = None if self.roi is None \
+            else np.array([self.roi[0][0], self.roi[1][0]]).flatten()
         if style.lower() == 'yaml':
             tilt = tilt.tolist()
             translation = translation.tolist()
             tvec = tvec.tolist()
+            roi = None if roi is None else roi.tolist()
 
         det_dict = dict(
             transform=dict(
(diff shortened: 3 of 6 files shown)


hexrd/instrument.py
```diff
@@ -165,6 +171,77 @@ def _fix_indices(idx, lo, hi):
...
+    try:
+        ims = imgser_dict[det_key]
+    except(KeyError):
+        if multi_ims_key in imgser_dict:
+            if roi is None:
+                raise RuntimeError(
+                    "roi must be specified to use shared imageseries"
+                )
+            images_in = imgser_dict[multi_ims_key]
+            if isinstance(images_in, ims_classes):
```
psavery, line 213:
```suggestion
            if isinstance(images_in, ImageSeries):
```

I believe the above will work because `ProcessedImageSeries` and `OmegaImageSeries` both subclass `ImageSeries`, which means that they will return `True` for `isinstance(images_in, ImageSeries)`. So I think we can remove the `ims_classes` variable.

hexrd/config/imageseries.py
```diff
@@ -26,7 +27,12 @@ def imageseries(self):
                 args = ispec['args']
                 ims = imageseries.open(fname, fmt, **args)
                 oms = imageseries.omega.OmegaImageSeries(ims)
-                panel = oms.metadata['panel']
+                # handle special case for single IMS
+                # for use with ROI
+                try:
+                    panel = oms.metadata['panel']
+                except(KeyError):
+                    panel = 'share_ims_key'
```
psavery, line 35:
```suggestion
                    panel = self.share_ims_key
```

hexrd/config/imageseries.py
```diff
@@ -6,6 +6,7 @@ class ImageSeries(Config):
     """
     """
     BASEKEY = 'image_series'
+    share_ims_key = 'mulitple'
```
psavery, line 9:
```suggestion
    share_ims_key = 'multiple'
```

hexrd/config/imageseries.py
```diff
@@ -26,7 +27,12 @@ def imageseries(self):
                 args = ispec['args']
                 ims = imageseries.open(fname, fmt, **args)
                 oms = imageseries.omega.OmegaImageSeries(ims)
-                panel = oms.metadata['panel']
+                # handle special case for single IMS
+                # for use with ROI
+                try:
+                    panel = oms.metadata['panel']
+                except(KeyError):
```
psavery, line 34: Note that if we ever run into a bug where `'panel'` is supposed to be there but is not, we won't get an error, but the code will instead assume it is a multi-element detector. This can make the bug harder to track down. But we can keep this as is if you prefer.

hexrd/instrument.py
```diff
@@ -112,6 +115,9 @@
 
 distortion_registry = distortion_pkg.Registry()
 
+multi_ims_key = 'mulitple'
```
psavery, line 118: Maybe we can make this key appear in one place, such as constants, so that we don't have to repeat it twice? It is both here and in the `hexrd.config.imageseries` file.

hexrd/instrument.py
```diff
@@ -474,6 +554,8 @@ def __init__(self, instrument_config=None,
                             raise RuntimeError(
                                 "panel buffer spec invalid for %s" % det_id
                             )
+                # optional roi
+                roi = None if 'roi' not in det_info else det_info['roi']
```
psavery, line 558:
```suggestion
                roi = det_info.get('roi')
```
This is the same

hexrd/instrument.py
```diff
@@ -112,6 +115,9 @@
 
 distortion_registry = distortion_pkg.Registry()
 
+multi_ims_key = 'mulitple'
+ims_classes = (ImageSeries, ProcessedImageSeries, OmegaImageSeries)
```
psavery, line 119: I think we can delete this. See note below.

hexrd/instrument.py
```diff
@@ -165,6 +171,77 @@ def _fix_indices(idx, lo, hi):
...
+                    ims = images_in[roi[0][0]:roi[0][1], roi[1][0]:roi[1][1]]
+                elif ndim == 3:
+                    nrows = roi[0][1] - roi[0][0]
+                    ncols = roi[1][1] - roi[1][0]
+                    n_images = len(images_in)
+                    ims = np.empty((n_images, nrows, ncols),
+                                   dtype=images_in.dtype)
+                    for i, image in images_in:
+                        ims[i, :, :] = \
+                            images_in[roi[0][0]:roi[0][1], roi[1][0]:roi[1][1]]
```
psavery, line 232: Nice!

joelvbernier:
Ok!  The HEDM workflow tests out for both a single shared detector _and_ multiple shared detectors.  example config files attached.
[single_shared_test.zip](https://github.com/HEXRD/hexrd/files/8382740/single_shared_test.zip)
[multiple_shared_test.zip](https://github.com/HEXRD/hexrd/files/8382747/multiple_shared_test.zip)

Shall we punt on getting this into the GUI for now?

hexrd/instrument.py
```diff
@@ -165,6 +171,209 @@ def _fix_indices(idx, lo, hi):
...
+        ims = imgser_dict[det_key]
+    except(KeyError):
+        matched_det_keys = [det_key in k for k in imgser_dict]
+        if multi_ims_key in imgser_dict:
+            images_in = imgser_dict[multi_ims_key]
+        elif np.any(matched_det_keys):
+            if sum(matched_det_keys) != 1:
+                raise RuntimeError(
+                    f"multiple entries found for '{det_key}'"
+                )
```
psavery, line 332: Nice validation

hexrd/instrument.py
```diff
@@ -3729,13 +3964,13 @@ def __init__(self, image_series_dict, instrument, plane_data,
...
-        
+
         # pack all detectors with masking
         # FIXME: add omega masking
         data_store = []
         for i_ring in range(n_rings):
-            # for convenience
-            map_shape = eta_mapping[det_key][i_ring].shape
+            # for convenience grab map shape from first
+            map_shape = next(iter(eta_mapping.values())).shape[1:]
```
psavery, line 3973: This can be outside of the loop if we just grab the first one, right?

psavery (approved):
Examples appear to be working for me

hexrd/instrument.py
joelvbernier: ah yes, you are correct!

hexrd/instrument.py
joelvbernier: We could -- I still need the `OmegaImageSeries` for the subsequent check though.

hexrd/instrument.py
joelvbernier: You are right; I'll make the edit.

hexrd/instrument.py
joelvbernier: Man, how did I not know this?

joelvbernier: will address in #421

hexrd/config/imageseries.py
joelvbernier: fixed below.

hexrd/fitgrains.py
```diff
@@ -238,7 +238,7 @@ def fit_grain_FF_reduced(grain_id):
                 culled_results_r[det_key] = []
                 continue
 
-            ims = imgser_dict[det_key]
+            ims = next(iter(imgser_dict.values()))  # grab first for the omes
```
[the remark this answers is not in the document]
joelvbernier: yep -- made the change in PR #421

PR #421, opened 2022-03-30
Update instrument.py

joelvbernier:
In response to @psavery's comment on PR #420


psavery (approved):

joelvbernier:
catching up with some unresolved edits from #420

PR #434, opened 2022-04-21
Reconcile fitting fixes

joelvbernier:
Making a draft PR to figure out what wasn't pushed.

### DIFF
--- a/hexrd/fitting/calibration.py
+++ b/hexrd/fitting/calibration.py
@@ -14,7 +14,9 @@ from hexrd.transforms import xfcapi
 from . import grains as grainutil
 from . import spectrum
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger()
+logger.setLevel('INFO')
+
 
 # =============================================================================
 # %% PARAMETERS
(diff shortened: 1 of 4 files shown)


joelvbernier:
Actually, it looks like there were important fixes here, but it doesn't include your multiprocessing @psavery.

In calibration, we need to figure out what to do with the calibration points ("picks table").  I had made it a class property so it would be easy for the class to see if it had been initialized to something other than `None` to avoid recalculation; I did, however, make the line extraction return `None`.  Various APIs have changed accordingly.  Do we need it to return the picks as well?  SInce we are going to change the container for the picks anyhow, this may be moot...

psavery:
@joelvbernier I have confirmed that there is no difference between this branch and the `fitting-fixes-jvb` branch other than:

1. The `fitting-fixes-jvb` branch was rebased on master at a different time (and therefore has different commit shas)
2. The `fitting-fixes-jvb` branch has my extra commits on top of it

Therefore, I think we can close this PR and proceed to merge `fitting-fixes-jvb`.

PR #437, opened 2022-04-29
Plane data updates

joelvbernier:
The `PlaneData` class is in need of some cleanup.  A few important features include 

- [x] Make the `getHKLID` method work for any symmetrically equivalent (h, k ,l) input
- [ ] Docstrings
- [ ] fix the use of `__` attributes and old-style property specificaltion
- [x] fix PEP8 warnings

There may be more...  @saransh13?

### DIFF
--- a/hexrd/crystallography.py
+++ b/hexrd/crystallography.py
@@ -724,7 +724,7 @@ class PlaneData(object):
             self.tThSort = np.arange(len(hklDataList))
             self.tThSortInv = np.arange(len(hklDataList))
             self.hklDataList = hklDataList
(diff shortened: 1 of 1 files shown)


hexrd/crystallography.py
```diff
@@ -52,7 +52,7 @@
 
 
 def hklToStr(x):
-    return re.sub('\[|\]|\(|\)', '', str(x))
+    return re.sub(r'[\[\]\(\)\{\},]', '', str(x))
```
psavery, line 55: Does this change anything?

psavery (approved):

PR #438, opened 2022-05-02
Material load

saransh13:
Changing behavior of material load. the beam energy is written to the h5 file now. changes summarized below:

1. If values for `dmin` and `kev` are passed, then they supersede values in the file or the default values.
2. If values are not passed, then file is checked for values of `dmin` and `kev`. If found, then these values are used.
3. If nothing is passed and values not present in file, then the default values are used.

Tested with GUI, with both old material files and new ones (with kev)

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -195,6 +195,11 @@ class Material(object):
             elif isinstance(material_file, h5py.Group) or form in h5_suffixes:
                 self._readHDFxtal(fhdf=material_file, xtal=name)
 
+            if not dmin is None:
+                self._dmin = dmin
+
+            if not kev is None:
+                self._beamEnergy = kev
         else:
             # default name
             self._name = Material.DFLT_XTAL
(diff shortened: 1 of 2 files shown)


psavery (commented):
@saransh13 Overally, I think this PR looks good.

I just pushed a commit that I think re-organizes things a little so it is easier to read. Can you let me know if you approve of the changes? You can see the diff [here](https://github.com/HEXRD/hexrd/pull/438/commits/aad77fd5cde15df77b0b812f2dc1a5bb99bd635f).

joelvbernier (approved):

saransh13:
I think lines 196-201 should not be omitted. The values are read from the
file in the readHDF and readCIF routines, so the values passed by the user
will be overwritten otherwise. The opted option is to use a flag to stop
the read in those routines.

psavery:
> I think lines 196-201 should not be omitted. The values are read from the file in the readHDF and readCIF routines, so the values passed by the user will be overwritten otherwise. The opted option is to use a flag to stop the read in those routines.

Those were actually just moved outside of the code block below, so they are still being used. I added a comment as well to clarify that they are overriding previously set values.

PR #440, opened 2022-05-11
restore functionality with array-like tth input

### LINKED ISSUE #441: Broken functionality in `instrument.HEDMInstrument.extract_line_positions()`
This merge of `fitting-fixes-jvb` broke the ability to use `instrument.HEDMInstrument.extract_line_positions()` with array-like input for the plane_data arg.  It may also have broken some of the other options specified by the kwargs, mainly the case where `do_fitting=False`.

joelvbernier:
A recent merge broke the ability to use `instrument.HEDMInstrument.extract_line_positions()` with array-like input for the `plane_data` arg.  This PR restores that functionality.  Some default values are also updated, noting that the 2theta tolerance is now in % difference, not absolute deg.

closes #441

### DIFF
--- a/hexrd/fitting/calibration.py
+++ b/hexrd/fitting/calibration.py
@@ -235,7 +235,7 @@ class PowderCalibrator(object):
             "pktype '%s' not understood"
         self._bgtype = x
 
-    def _extract_powder_lines(self, fit_tth_tol=None, int_cutoff=1e-4):
+    def _extract_powder_lines(self, fit_tth_tol=5., int_cutoff=1e-4):
         """
         return the RHS for the instrument DOF and image dict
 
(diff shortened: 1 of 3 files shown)


joelvbernier:
Checks out on Linux; seems to fail for `mp_context=spawn` however.

joelvbernier:
@psavery -- confirmed that the attached problem fails when I try to run under the `spawn` context on Mac and WIndows.  Runs fine under Linux/`fork`.
[DCS_powder_cal_example.zip](https://github.com/HEXRD/hexrd/files/8691117/DCS_powder_cal_example.zip)

psavery:
I think you just need to change the last line in your script to:

```python
if __name__ == '__main__':
    multiprocessing.freeze_support()
    pwdr_lines = instr.extract_line_positions(**kwargs)
```

See [#441](https://github.com/HEXRD/hexrd/issues/441#issuecomment-1128200901) for more details.

psavery (approved):
LGTM. It seems that the spawn multiprocessing issue you are encountering is to be fixed in the script, rather than in this PR.

joelvbernier:
This fix won't work if attempting to run in a IDE like Spyder in the `spawn` context; any ideas on how to fix that use case?

PR #443, opened 2022-05-23
add pinhole util module

joelvbernier:
This module will house the utilities needed for applying pinhole camera corrections to apparent Bragg angle.


psavery (approved):
LGTM

PR #449, opened 2022-06-14
Fix use of collections.abc.Iterator

cjh1:
It has been removed in 3.10, use collections.abc.Iterator instead.

Signed-off-by: Chris Harris <chris.harris@kitware.com>


psavery (approved):
LGTM

PR #450, opened 2022-06-17
Remove duplicate atom sync

saransh13:
sync `Material.atominfo`, `Material.atomtype` and `Material.charge` to the updated values after `Material.remove_duplicate_atoms` is called.


psavery (approved):
LGTM

PR #455, opened 2022-08-06
Update spectrum fitting utils to support wildcards in parameter name matching

joelvbernier:
Enhance parameter name matching to take wildcards.

### DIFF
--- a/hexrd/fitting/utils.py
+++ b/hexrd/fitting/utils.py
@@ -1,3 +1,5 @@
+import fnmatch
+
 import numpy as np
 
 from hexrd.constants import (
@@ -17,7 +19,8 @@ def _parameter_arg_constructor(pdict, pargs):
 
 
 def _extract_parameters_by_name(params, pname_root):
-    return [s for s in params.keys() if s.__contains__(pname_root)]
+    # return [s for s in params.keys() if s.__contains__(pname_root)]
+    return fnmatch.filter(list(params.keys()), pname_pattern)
 
 
 def _set_refinement_by_name(params, pname_root, vary=True):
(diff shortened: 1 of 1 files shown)


joelvbernier:
This needs a little more work, since it will _*only*_ supports wildcard strings now...  either I live with it or find a way to preserve (or choose by kwarg?) the old functionality using `__contains__` where the user could specify any substring w/o wildcards.

Any ideas here @psavery or @cjh1 or @donald-e-boyce ?

psavery (approved):
LGTM

hexrd/fitting/utils.py
```diff
@@ -17,7 +19,8 @@ def _parameter_arg_constructor(pdict, pargs):
 
 
 def _extract_parameters_by_name(params, pname_root):
-    return [s for s in params.keys() if s.__contains__(pname_root)]
+    # return [s for s in params.keys() if s.__contains__(pname_root)]
+    return fnmatch.filter(list(params.keys()), pname_pattern)
```
psavery, line 23: @joelvbernier If you wanted to continue supporting substring matches, one option would be to just add `*` to the ends of the pattern, like so:

`f'*{pname_pattern}*'`

Then substrings would still match, and you can do `*` inside the string as well.

Alternatively, you could add an argument such as `match_substring=True` where, if it is true, then you add the `*` to the ends of the pattern.

joelvbernier: of course -- bracketing with '*' might be the most elegant solution :-)

joelvbernier:
@psavery -- I think the previous commit is providing the behavior we want!  Nice one!

psavery (approved):
LGTM

PR #456, opened 2022-08-08
force cast imput to ndarray

joelvbernier:
Got a weird error in the GUI when running repeated "Fast Powder" calibrations with the GE distortion function enabled.
image

When I turned off the Numba implementation to debug, I got an error indicating that the input coordinates were coming in as a `list` rather than a `numpy.ndarray`.  Force-casting the input seems to have solved the problem.

@psavery -- strange I was only getting this in the GUI, specifically when running the cal more than once in the same session.  I suppose the overlay plotting uses the distortion when active too?

### DIFF
--- a/hexrd/distortion/ge_41rt.py
+++ b/hexrd/distortion/ge_41rt.py
@@ -210,6 +210,7 @@ class GE_41RT(DistortionABC, metaclass=_RegisterDistortionClass):
         if self.is_trivial:
             return xy_in
         else:
+            xy_in = np.asarray(xy_in, dtype=float)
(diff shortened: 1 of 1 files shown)


psavery:
@joelvbernier Looks to me that we are passing it in as a list from [here](https://github.com/HEXRD/hexrdgui/blob/a3e45bc324d14560cf574cde5bdb8f28dffd803d/hexrd/ui/calibration/auto/powder_runner.py#L285). I think casting it to a numpy array in the distortion function is a great solution, as it will prevent this error from happening in other places as well in case we ever pass in a list.

psavery (approved):

PR #465, opened 2022-08-24
replacing "if" with "elif" in line 604 of materials.py. 

saransh13:
Was messing up a case with positive ions in the crystal.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -601,7 +601,7 @@ class Material(object):
                 c = s[-2:]
                 if c[0] == '+':
                     c = c[::-1]
-            if "-" in s:
+            elif "-" in s:
                 ss = s[:-2]
                 c = s[-2:]
                 if c[0] == '-':


cjh1 (approved):

saransh13:
@cjh1 Looks like some checks are failing.

cjh1:
@saransh13 I think this is because of the updates that @joelvbernier made to [examples](https://github.com/HEXRD/examples) for #454. I am guess it will be fixed when #454 is merged, @joelvbernier is that going to happen soon? Remember we don't have to cut 0.9.0 straight away. It would be good to have CI working though.

saransh13:
Merging this since it only affects `material.py` and fixes a critical cif read bug for charged species.

joelvbernier:
I'm may roll back that commit to examples until we are ready to merge both hexrd and hexrdgui's PR's with the new API for the `EtaOmegaMaps`

PR #468, opened 2022-08-31
change default source_distance to inf

joelvbernier:
Was `None`, which would be problematic for serialization to HDF5.


psavery (approved):
LGTM

PR #469, opened 2022-09-01
Nf chunking

rachelelim:
Fixed chunking with no leftovers

### DIFF
--- a/scripts/process_nf.py
+++ b/scripts/process_nf.py
@@ -231,30 +231,15 @@ bits_for_arrays = 64*n_oris*n_voxels + 192 * \
 bytes_for_array = bits_for_arrays/8.
 
 n_groups = np.floor(bytes_for_array/RAM_to_use)  # number of full groups
+leftover_voxels = np.mod(n_voxels, n_groups)
 
-<<<<<<< HEAD
 print('Splitting data into %d groups with %d leftover voxels' %
       (int(n_groups), int(leftover_voxels)))
-=======
-voxels_per_group = np.floor(n_voxels/n_groups)
-leftover_voxels = n_voxels - (voxels_per_group * n_groups)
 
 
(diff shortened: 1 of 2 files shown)


joelvbernier:
is this a dupicate PR to #461 @rachelelim ?

scripts/process_nf.py
```diff
@@ -127,6 +126,8 @@
 beam_stop_parms=np.array([beam_stop_y_cen,beam_stop_width])
 
 
+max_RAM = 256 #max amount of memory to available in GB
```
joelvbernier, line 129: What is this hardcoded value?  Don't like it, as it is generally bad practice.  Python has utils for finding available memory:
```
import psutil

mem_info = psutil.virtual_memory()

# to get available memory in bytes, do:
# mem_info.available

```

scripts/process_nf.py
```diff
@@ -207,7 +208,36 @@
                                               process_args=process_args,threshold=img_threshold,ome_dilation_iter=ome_dilation_iter,num_digits=6,stem=stem)
 
     
-    
+#==============================================================================
+# %% NEAR FIELD - splitting
+#==============================================================================
+
+max_RAM = 1  # in GB
```
joelvbernier, line 215: See above.  Not good to hardcode things when you can ask the system for available memory.  Available means not having to use disk.

joelvbernier:
@rachelelim 

your tasks to merge this:

- [ ] use `psutil` to set memory chunks (I'd use `available_memory` from the `psutil.virtual_memory()` object)
- [ ] fix the PEP8 issues.  You can use Spyder (version >= 5) to auto-pep8 your code.

Screen Shot 2022-09-07 at 9 17 36 AM

joelvbernier (approved):

PR #477, opened 2022-09-26
Added ability to import HDF5 imageseries with 2-d data

joelvbernier:
Previously, the HDF5 adapter was tacitly assuming the HDF5 dataset specified was 3-d; this automatically handles the case of 2-d data by stacking a 2-d array in the zeroth (left-most) axis and raises an exception if the specified dataset is neither 2-d or 3-d.

### DIFF
--- a/hexrd/imageseries/load/hdf5.py
+++ b/hexrd/imageseries/load/hdf5.py
@@ -107,7 +107,7 @@ class HDF5ImageSeriesAdapter(ImageSeriesAdapter):
     def _load_data(self):
         data = np.asarray(self.__h5file[self.__images])
         if data.ndim == 2:
-            self.__image_dataset = np.dstack(data)
+            self.__image_dataset = np.asarray([data, ])
         elif data.ndim == 3:
             self.__image_dataset = data
         else:
--- a/hexrd/imageseries/load/hdf5.py
+++ b/hexrd/imageseries/load/hdf5.py
@@ -15,12 +15,29 @@ class HDF5ImageSeriesAdapter(ImageSeriesAdapter):
     format = 'hdf5'
 
     def __init__(self, fname, **kwargs):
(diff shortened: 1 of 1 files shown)


hexrd/imageseries/load/hdf5.py
```diff
@@ -86,26 +99,30 @@ def __getstate__(self):
...
         self._load_data()
 
-
     def _load_data(self):
-        self.__image_dataset = self.__h5file[self.__images]
+        data = np.asarray(self.__h5file[self.__images])
+        if data.ndim == 2:
+            self.__image_dataset = np.asarray([data, ])
+        elif data.ndim == 3:
+            self.__image_dataset = data
```
psavery, line 112: So, the previous behavior was to not load the HDF5 dataset into memory, but to instead load arrays from the HDF5 file as they are requested.

Putting `np.asarray()` around the HDF5 dataset will load it all into memory. Is this potentially a problem? Do we work with some HDF5 datasets that are too large to load into memory?

If it is a problem, maybe we can change the behavior to only load 2D datasets into memory? For instance:

```python
        data = self.__h5file[self.__images]
        if data.ndim == 2:
            self.__image_dataset = np.asarray([data, ])
        elif data.ndim == 3:
            self.__image_dataset = data
```

This would mean that if `data.ndim` was `2`, then `self.__image_dataset` will be a numpy array loaded into memory (hopefully, since the data was just a 2D array, it would be small), and if `data.ndim` was `3`, then `self.__image_dataset` would be an HDF5 dataset *not* loaded into memory.

joelvbernier: ah -- good point @psavery.  Damn, that would be *BAAAD* for large imageseries.  I will fix it and maybe rebase this commit away...

PR #485, opened 2022-11-10
tolerancing for detecting binary rotation was failing

joelvbernier:
I ran into a case really close (angle was within ~2e-8 of pi) to a binary rotation that was not triggering the special axis calculation in `rotations.angleAxisOfRotMat`.  I adjusted the tolerance, but it could use a closer look @donald-e-boyce.

Two questions:

1. is the case of small angle sufficiently handled so as not to clobber rotations for which the axis is still calculable?
2. is the tolerance of 1e-6 the best choice for detecting the binary rotation?  Using 1e-8 wasn't quite working...

We really need unit tests for the rotations module...

### DIFF
--- a/hexrd/rotations.py
+++ b/hexrd/rotations.py
@@ -782,7 +782,7 @@ def angleAxisOfRotMat(R):
     #  *   near pi   -- binary rotation; need to find axis
     #  *   neither   -- general case; can use skew part
     #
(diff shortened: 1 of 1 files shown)


donald-e-boyce (commented):
Looks good.

donald-e-boyce (approved):
Looks good.

Issue #487, opened 2022-11-28
Uniform sampling of orientation space for given point group


saransh13:
Uniformly sample the fundamental zone of any point group with a given sampling step size.

donald-e-boyce:
What exactly do you mean by this? 

It's straightforward to generate a random sample on a uniform distribution on orientation space. Then by putting it into the fundamental region of that point group, you can get a random sample of uniform distribution on that point group.

However, if you are looking for some kind of regular grid with equal spacing in each direction, then I'm not sure that is possible.

saransh13:
By uniform, I meant that the number of points sampled for a given volume in orientation space is constant i.e. constant sampling density. I was going to implement some of my grad school work for hexrd [S. Singh and M. De Graef, "Orientation sampling for dictionary-based diffraction pattern indexing methods". MSMSE 24, 085013 (2016)](https://doi.org/10.1088/0965-0393/24/8/085013)

The implementation uses a different representation of orientations, and the sampling is on a uniform grid. But the property of the space is such that it's constant volume but not constant angular distance. So the sampling is uniform in the density sense, but not in the angular sense.

PR #490, opened 2023-01-03
Sample Rodrigues Fundamental Zone

saransh13:
This is a fix for issue #487 
Tested against the SampleRFZ package (https://github.com/marcdegraef/sampleRFZ) for 6, 622, 6mm, 4mm, 23, m3m representing cyclic, dihedral and cubic FZs.


hexrd/sampleOrientations/conversions.py
```diff
@@ -0,0 +1,152 @@
...
+    for ii in range(2, 21):
+        hm = hm*hmag
+        s = s + constants.tfit[ii] * hm
+    s = 2.0 * np.arccos(s)
+    diff = np.abs(s - np.pi)
+    if diff < 1E-8:
+        return np.array([hn[0], hn[1], hn[2], np.pi])
+    else:
+        return np.array([hn[0], hn[1], hn[2], s])
+    return ax
```
psavery, line 115:
```suggestion
```

It looks to me that this `return ax` is never reached.

hexrd/sampleOrientations/sampleRFZ.py
```diff
@@ -0,0 +1,162 @@
...
+from hexrd.sampleOrientations.conversions import cu2ro, ro2qu
+from hexrd.sampleOrientations.rfz import insideFZ
+from hexrd import constants
+
+if constants.USE_NUMBA:
+    from numba import prange
+else:
+    prange = range
+
+@numba_njit_if_available(cache=True, nogil=True)
```
psavery, line 12:
```suggestion
@numba_njit_if_available(cache=True, nogil=True, parallel=True)
```
I believe that in order to use `prange`, you must pass `parallel=True`.

However, it is possible that the parallel version of this could be slower than the serial version, since some of the numpy functions might already be parallel underneath. So if this does get changed, be sure to test it and see the performance difference.

hexrd/sampleOrientations/sampleRFZ.py
```diff
@@ -0,0 +1,162 @@
...
+    def avg_ang_spacing(self):
+        return self._avg_ang_spacing
+
+
+    @avg_ang_spacing.setter
+    def avg_ang_spacing(self, ang):
+        self._avg_ang_spacing = ang
+        if hasattr(self, 'pgnum'):
+            if hasattr(self, 'sampling_type'):
+                self.sample()
```
psavery, line 144: One alternative you could have for all of these `if hasattr()` checks is to make a separate function like this:

```python
    def sample_if_possible(self):
        required_attributes = ('pgnum', 'avg_ang_spacing', 'sampling_type')
        if not all(hasattr(self, x) for x in required_attributes):
            return

        self.sample()
```

And then in all of the setters after setting, just call this function. For instance:

```python
    @avg_ang_spacing.setter
    def avg_ang_spacing(self, ang):
        self._avg_ang_spacing = ang
        self.sample_if_possible()
```

psavery (approved):
I left a few suggestions, but it is still good to merge as-is.

Looks good to me!

hexrd/sampleOrientations/sampleRFZ.py
```diff
@@ -0,0 +1,162 @@
...
+            for kk in prange(-N, N+1):
+                zz = (kk + shift) * delta
+                cu = np.array([xx, yy, zz])
+                ma = np.max(np.abs(cu))
+
+                if ma <= ap_2:
+                    ro = cu2ro(cu)
+                    if insideFZ(ro, pgnum):
+                        res[ctr,:] = ro2qu(ro)
+                        ctr += 1
```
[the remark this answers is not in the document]
saransh13: Is there any way around it? I don't know before the call what the final size of the array will be, so I allocate the maximum possible size and then return only the number that gets generated.

hexrd/sampleOrientations/sampleRFZ.py
saransh13: Yes, the code looks much cleaner with this. I'll apply these changes.

hexrd/sampleOrientations/sampleRFZ.py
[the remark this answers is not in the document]
psavery: Maybe one way would be something to initialize `res` as `np.full((N3, 4), np.nan, dtype=np.float64)` (initialize it to nan), and then get rid of `ctr` entirely and set `res[idx] = ro2qu(ro)` (where `idx` would be the flattened index of `ii, jj, kk`), and then get rid of the nan entries at the end like `res = res[~np.isnan(res)]`.

PR #491, opened 2023-01-09
Add Rygg pinhole to phutil

psavery:
This follows [`calc_twotheta_pinhole()`](https://github.com/HEXRD/hexrd/blob/cb7bbaa341d07596754f3b55435f17654f89e288/scripts/pinhole_2theta.py#L176), but assembles the inputs from the hexrd classes.

~~Once we have two versions of this function: one that takes arbitrary xys and calculates the tth distortion for them, and one that outputs a tth distortion correlation map for all detector pixels, it will be easy to integrate into hexrdgui.~~

~~The function is not complete - we do not yet have the pixel positions in spherical coordinates.~~

~~This function is now working, but the results are not correct. We need to review the inputs to ensure they are being converted properly.~~

This now appears to be working exactly as it should!

The only thing left is to look into improving performance, which may be done on a separate PR.

To fix:
- [x] Shift for both overlays and polar field appear to be going in the wrong directions

### DIFF
--- a/hexrd/utils/decorators.py
+++ b/hexrd/utils/decorators.py
@@ -132,3 +132,10 @@ def numba_njit_if_available(func=None, *args, **kwargs):
         return decorator
     else:
         return decorator(func)
+
+
+# Also expose prange depending on whether we have numba or not
+if USE_NUMBA:
+    from numba import prange
+else:
+    prange = range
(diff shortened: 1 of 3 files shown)


psavery:
This PR is ready to go! The answers seem correct. Speeds are reasonable (although probably still some room for improvement somewhere).

Here was a nice confirmation test I was using (although it is using `pkl`): [pinhole_confirmation.zip](https://github.com/HEXRD/hexrd/files/11095016/pinhole_confirmation.zip)

saransh13 (approved):
All features work as expected! LGTM.

PR #493, opened 2023-01-25
Ignore anomalous scattering factors for Z > 92

saransh13:
As the title says. Ignores `f'`, `f"`, `fNT` and `frel` for atoms with Z > 92. Came up during analysis of alpha-Pu.

### DIFF
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -578,7 +578,7 @@ class Material(object):
                 if x:
                     try:
                         U[ii] = float(U[ii])
-                    except:
+                    except ValueError:
                         U[ii] = self.DFLT_U[0]
 
             self._U = numpy.asarray(U)
--- a/hexrd/material.py
+++ b/hexrd/material.py
@@ -554,7 +554,7 @@ class Material(object):
             msg = (f"'Debye-Waller factors not present. "
                    f"setting to same values for all atoms.'")
             warn(msg)
-            U = 1.0/numpy.pi/2./numpy.sqrt(2.) * numpy.ones(atompos[0].shape)
+            U = self.DFLT_U[0] * numpy.ones(atompos[0].shape)
             self._U = U
         else:
             if(occ_U[1] in cifdata.keys()):
@@ -576,7 +576,10 @@ class Material(object):
 
             for ii, x in enumerate(chkstr):
                 if x:
-                    U[ii] = 1.0/numpy.pi/2./numpy.sqrt(2.)
+                    try:
+                        U[ii] = float(U[ii])
+                    except:
+                        U[ii] = self.DFLT_U[0]
 
             self._U = numpy.asarray(U)
         '''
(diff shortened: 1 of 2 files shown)


hexrd/material.py
```diff
@@ -576,7 +576,10 @@ def _readCif(self, fcif=DFLT_NAME+'.cif'):
 
             for ii, x in enumerate(chkstr):
                 if x:
-                    U[ii] = 1.0/numpy.pi/2./numpy.sqrt(2.)
+                    try:
+                        U[ii] = float(U[ii])
+                    except:
+                        U[ii] = self.DFLT_U[0]
```
psavery, line 582: It's usually best to avoid bare `except` since it will literally catch anything, including errors that we would want to know about (which can lead to bugs that are hard to track down).

What are the kinds of errors that can occur here? `IndexError` from the `ii` index being out of range? We could potentially catch that instead. Or there could also be a `ValueError` converting to float?

hexrd/unitcell.py
```diff
@@ -747,11 +747,20 @@ def InitializeInterpTable(self):
...
-                gid = fid.get('/'+elem)
-                data = np.array(gid.get('data'))
-                self.pe_cs[elem] = interp1d(data[:, 7], data[:, 3])
-                data = data[:, [7, 1, 2]]
-                f_anomalous_data.append(data)
+
+                wav = np.linspace(1.16E2, 2.86399992e-03, 189)
+                zs = np.ones_like(wav)*Z
+                zrs = np.zeros_like(wav)
+                data_zs = np.vstack((wav, zs, zrs)).T
```
psavery, line 754: If this section of code is only used by the `else` block, maybe we should move it into the `else` block?

psavery (approved):
I left a couple of comments, but overall LGTM

psavery:
I just pushed a commit that should fix the failing Mac test.

hexrd/material.py
saransh13: `U[ii]`, when read from the file returns a string. Now, this string could be a number like `0.0` or a `?` to denote that the number is unavailable. The try was to make sure `float(U[ii])` was successful, and if it was not then we can just assign the default value. Can you think of a better way to do this?

hexrd/unitcell.py
saransh13: done

hexrd/material.py
psavery: We could do `if U[ii] == '?'` to check if it is a question mark, if the only allowable non-float value is a question mark (that way, if some typo is entered instead, it would produce an error).

Or, the exception could catch `ValueError` to catch any errors when converting to float.

hexrd/unitcell.py
psavery: Looks like it isn't pushed yet.

PR #500, opened 2023-02-21
IPF coloring added to NF plotting

rachelelim:
Plotting post-NF reconstruction now uses IPF colors instead of some arbitrary color scheme.


scripts/process_nf.py
```diff
@@ -354,25 +354,11 @@
...
-
-        plt.figure(1)
-        plt.imshow(confidence_map[0])
-        plt.title('Bottom Layer Confidence')
 
-        plt.figure(2)
-        plt.imshow(confidence_map[-1])
-        plt.title('Bottom Layer Confidence')
+mat = Material(name=mat_name, material_file=mat_file, dmin=valWUnit(
+    'lp', 'length',  0.05, 'nm'), kev=valWUnit('kev', 'energy', beam_energy, 'keV'))
```
psavery, line 359: Maybe this should be moved into the `if output_plot_check:` block since it is only used there?

psavery (approved):
This looks good to me.

@darrencpagan, did you want to look over this and/or test it as well?

rachelelim:
I'm not sure how to revert those changes that accidentally got added on top of the IPF color change

psavery:
The last 6 commits on this PR were merged accidentally. However, considering that they only added new files, and did not edit any existing files, we are going to just leave them.

PR #501, opened 2023-02-24
Revert "IPF coloring added to NF plotting"

joelvbernier:
Reverts HEXRD/hexrd#500

Need to remove commits after 136bc829e311d915deb9a71524a1aaf1803007b7 and package those in a new NF config-related PR.


psavery:
I took a look over the commits accidentally added in #500, and it looks like the extra commits only added new files, mostly under a new `nf_config` directory. The extra commits did not modify any existing files. Therefore, I think the best course of action here is to just leave them. We can edit/update the new files in future pull requests.

Issue #502, opened 2023-03-08
Absorption length


saransh13:
The `unitcell.absorption_length` needs to be modified to include the coherent and incoherent scattering factors. Currently, it only accounts for the photo-effect cross-section. This becomes important at higher energies.

PR #503, opened 2023-03-08
add incoherent and coherent scattering to the absorption cross-section.

saransh13:
This fixes issue #502.

### DIFF
--- a/hexrd/unitcell.py
+++ b/hexrd/unitcell.py
@@ -15,6 +15,13 @@ import time
 from hexrd.utils.decorators import numba_njit_if_available
 
 eps = constants.sqrt_epsf
+ENERGY_ID = 0
+REAL_F1_ID = 1
+IMAG_F2_ID = 2
+MU_ID = 3
+COH_INCOH_ID = 4
+MU_K_ID = 6
+WAV_ID = 7
 
 ''' calculate dot product of two vectors in any space 'd' 'r' or 'c' '''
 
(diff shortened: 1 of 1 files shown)


hexrd/unitcell.py
```diff
@@ -751,7 +751,7 @@ def InitializeInterpTable(self):
                 if Z <= 92:
                     gid = fid.get('/'+elem)
                     data = np.array(gid.get('data'))
-                    self.pe_cs[elem] = interp1d(data[:, 7], data[:, 3])
+                    self.pe_cs[elem] = interp1d(data[:, 7], data[:, 3]+data[:,4])
```
psavery, line 754: It would be nice to have these columns stored as variables to make this easier to read (such as `SCATTERING_FACTORS_IDX` or something like that).

psavery (approved):
LGTM

saransh13:
Good idea. I can add that to the PR.

hexrd/unitcell.py
```diff
@@ -751,8 +758,9 @@ def InitializeInterpTable(self):
                 if Z <= 92:
                     gid = fid.get('/'+elem)
                     data = np.array(gid.get('data'))
-                    self.pe_cs[elem] = interp1d(data[:, 7], data[:, 3])
-                    data = data[:, [7, 1, 2]]
+                    self.pe_cs[elem] = interp1d(data[:, WAV_ID], \
+                    data[:, MU_ID]+data[:,COH_INCOH_ID])
```
psavery, line 762:
```suggestion
                    self.pe_cs[elem] = interp1d(data[:, WAV_ID],
                        data[:, MU_ID]+data[:,COH_INCOH_ID])
```
FYI: you don't need the `\` since there are parentheses, and it's better to indent the next line to show it is continuing from the first line.

PR #504, opened 2023-03-13
Cylindrical detector

saransh13:
Add cylindrical detector or "rind" for TARDIS. Task list below:

- [x] fix bugs for non-zero tilt
- [x] add support for pixel solid angles for intensity correction
- [ ] test overlays (laue and powder)
- [x] test stereographic projection
- [ ] test pinhole corrections
- [x] outlines for IMAGE PLATE 3
- [x] loading of IMAGE PLATE 3 in LLNL import

### DIFF
--- a/hexrd/instrument/planar_detector.py
+++ b/hexrd/instrument/planar_detector.py
@@ -251,8 +251,5 @@ def _pixel_solid_angles(rows, cols, pixel_size_row, pixel_size_col,
     # Concatenate all the results together
     solid_angs[:] = np.concatenate(list(results))
     solid_angs = solid_angs.reshape(rows, cols)
-    mi = solid_angs.min()
-    if mi > 0.:
-        solid_angs = solid_angs/mi
 
     return solid_angs


hexrd/instrument.py
```diff
@@ -681,6 +681,10 @@ def __init__(self, instrument_config=None,
             for det_id, det_info in detectors_config.items():
                 pixel_info = det_info['pixels']
                 affine_info = det_info['transform']
+                if 'detector_type' in det_info:
+                    detector_type = det_info['detector_type']
+                else:
+                    detector_type = 'planar'
```
psavery, line 687: Here, this is simpler:
```suggestion
                detector_type = det_info.get('detector_type', 'planar')
```

hexrd/instrument.py
```diff
@@ -725,8 +729,11 @@ def __init__(self, instrument_config=None,
                             raise RuntimeError(
                                 "problem with distortion specification"
                             )
+                DetectorClass = PlanarDetector
+                if detector_type.lower() == 'cylindrical':
+                     DetectorClass = CylindricalDetector
```
psavery, line 734: We might want to put a dict of detector types at the top of the file like:
```
DETECTOR_TYPES = {
    'planar': PlanarDetector,
    'cylindrical': CylindricalDetector,
}
```

And then this code here would have some error checking as well:

```python
    if detector_type.lower() not in DETECTOR_TYPES:
        raise NotImplementedError(f'Unknown detector type: {detector_type}')

    DetectorClass = DETECTOR_TYPES[detector_type.lower()]
```

That error checking can be helpful because if a user mistypes the word `cylindrical`, they would probably prefer to see an error rather than it continuing as normal with a `PlanarDetector` (when they might think it is using a `CylindricalDetector`).

hexrd/instrument.py
```diff
@@ -3573,7 +3580,245 @@ def simulate_laue_pattern(self, crystal_data,
...
+                 name='rind',
+                 bvec=ct.beam_vec,
+                 xrs_dist=None,
+                 evec=ct.eta_vec,
+                 saturation_level=None,
+                 panel_buffer=None,
+                 tth_distortion=None,
+                 roi=None,
+                 distortion=None,
+                 max_workers=max_workers_DFLT):
```
psavery: I think this is a good first pass for the initial implementation to get it working, but we will probably want to make a base `Detector` class that both the `CylindricalDetector` and `PlanarDetector` inherit from.

Inheritance carries an `is-a` relationship. As in, if `CylindricalDetector` inherits from `PlanarDetector`, that means `CylindricalDetector` "is a" `PlanarDetector`, which isn't the case. However, `CylindricalDetector` "is a" `Detector`.

Then if we have a base `Detector` class, we can have the `__init__` args for `CylindricalDetector` be like:

```python
def __init__(self, cylindrical_arg1, cylindrical_arg2, **detector_kwargs):
    super().__init__(**detector_kwargs)
    ...

```
And the arguments and logic for initializing the detector would be the same for all detectors and be done in the `Detector.__init__()` method.

hexrd/resources/tardis_reference_config.yml
```diff
@@ -8,6 +8,7 @@ beam:
     polar_angle: 90.0
 detectors:
   IMAGE-PLATE-2:
+    detector_type: 'planar'
```
psavery, line 11: Since most detectors are `planar`, I think it's fine if we just leave out the `detector_type` if it is planar (as it will default to being `planar`). But we can still put it here if it helps clarify that it is planar.

hexrd/instrument.py
```diff
@@ -3573,7 +3580,257 @@ def simulate_laue_pattern(self, crystal_data,
...
+
+    @property
+    def beam_position(self):
+        """
+        returns the coordinates of the beam in the cartesian detector
+        frame {Xd, Yd, Zd}.  NaNs if no intersection.
+        """
+        output = np.nan * np.ones(2)
+        pt_on_cylinder = self._unitvec_to_cylinder(
+                         np.atleast_2d(bvec))
```
psavery, line 3823: Add indentation here to make it more readable.
```suggestion
                             np.atleast_2d(bvec))
```

hexrd/instrument.py
```diff
@@ -3573,7 +3580,257 @@ def simulate_laue_pattern(self, crystal_data,
...
+        return np.array([self.rows*self.pixel_size_row,
+                         self.cols*self.pixel_size_col])
+
+    @property
+    def beam_position(self):
+        """
+        returns the coordinates of the beam in the cartesian detector
+        frame {Xd, Yd, Zd}.  NaNs if no intersection.
+        """
+        output = np.nan * np.ones(2)
```
psavery, line 3821: If you reassign `output` below, you don't need to initialize it beforehand (it will be overwritten).

hexrd/instrument.py
saransh13: Implemented these changes.

hexrd/instrument.py
saransh13: Implemented.

hexrd/instrument.py
```diff
@@ -3573,7 +3578,257 @@ def simulate_laue_pattern(self, crystal_data,
...
 
+    @property
+    def beam_position(self):
+        """
+        returns the coordinates of the beam in the cartesian detector
+        frame {Xd, Yd, Zd}.  NaNs if no intersection.
+        """
+        output = np.nan * np.ones(2)
+        pt_on_cylinder = self._unitvec_to_cylinder(
+                             np.atleast_2d(bvec))
```
psavery, line 3821: @saransh13 This `bvec` is undefined here.

saransh13: Good catch. WIll fix this.

joelvbernier:
What needs to be done is to at a `type` label to each detector in the config that gets parsed in the init.  The default value will be `planar`, which will maintain backward compatibility with earlier config files (v0.8).

joelvbernier:
I also *strongly* suggest package-ifying the instrument module as a separate PR.

psavery:
> I also _strongly_ suggest package-ifying the instrument module as a separate PR.

Adding the new detector class was making the `instrument.py` file even larger, and putting it off just makes it harder in the future. I think it was a good time to split it up. It's not going to break anything - no import changes are needed, and you can still do things like `from hexrd.instrument import HEDMInstrument` as before (since we are exposing those in the `__init__.py` file).

psavery:
A warning is now printed the first time a cylindrical detector is initialized in either hexrd/hexrdgui, indicating that it is still a work in progress, and that some features may not work.

We are planning to proceed to merge this PR very soon to prevent merge conflicts, considering that the PR has gotten large.

PR #505, opened 2023-03-13
added npz file and namedtuple for grains output

donald-e-boyce:
This is just for convenience of the form of output of grains data. In addition to the `grains.out` file, this writes a `grains.npz` containing arrays for each field in the grains.out file. There is a namedtuple that carries the data that can be easily written to an npz file or read from an npz file. You can load a file like this:

```
from hexrd.fitgrains import GrainData
gd = GrainData.load("grains.npz")
```
The convenience is that all the data can be accessed by name, e.g. `gd.completeness` or `gd.centroids`.


psavery (approved):
Looks good to me, Don!

PR #511, opened 2023-04-13
Make reflections

donald-e-boyce:
This converts the function make_reflection_patches into an iterator.  The changes were in the module `xrdutils.utils`. The main file using was `instrument.py,` but in both places, it uses it as an iterator anyway, so no changes were needed. Other files affected were:

- `hexrd/calibration/laue_calibrator.py`: Here patches are used as a list because the number of patches is set. So I used `list(patches)` where the list was needed.  In fact, this module is not functional. It doesn't even import. It doesn't look close to being usable.
- `scripts/calibration_mockup.py`:  this imported `make_reflection_patches` but did not use it, so I removed the import
- `scripts/calibrate_from_powder.py`:  this uses `make_reflection_patches` once as an iterator, so no changes were made


psavery (approved):
LGTM. Great changes, Don!

psavery:
> * `hexrd/calibration/laue_calibrator.py`: Here patches are used as a list because the number of patches is set. So I used `list(patches)` where the list was needed.  In fact, this module is not functional. It doesn't even import. It doesn't look close to being usable.

This also brings up the fact that we may want to clean up some of these unused modules at some point. Some of them have duplicated code from other modules.

psavery:
> We use `make_reflection_patches` in hexrdgui in one place. I may need to make a small modification to get it to work there, but it shouldn't be difficult.

I put up hexrd/hexrdgui#1396 to fix the one place we were using this in hexrdgui.

psavery:
I gave this a try with the fit-grains monolithic ruby example. It went from taking around 10 GB per process to around 100 MB per process. A massive improvement! It allows us to utilize significantly more parallelism when there are many grains.

I think we can hopefully conclude this fixes https://github.com/HEXRD/hexrdgui/issues/1329

PR #513, opened 2023-04-24
fixing HKL reset problem in fitgrains

### LINKED ISSUE #512: tth_max setting cannot be used with min_sfac in fit-grains
In running fit-grains, the `tth_max` setting can override the `min_sfac` setting.  If `tth_max` is either True or a number > 0, then it resets all exclusions to False when executing fit_grains. The effect is that you cannot specify both a maxium 2-theta and a minimum structure factor. If you want a minimum structure factor, you need to turn off `tth_max`.

I think it makes more sense to be able to use both together. Then you could restrict the range of the rings while eliminating low-intensity HKLs.

donald-e-boyce:
This fixes #512 . 

Changes include:

- A new `exclude()` method in PlaneData for selection of HKLs using min/max cutoff values for d-spacing, two theta, structure factor and powder intensity. See  [test_exclude.py](https://github.com/HEXRD/hexrd/files/11314214/test_exclude.py.txt) for a script for testing the exclude() method. I used the ruby material in the single_GE example.

- A simple fix for a problem in handling multiple configuration files (each subsequent config was merged with the first instead of being merged with the previous, as described; and the `save() `function was out of date.)
- Setting plane data exclusions in `cli.fit_grains` before the main fitting method is called instead of in `hexrd.fitgrains` inside the main fitting method. This was the main problem because it reset the HKLs inside the main call; this applies to the GUI as well as the CLI.
- Had the material config file process the plane data only once, instead of on each call.

To verify this, I ran a completeness study on the single_GE example, which I will post shortly.

### DIFF
--- a/hexrd/crystallography.py
+++ b/hexrd/crystallography.py
@@ -816,6 +816,73 @@ class PlaneData(object):
         return
     exclusions = property(get_exclusions, set_exclusions, None)
 
+    def exclude(
+            self, dmin=None, dmax=None, tthmin=None, tthmax=None,
+            sfacmin=None, sfacmax=None, pintmin=None, pintmax=None
+    ):
+        """Set exclusions according to various parameters
+
+        Any hkl with a value below any min or above any max will be excluded. So
+        to be included, an hkl needs to have values between the min and max
+        for all of the conditions given.
+
+        Note that method resets the tThMax attribute to None.
+
+        PARAMETERS
+        ----------
+        dmin: float > 0
+            minimum lattice spacing (angstroms)
+        dmax: float > 0
+            maximum lattice spacing (angstroms)
+        tthmin: float > 0
+            minimum two theta (radians)
+        tthmax: float > 0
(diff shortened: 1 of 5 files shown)


hexrd/crystallography.py
```diff
@@ -816,6 +816,73 @@ def set_exclusions(self, exclusions):
         return
     exclusions = property(get_exclusions, set_exclusions, None)
 
+    def exclude(
+            self, dmin=None, dmax=None, tthmin=None, tthmax=None,
+            sfacmin=None, sfacmax=None, pintmin=None, pintmax=None
+    ):
```
psavery, line 822: I think it would be valuable if we modified your test script for this function and put it in the [`tests`](https://github.com/HEXRD/hexrd/tree/master/tests) so that it runs with the CI.

In general, I think having more tests running in the HEXRD CI would be very helpful for future development.

psavery (approved):

PR #514, opened 2023-04-27
Add specific exception for buffer mismatch

psavery:
We can catch this exception in HEXRDGUI, and this will prevent us having to check the buffer shape every time, as is done in hexrd/hexrdgui#1419.


bnmajor (approved):
LGTM

PR #515, opened 2023-04-27
Fix material circular import

psavery:
If you start a program by running `from hexrd import material`, this triggers
a circular import, because the material ends up triggering an import of the
instrument, which triggers an import of `hexrd.material` again.

Do not import the instrument in `xrdutils`. Let's keep it simple. This fixes
the circular import.


donald-e-boyce (approved):
Looks good. The Material import works, and I double checked the detector_type spelling.

psavery:
> Looks good. The Material import works, and I double checked the detector_type spelling.

Thank you Don!

PR #516, opened 2023-04-27
Remove duplicate Laue spots from simulator

psavery:
This issue arose from #509, due to somewhat complex details.

In short, the old `gvecToDetectorXY()` would filter out g-vectors going the wrong way (they would result in `nan`), but `gvecToDetectorXYArray()` would not.

The new `gvec_to_xy()` has behavior matching `gvecToDetectorXYArray()`. But that means that places where `gvecToDetectorXY()` used to be used, we need to verify that the function does not receive g-vectors going the wrong way.

Looking through the code, however, I only found one place where `gvecToDetectorXY()` used to receive g-vectors going the wrong way. It was the Laue overlay simulator, which would compute g-vectors from symmetrical HKLs (which is why it would have duplicate g-vectors going the wrong way).

All other cases where `gvecToDetectorXY()` was called would be after the g-vectors were computed from `anglesToGvec()`, which I believe should always produce g-vectors going in the correct direction.

As such, we only need to filter out wrong-direction g-vectors in the Laue simulator.

Fixes: hexrd/hexrdgui#1412

### DIFF
--- a/hexrd/instrument/detector.py
+++ b/hexrd/instrument/detector.py
@@ -1367,7 +1367,7 @@ class Detector:
 
             # Filter out g-vectors going in the wrong direction. `gvec_to_xy()` used
             # to do this, but not anymore.
-            to_keep = gvec_c[2] * self.bvec[2] <= 0
+            to_keep = np.dot(gvec_c.T, self.bvec) <= 0
 
             hkls = hkls[:, to_keep]
             gvec_c = gvec_c[:, to_keep]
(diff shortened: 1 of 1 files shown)


saransh13 (commented):
I think for the most general case Line 1370 is probably not always accurate. 
I think the right logic should be `np.dot(gvec_c, self.bvec) <= 0`.

psavery:
For my example, this produces nearly the same results with a few differences. One of which is that `(2, -4, 0)` near the bottom becomes `(-2, 4, 0)`. It looks like that is what it actually was previously (before the transforms were merged), so that also matches better.

Previous:
current

Using `np.dot()`:
new

psavery:
And actually, this now matches exactly what we had before migrating to the new transforms. Here is what we used to have in the old transforms:

oscar

It appears the dot product version is an exact match for this example.

saransh13 (approved):
LGTM

Issue #519, opened 2023-05-03
material config instantation requires hedm instrument


donald-e-boyce:
When the root config is creating the material config, it uses the beam energy from the instrument to set the energy for the material.  The problem arose trying to update the existing unit tests for the material config, which mainly were verifying the yaml and defaults for various keys in the material config file. It seems like there should be a better way to do this. Maybe set the material beam energy after you instantiate the instrument instead of creating an instrument before you can instantiate the material.

https://github.com/HEXRD/hexrd/blob/212028f67a6c423963f7a68518c33868fce54bb8/hexrd/config/root.py#L51-L58

PR #520, opened 2023-05-09
Config update for using existing exclusions

### LINKED ISSUE #518: Add option to not reset hkls in fit-grains
Before #513, if the minimum structure factor was `None` and the max tth was `False`, the HKLs from the materials file itself would be used in fit-grains. This was not used in the HEXRD examples repo, but it was used in HEXRDGUI when exporting the HEDM workflow, because the users are able to manually modify the selected HKLs in HEXRDGUI, and setting those in the materials file was the best way to make sure they matched the GUI.

But with #513, the HKLs now get reset at the beginning of the fit-grains CLI if minimum structure factor is `None` and the max tth is `False`.

We should add an option to just use the HKLs from the file, so that we can set this option when exporting the workflow from the GUI.

donald-e-boyce:
This allows saved exclusions to be used in command line processing.
   Fixes #518 and #519.

   Exclusion options are set in both the material config and in the fit-grains
   config. In each case, there is a boolean `reset_exclusions` parameter.
   If that parameter is `False`, then current exclusions are not changed.
   In the material config, that means the exclusions read from the file are
   unchanged. In the fit-grains config, it means that the current material
   exclusions are not altered.  If the `reset_exclusions` parameter is `True`,
   then eight possible parameters are available in the config file: `dmin`, `dmax`,
   `tthmin`, `tthmax`, `sfacmin`, `sfacmax`, `pintmin` and `pintmax`; these parameters
   set the exclusions using the `PlaneData.exclude() `method.

   Legacy specifications are checked for. In the material config, the `dmin`
   and `min_sfac_ratio` parameters. The `dmin` still has the same effect; it is
   used to initialize the material and determines the set of HKLs to use.
   The `min_sfac_ratio` is equivalent to the new `sfacmin` parameter. If
   `min_sfac_ratio` is found, a `DeprecationWarning` is issued, and the value
   is used as the default for the `sfacmin` exclusion parameter, but if both
   are specified, the `sfacmin` value is used. In the fit_grains config, the
   old `tth_max` parameter is checked for. If `True`, it gets a computed value of
   the instrument max two-theta. If both `tth_max` and the new `tthmax` are given,
   the newer `tthmax` value is used.

   The config package unit tests were moved into the main tests directory.
   All tests except for the instrument config were updated; the instrument
   tests were diabled for now. New tests were created for the latest changes.

   Regarding Issue #519, the root config instrument property was changed so that
   the instrument file can be `None`. and a check was added in the
   config.instrument module raising a `RuntimeError` if the instrument file was
   `None`. This allows the material config to be checked without instantiating
   and instrument.


hexrd/config/utils.py
```diff
@@ -23,8 +38,36 @@ def _merge_dicts(a, b):
...
+            dmin = cfg.get(yaml_key("dmin"), None),
+            dmax = cfg.get(yaml_key("dmax"), None),
+            tthmin = cfg.get(yaml_key("tthmin"), None),
+            tthmax = cfg.get(yaml_key("tthmax"), None),
+            sfacmin = cfg.get(yaml_key("sfacmin"), sfmin_dflt),
+            sfacmax = cfg.get(yaml_key("sfacmax"), None),
+            pintmin = cfg.get(yaml_key("pintmin"), None),
+            pintmax = cfg.get(yaml_key("pintmax"), None),
+        )
+    )
```
psavery, line 73: An alternative way (although you don't need to do it):

```python
default_params = {
    "dmin": None,
    "dmax": None,
    "tthmin": None,
    "tthmax": None,
    "sfacmin": sfmin_dflt,
    "sfacmax": None,
    "pintmin": None,
    "pintmax": None,
}

params = {k: cfg.get(yaml_key(k), v) for k, v in default_params.items()}

return reset_exclusions, ExclusionParameters(**params)
```

hexrd/config/material.py
```diff
@@ -35,11 +42,16 @@ def active(self):
...
     def materials(self):
         """Return a dict of materials."""
+        #
+        # If reset_exclusions is False, we use the material as read from
+        # the file. This includes not using the dmin value, which may alter
+        # the HKLs available.
+        #
+        kwa = {"f": self.definitions}
+        if self.reset_exclusions:
+            kwa["dmin"] = valWUnit("dmin", "length", self.dmin, "angstrom")
```
psavery, line 52: These three lines could be inside the `if not hasattr(self, '_materials')`, right?

hexrd/config/material.py
```diff
@@ -76,14 +97,8 @@ def _make_plane_data(self):
...
-        if self.fminr is not None:
-            # !!! exclusions are all False upon generation; setting
-            #     the tThMax attribute later won't affect these.
-            excl_full = np.array(pd.exclusions, dtype=bool)
-            pd.exclusions = None
-            mod_f_sq = pd.structFact
-            excl_these = mod_f_sq/np.max(mod_f_sq) < self.fminr
-            pd.exclusions = np.logical_or(excl_full, excl_these)
+        if self.reset_exclusions:
+            pd.exclude(**self.exclusion_parameters._asdict())
```
psavery, line 101: This is much nicer!

tests/config/test_material.py
```diff
@@ -0,0 +1,145 @@
...
+class TestMaterialConfig(TestConfig):
+
+
+    @classmethod
+    def get_reference_data(cls):
+        return reference_data % test_data
+
+
+    def test_definitions(self):
+        self.assertEqual(
```
psavery, line 58: I like the regular `assert ...` statements we use in pytest. But I guess we need to use these in the unittests? It's okay, we can leave them.

psavery (approved):
Just a few minor comments, but overall looks good to me! It's great that we will have more tests!

psavery:
The Mac testing has an assertion error. It looks like the multiprocessing somehow ended up being `1.5` instead of `1`?

hexrd/config/root.py
```diff
@@ -52,9 +53,12 @@ def instrument(self, instr_config):
...
         if not hasattr(self, '_material_config'):
             self._material_config = MaterialConfig(self)
-        # !!! must make matl beam energy consistent with the instrument
-        beam_energy = self.instrument.hedm.beam_energy
-        self._material_config.beam_energy = beam_energy
+
+        if self.instrument.configuration is not None:
+            # !!! must make matl beam energy consistent with the instrument
+            beam_energy = self.instrument.hedm.beam_energy
+            self._material_config.beam_energy = beam_energy
```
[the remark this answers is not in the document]
donald-e-boyce: I agree with you here, getters shouldn't have unexpected side effects involving a different object. Here, I just kicked the ball down the road. It makes sense for the instrument to carry the beam energy, not so much for the material. The material has properties for two-theta and d-spacing. I would rather these were functions of beam energy than properties. However, there are other more important things to spend time on.

tests/config/test_material.py
donald-e-boyce: I like the simpler assert better also, and for new tests I plan to use the pytest interface.  I'm not sure if you can mix the unittest asserts and regular asserts, so I just kept it consistent in the file.

hexrd/config/material.py
donald-e-boyce: Did this.

donald-e-boyce:
I resolved the multiprocessing issue. It was an issue with the test, not with the actual config file. The test used `ncpus/2` instead of `ncpus // 2`

psavery (approved):
Looks good to me!

donald-e-boyce:
I just pushed a few more minor changes. The config.instrument module was modified to work with the GUI; the check for instrument config file was moved inside the check for the _hedm attribute, which is set directly by the GUI.  

In addition, I added a log message giving the number of HKLs used in fit-grains, and I added a warning when the number of valid reflections was too few for a fit. 

Overall, I feel good about this PR.  All pytest tests succeed, and the completeness study in the examples gets identical results.  We still have to get the saved workflow from the GUI to work though.  

For the example I ran in the GUI, the problem came in the find-orientations step. From the GUI
```
building eta_ome maps using hkls: ['[0 1 2]', '[1 0 4]', '[1 1 0]', '[1 1 3]', '[0 2 4]']
working on detector 'GE'...
		...took 4.442435 seconds
material:min_sfac_ratio not specified, defaulting to None
```

But in the exported `workflow.yml` file, I got:
```
  orientation_maps:
    active_hkls:
    - - 0
      - 1
      - 2
    bin_frames: 1
    file: null
    filter_maps: false
    threshold: 25.0
  seed_search:
    fiber_step: 0.5
    hkl_seeds:
    - 0
```

Maybe we should continue by opening an Issue on hexrdgui.

PR #523, opened 2023-05-10
Add support for .gel files

bnmajor:
Apply a bitwise not operation before scaling the data. The scale factor was computed to be 2.945215294651886e-07 by @psavery and @saransh13. See [this comment](https://github.com/HEXRD/hexrdgui/issues/1387#issuecomment-1498177302) for more information.

Fixes HEXRD/hexrdgui#1387

### DIFF
--- a/hexrd/imageseries/load/imagefiles.py
+++ b/hexrd/imageseries/load/imagefiles.py
@@ -19,6 +19,8 @@ from . import ImageSeriesAdapter
 from .metadata import yamlmeta
 from ..imageseriesiter import ImageSeriesIterator
 
+SCALE_FACTOR = 2.9452155399372724e-07
+
 
 class ImageFilesImageSeriesAdapter(ImageSeriesAdapter):
     """collection of image files"""
@@ -45,9 +47,10 @@ class ImageFilesImageSeriesAdapter(ImageSeriesAdapter):
             return self._nframes
 
     def __getitem__(self, key):
+        filename = self._files[key]
+        file_extension = os.path.splitext(filename)[-1]
         if self.singleframes:
-            imgf = self._files[key]
-            with fabio.open(imgf) as img:
+            with fabio.open(filename) as img:
                 data = img.data
         else:
             (fnum, frame) = self._file_and_frame(key)
@@ -64,6 +67,8 @@ class ImageFilesImageSeriesAdapter(ImageSeriesAdapter):
             if np.max(data) > dinfo.max:
                 raise RuntimeError("specified dtype will truncate image")
             return np.array(data, dtype=self._dtype)
+        elif file_extension == '.gel':
+            return data.astype(np.float64)**2 * SCALE_FACTOR
         else:
             return data
 
@@ -219,6 +224,9 @@ class FileInfo(object):
             self._fabioclass = img.classname
             self._imgframes = img.nframes
             self.dat = img.data
+            if os.path.splitext(filename)[-1] == '.gel':
+                raw_img = np.invert(img.data)
+                self.dat = raw_img.astype(np.float64)**2 * SCALE_FACTOR
 
         d = kwargs.copy()
         self._empty = d.pop('empty', 0)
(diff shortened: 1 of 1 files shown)


hexrd/imageseries/load/imagefiles.py
```diff
@@ -19,6 +19,8 @@
 from .metadata import yamlmeta
 from ..imageseriesiter import ImageSeriesIterator
 
+SCALE_FACTOR = 2.9452155399372724e-07
```
psavery, line 22: Can we rename this to `GEL_SCALE_FACTOR` for clarity?

bnmajor: Sure thing! Done

hexrd/imageseries/load/imagefiles.py
```diff
@@ -64,6 +67,8 @@ def __getitem__(self, key):
             if np.max(data) > dinfo.max:
                 raise RuntimeError("specified dtype will truncate image")
             return np.array(data, dtype=self._dtype)
+        elif file_extension == '.gel':
+            return data.astype(np.float64)**2 * GEL_SCALE_FACTOR
```
donald-e-boyce, line 71: It looks like the gel processing should go before the dtype check, not as an elif.  If the dtype is not None, then it returns the image, so it will usually never get to the check for gel data.  In fact, we can probably eliminate the whole dtype check.  That is checking whether the data max is larger than the max for that dtype, and I don't see how that can happen if the image is that dtype.  Also, it wastes time since it has to process the entire image to find the max. 
 Hard to understand my own code! In any case, the gel data processing  should come before that check if we keep it.

hexrd/imageseries/load/imagefiles.py
```diff
@@ -219,6 +224,9 @@ def __init__(self, filename, **kwargs):
             self._fabioclass = img.classname
             self._imgframes = img.nframes
             self.dat = img.data
+            if os.path.splitext(filename)[-1] == '.gel':
+                raw_img = np.invert(img.data)
```
donald-e-boyce, line 228: Why is the data inverted here and not above in the __get_item__() method?

donald-e-boyce (changes requested):
Either put the gel data processing before the dtype check or eliminate the dtype check completely, if you agree that it doesn't actually do anything.

donald-e-boyce:
Actually, I take back that we can get rid of the test for dtype and check for  max value in the __get_item__(). You can set the dtype through the YAML, then it will convet the image to that type. So you can read 32 bit data as 16 bit if you want.  I don't recall why I put that in, but maybe Joel had asked for that.

hexrd/imageseries/load/imagefiles.py
psavery: I checked with @bnmajor, and it should be inverted in both. I'm modifying this logic to be in a function instead.

psavery:
> Actually, I take back that we can get rid of the test for dtype and check for max value in the **get_item**(). You can set the dtype through the YAML, then it will convet the image to that type. So you can read 32 bit data as 16 bit if you want. I don't recall why I put that in, but maybe Joel had asked for that.

I didn't make these changes yet. Maybe we should refactor the dtype logic in a separate PR if we need to?

By the way, one way we can check if an overflow/underflow occurred is by doing this:

```python
new_data = np.array(data, dtype=self._dtype)
if not np.array_equal(new_data, data):
    raise RuntimeError("Overflow/underflow occurred")
```

Basically, checking to see if the numbers still match (when it does the second comparison, it should promote the smaller dtype to the larger one).

psavery (approved):

psavery:
> Just one more question. In Patrick's slack message to me, he said the gel data was "compressed into a unit16". So that sounds like it was float before. The question is whether you have to convert the image data (uint16) back to float after you load the image. Likewise, if you load an image of ints and multiply by 2.xxxxxxxxe-7, it seems you're expecting a float output.

Hey Don! Yes it is converted to a float after we "decompress" it. Note that the `_process_gel_data()` function does a type conversion: `return array.astype(np.float64)**2 * GEL_SCALE_FACTOR`

PR #527, opened 2023-05-26
Structureless calibration

saransh13:
New class to perform structureless calibration when the line positions are not known. This branch is required for PR https://github.com/HEXRD/hexrdgui/pull/1484

Fixes: hexrd/hexrdgui#1447


psavery (approved):
When we are ready to merge hexrd/hexrdgui#1484, let's merge this one and then that one shortly after. The PRs are getting big, so we can hopefully do this soon and patch more in the future!

PR #529, opened 2023-06-08
Change all cases of np.bool to just bool

psavery:
`np.bool` was a deprecated alias for the built-in `bool`. Support for it was removed recently, and thus we really need to just use the built-in bool.

See: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

Corresponding hexrdgui PR: hexrd/hexrdgui#1494

Fixes the issue mentioned in [this comment](https://github.com/HEXRD/hexrdgui/issues/1493#issuecomment-1583175760)


bnmajor (approved):
LGTM!

PR #531, opened 2023-06-09
Use latest lxml

psavery:
Fabio uses lxml (it is optional, but installing fabio installs it by default).

On Windows, lxml 4.9.1 is causing the application to crash silently. There are no errors - the application just crashes when fabio is imported.

We need to force it to use lxml 4.9.2 or greater to avoid this crash. We could alternatively just stop it from being installed as well if we run into more issues in the future.


bnmajor (approved):
LGTM!

Issue #533, opened 2023-06-11
`np.int` deprecated in NumPy 1.20 and onwards causes AttributeError in `hexrd.imageseries.omega`


minety:
Hello,

I have encountered an issue related to the deprecation of `np.int` in NumPy version 1.20 and onwards. When I use your library, it causes an AttributeError in the `hexrd.imageseries.omega` module of your codebase.

Here is the traceback:

```text
File "/Users/yetian/opt/anaconda3/envs/test_hedmpre_env/lib/python3.10/site-packages/hexrd/imageseries/omega.py", line 142, in omegas
    if self.nframes != self.wframes:
  File "/Users/yetian/opt/anaconda3/envs/test_hedmpre_env/lib/python3.10/site-packages/hexrd/imageseries/omega.py", line 185, in wframes
    return np.int(np.sum(wf))
  File "/Users/yetian/opt/anaconda3/envs/test_hedmpre_env/lib/python3.10/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'int'.
`np.int` was a deprecated alias for the builtin `int`. To avoid this error in existing code, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations. Did you mean: 'inf'?
```

The issue arises when attempting to cast to `numpy.int` for type conversion. `numpy.int` has been deprecated in NumPy since version 1.20, because it was merely an alias for the built-in Python `int` type.

A suggested solution would be to replace `np.int` with `int`, `np.int32`, or `np.int64` depending on the required precision.

For example, here's a possible modification for line 185 in `hexrd.imageseries.omega.py`:

```python
return int(np.sum(wf))  # or np.int32(np.sum(wf)) or np.int64(np.sum(wf))
```

Could you please take a look at this? Thank you.

psavery:
Hi @minety! In the latest version of numpy (1.24), aliases like `np.int()` were removed. We have almost finished fixing these in the repo, and we are planning to have a new release in the next couple of days, where this issue should be resolved.

PR #534, opened 2023-06-12
Fix remaining deprecated numpy aliases

psavery:
See [here](https://numpy.org/doc/stable/release/1.24.0-notes.html).

The list of deprecated aliases is:

```
np.object, np.bool, np.float, np.complex, np.str, and np.int
```

Fixes: #533


bnmajor (approved):
LGTM!

PR #539, opened 2023-06-26
Fix memoization size updates for detectors

psavery:
This logic was not working before, and we needed to add the `_pixel_solid_angles` functions to the memoization size updates.

It appears to be working correctly now.


saransh13 (approved):
LGTM!

PR #543, opened 2023-07-08
Add ability to cancel fit-grains programmatically

psavery:
This adds an optional argument to `fit_grains()` named `check_if_canceled_func`. It should be a nullary function that returns a boolean.

If provided, it will be checked every 0.25 seconds while fit-grains is running to see if the operation should cancel.

If it returns `True`, the multiprocessing pool is terminated and `None` is returned.

Because we rely on the multiprocessing pool to cancel, if the cancel function is provided, we use the multiprocessing pool, even if there is only one process to run.

### DIFF
--- a/hexrd/fitgrains.py
+++ b/hexrd/fitgrains.py
@@ -306,11 +306,23 @@ def fit_grains(cfg,
                grains_table,
                show_progress=False,
                ids_to_refine=None,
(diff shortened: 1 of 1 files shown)


psavery:
This was approved verbally by @donald-e-boyce

PR #545, opened 2023-07-12
Update hedm_instrument.py

darrencpagan:
Adding a catch if the instrument panel does not have a buffer.


donald-e-boyce:
That looks good. Can you delete the string with code inside? It doesn't do anything, we may as well clear it out.

psavery:
@darrencpagan Note that you don't need to make a new PR if there are changes needed. You can either add new commits and push to your branch, or amend previous commits and force push to your branch. The PR will see the updates to your branch automatically.

PR #546, opened 2023-07-12
Build time performance optimization

ZackAttack614:
This change adds gcc compiler flags to optimize the runtime for the C API calls.

**Estimated performance increase: 10%**

Changes include:
- Add `-O3` to compiler flags for `transforms_CAPI.c`, `transforms_CFUNC.c`, `hexrd/transforms/new_capi/module.c`
- Add `-ftree-vectorize` to compiler flags for `transforms_CAPI.c`, `transforms_CFUNC.c`, `hexrd/transforms/new_capi/module.c`

Future proposed changes may include:
- Add the `-march=native` flag
- Add the `-mtune=native` flag

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
+import distutils.ccompiler
 import os
-import sys
-from setuptools import setup, find_packages, Extension
 from pathlib import Path
+from setuptools import setup, find_packages, Extension
+import sys
+
 import numpy
 np_include_dir = numpy.get_include()
-import distutils.ccompiler
 
 install_reqs = [
     'fabio>=0.11',
@@ -28,7 +29,7 @@ compiler_type = distutils.ccompiler.get_default_compiler()
 if compiler_type in ("unix", "mingw32"):
     compiler_optimize_flags = ['-O3', '-ftree-vectorize']
 elif compiler_type == "msvc":
-    compiler_optimize_flags = ['/Ox', '/Ob2', '/Oi', '/Ot', '/Oy', '/GL']
+    compiler_optimize_flags = ['/Ox', '/GL']
 else:
     compiler_optimize_flags = []
 
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ def get_old_xfcapi_extension_modules():
         'hexrd.extensions._transforms_CAPI',
         sources=srclist,
         include_dirs=[np_include_dir],
+        extra_compile_args=['-O3', '-ftree-vectorize'],
     )
 
     return [transforms_mod]
--- a/setup.py
+++ b/setup.py
@@ -70,6 +70,7 @@ def get_new_xfcapi_extension_modules():
         'hexrd.extensions._new_transforms_capi',
         sources=['hexrd/transforms/new_capi/module.c'],
         include_dirs=[np_include_dir],
+        extra_compile_args=['-O3', '-ftree-vectorize'],
     )
 
     return [transforms_mod]
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@ from setuptools import setup, find_packages, Extension
 from pathlib import Path
 import numpy
 np_include_dir = numpy.get_include()
+import distutils.ccompiler
 
 install_reqs = [
     'fabio>=0.11',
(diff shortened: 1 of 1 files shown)


psavery:
This is great! It will be awesome if we can get a decent speedup just from adding optimization flags. I did test this on our [monolithic ruby example](https://github.com/HEXRD/examples/blob/master/NIST_ruby/multiruby_dexelas/include/mruby_config_monolithic.yml) on Linux and I was able to get a ~10% speedup for running it both serially and with multiprocessing.

I believe (although I'm not 100% certain) that our Windows packaging is using MSVC to build these. If this is the case, then I think MSVC is just ignoring these compiler options. It would be great if we could add optimization flags for Windows as well.

See the output from my Windows machine running `pip install -v -e hexrd` below:

```
    "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.28.29333\b
in\HostX86\x64\cl.exe" /c /nologo /O2 /W3 /GL /DNDEBUG /MD -IC:\Users\patrick\AppData\Local\Tem
p\pip-build-env-sa6fldvz\overlay\Lib\site-packages\numpy\core\include -Ic:\users\patrick\virtua
lenvs\hexrd-test\include -IC:\Users\patrick\sw\Python\Python39\include -IC:\Users\patrick\sw\Py
thon\Python39\Include "-IC:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tool
s\MSVC\14.28.29333\include" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\ucrt
" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared" "-IC:\Program Files (x
86)\Windows Kits\10\include\10.0.18362.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\
10.0.18362.0\winrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\cppwinrt" /T
chexrd/transforms\transforms_CAPI.c /Fobuild\temp.win-amd64-cpython-39\Release\hexrd/transforms
\transforms_CAPI.obj -O3 -ftree-vectorize
    cl : Command line warning D9002 : ignoring unknown option '-O3'
    cl : Command line warning D9002 : ignoring unknown option '-ftree-vectorize'
```

Unfortunately, getting our packaging CI to print the output of the build is likely a pain. We are using CPack to generate the various packages, and CPack only prints out the install output if it fails (see [here](https://gitlab.kitware.com/cmake/cmake/-/issues/23465) for my request to add an option to print it out without failure).

Then it may still not print out the build information unless we add a verbose flag somewhere, like I did for my `pip install -v ...` command above.

But either way, it would be great to verify that the optimization flags are working for all three operating systems.

(I'm assuming that our apple compiler accepts those compiler flags as well - we are probably using Apple clang)

psavery:
By the way, I just noticed in the above Windows output that it looks like the `Extension` class is already setting `/O2` automatically.

Then I tried it on Linux and I see it is apparently setting `-O2` automatically as well. Note all of the flags here:

```
  x86_64-linux-gnu-gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -fPIC -I/tmp/pip-build-env-t4dmhhec/overlay/lib/python3.10/site-packages/numpy/core/include -I/home/patrick/virtualenvs/hexrd/include -I/usr/include/python3.10 -c hexrd/transforms/transforms_CAPI.c -o /tmp/tmp9bdidjxd.build-temp/hexrd/transforms/transforms_CAPI.o
```
However, changing it to `-O3` in this PR is indeed providing a nice performance boost of ~10%, so it is a great addition.

setup.py
```diff
@@ -22,6 +23,15 @@
     'xxhash',
 ]
 
+# This will determine which compiler is being used to build the C modules
+compiler_type = distutils.ccompiler.get_default_compiler()
+if compiler_type in ("unix", "mingw32"):
+    compiler_optimize_flags = ['-O3', '-ftree-vectorize']
+elif compiler_type == "msvc":
+    compiler_optimize_flags = ['/Ox', '/Ob2', '/Oi', '/Ot', '/Oy', '/GL']
```
psavery, line 31:
```suggestion
    compiler_optimize_flags = ['/Ox', '/GL']
```

Note that according to [here](https://learn.microsoft.com/en-us/cpp/build/reference/ox-full-optimization?view=msvc-170#remarks), `/Ox` is the same as using `/Ob2`, `/Oi`, `/Ot`, and `/Oy` in combination, so I believe we can just remove these.

psavery (commented):
Looking good, let's just make the one suggested change and then I think it will be ready!

psavery:
For the record: I tried out Windows and did not see a noticeable performance improvement there. I did not try out Mac.

PR #553, opened 2023-08-09
fix weights and definition of chi^2

saransh13:
Recently discovered a bug in the statistical weights and definition of goodness of fit (chi^2). This was causing chi^2 to be less than 1 which should not be the case. Fixing those bugs. 

Reference:
Brian H. Toby, R factors in Rietveld analysis: How good is good enough? Powder Diffraction 21(1), March 2006.


psavery (approved):
I trust you!

PR #554, opened 2023-08-12
Performance tuning

ZackAttack614:
This PR includes a number of significant performance improvements in most workflows, in particular due to the fact that it now leverages the C++ libraries xsimd and Eigen to perform efficient math.

- The function [_ge_41rt_inverse_distortion](https://github.com/HEXRD/hexrd/blob/master/hexrd/distortion/ge_41rt.py#L17-L57) previously attempted to solve for the roots of a 6th-degree polynomial. However, as @piguy1010 noticed, degrees 4, 5, and 6 are always 2. Therefore, this function can be simplified to solve for the roots of a 3rd-degree polynomial instead, yielding a 4x speedup. [The updated Python code](https://github.com/HEXRD/hexrd/blob/performance-tuning/hexrd/distortion/ge_41rt.py#L18-L27) which does this simpler computation is included. However, using [the C++ version](https://github.com/HEXRD/hexrd/blob/performance-tuning/hexrd/transforms/cpp_sublibrary/src/inverse_distortion.cpp#L11-L22) yields a 2x speedup over the Python version. This is used by default in this branch.
- The transforms module, previously written in C, has been [partly refactored into C++](https://github.com/HEXRD/hexrd/blob/32cb8418503c231852953a93e0c7214530b7e47c/hexrd/transforms/cpp_sublibrary/src/transforms.cpp), heavily leveraging SIMD instructions. Using this built library yields an expected 1.5-4x performance increase, depending on the functions used. Note, in particular, the combination function (which needs a less verbose name) [anglesToGvecToDetectorXYFromAngles](https://github.com/HEXRD/hexrd/blob/performance-tuning/hexrd/transforms/cpp_sublibrary/src/transforms.cpp#L297-L320) which combines the logic of `makeOscillRotMat`, `anglesToGvec`, and `gvecToDetectorXY` to avoid unnecessary data conversion between numpy arrays in Python and C++ arrays, yielding an additional 1.25x speedup by itself.

More specific benchmarking results to follow, but in my testing with the ruby dataset, this PR cuts out an estimated 40% of the runtime in the single ruby dataset from the examples repository. Further analysis is required, as this was only tested on one machine with 20 CPU threads, and may not reflect results on a more diverse compute field.

Anecdotally, the simple change of `_ge_41rt_inverse_distortion` in pure Python to solve a 3rd degree polynomial gave Chris Budrow a 1.5x speedup in his workflow. So, I'm very eager to see the results of a benchmark including all of the changes in this PR.

@psavery I would like your advice in generating more robust unit testing for these changes.

### DIFF
--- a/.github/workflows/container_build.sh
+++ b/.github/workflows/container_build.sh
@@ -2,7 +2,12 @@
 cd /github/workspace/
 
 # Install dependencies
-yum install -y wget git centos-release-scl
+yum install -y wget git centos-release-scl eigen3-devel
+
+# Install xsimd headers
+git clone https://github.com/xtensor-stack/xsimd.git
+mkdir -p /usr/include/xsimd
+cp -r xsimd/include/xsimd/* /usr/include/xsimd/
 
 # Need to install packages that depend on centos-release-scl on a different line.
 # This will use gcc==10, which is the same as what manylinux2014 uses.
@@ -23,6 +28,7 @@ ${HOME}/miniconda3/bin/conda activate hexrd
 
 # Install conda build and create output directory
 ${HOME}/miniconda3/bin/conda install --override-channels -c conda-forge conda-build -y
+${HOME}/miniconda3/bin/conda install -c conda-forge pybind11 -y
 mkdir output
 
 # Build the package
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,12 @@ name: test
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, refactor ]
   pull_request:
     branches: [ master ]
+
+env:
+  EIGEN3_INCLUDE_DIR: /eigen3/eigen-3.4.0
 
 jobs:
   pytest:
@@ -59,6 +62,75 @@ jobs:
       run: |
           pip install fabio==0.12
       if: ${{ matrix.config.os == 'macos-latest'}}
+
+    - name: Install Eigen on Linux
+      run: |
+          sudo apt-get update
+          sudo apt-get install -y libeigen3-dev
+      if: runner.os == 'Linux'
+
+    - name: Install Eigen on MacOS
+      run: |
+          brew update
+          brew install eigen
+      if: runner.os == 'macOS'
+
+    - name: Install Eigen on Windows
+      run: |
+          Invoke-WebRequest -Uri https://gitlab.com/libeigen/eigen/-/archive/3.4.0/eigen-3.4.0.zip -o eigen.zip
+          Expand-Archive -Path eigen.zip -DestinationPath C:\eigen3
+          cp -r C:\eigen3\eigen-3.4.0 D:/a/hexrd/hexrd/eigen
+      if: runner.os == 'Windows'
+      shell: pwsh
+
+    - name: Install xsimd on Linux
+      run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake g++ wget unzip
+          wget -O xsimd.tar.gz https://github.com/xtensor-stack/xsimd/archive/refs/tags/7.4.10.tar.gz
+          tar -xf xsimd.tar.gz
+          cd xsimd-7.4.10
+          mkdir build
+          cd build
+          cmake ..
+          make -j$(nproc)
+          sudo make install
+      if: runner.os == 'Linux'
+      
+    - name: Install xsimd on macOS
+      run: |
+          brew install cmake wget
+          wget -O xsimd.tar.gz https://github.com/xtensor-stack/xsimd/archive/refs/tags/7.4.10.tar.gz
+          tar -xf xsimd.tar.gz
+          cd xsimd-7.4.10
+          mkdir build
+          cd build
+          cmake ..
+          make -j$(sysctl -n hw.physicalcpu)
+          sudo make install
+      if: runner.os == 'macOS'
+    
+    - name: Install xsimd on Windows
+      run: |
+          # Download and build xsimd from source
+          Invoke-WebRequest -Uri https://github.com/xtensor-stack/xsimd/archive/refs/tags/7.4.10.tar.gz  -o xsimd.tar.gz
+          tar -xf xsimd.tar.gz
+          mv xsimd-7.4.10 xsimd
+          cd xsimd
+          mkdir build
+          cd build
+          cmake -DCMAKE_INSTALL_PREFIX=/xsimd ..
+          cmake --build .
+          cmake --install .
+          echo "XSIMD_INCLUDE_DIR=/xsimd/include" >> $GITHUB_ENV
+      if: runner.os == 'Windows'
+      shell: pwsh
+
+    - name: Set up environment variables for Eigen3 and xsimd
+      run: |
+          echo "EIGEN3_INCLUDE_DIR=$(pwd)/eigen" >> $GITHUB_ENV
+          echo "XSIMD_INCLUDE_DIR=$(pwd)/xsimd/include" >> $GITHUB_ENV
+      if: runner.os == 'Windows'
 
     - name: Install HEXRD
       run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, performance-tuning ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,7 @@ name: test
 
 on:
   push:
-    branches: [ master, refactor ]
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -19,6 +19,8 @@ requirements:
     - python {{ python }}
     - setuptools
     - setuptools_scm
+    - eigen
+    - xsimd
 
   run:
     - appdirs
--- a/hexrd/distortion/ge_41rt.py
+++ b/hexrd/distortion/ge_41rt.py
@@ -9,52 +9,22 @@ if USE_NUMBA:
 from .distortionabc import DistortionABC
 from .registry import _RegisterDistortionClass
 from .utils import newton
+from hexrd.extensions import inverse_distortion
 
 RHO_MAX = 204.8  # max radius in mm for ge detector
 
 if USE_NUMBA:
-    @numba.njit(nogil=True, cache=True)
-    def _ge_41rt_inverse_distortion(out, in_, rhoMax, params):
-        maxiter = 100
-        prec = cnst.epsf
+    @numba.njit(nogil=True, cache=True, fastmath=True)
+    def _ge_41rt_inverse_distortion(inputs, rhoMax, params):
+        radii = np.hypot(inputs[:, 0], inputs[:, 1])
+        inverted_radii = np.reciprocal(radii)
+        cosines = inputs[:, 0]*inverted_radii
+        cosine_double_angles = 2*np.square(cosines)-1
+        cosine_quadruple_angles = 2*np.square(cosine_double_angles)-1
+        sqrt_p_is = rhoMax / np.sqrt(-(params[0]*cosine_double_angles+params[1]*cosine_quadruple_angles+params[2]))
+        solutions = (2/np.sqrt(3))*sqrt_p_is*np.cos(np.arccos(((-3*np.sqrt(3)/2)*radii/sqrt_p_is))/3+4*np.pi/3)
 
-        p0, p1, p2, p3, p4, p5 = params[0:6]
-        rxi = 1.0/rhoMax
-        for el in range(len(in_)):
-            xi, yi = in_[el, 0:2]
-            ri = np.sqrt(xi*xi + yi*yi)
-            if ri < cnst.sqrt_epsf:
-                ri_inv = 0.0
-            else:
-                ri_inv = 1.0/ri
-            sinni = yi*ri_inv
-            cosni = xi*ri_inv
-            ro = ri
-            cos2ni = cosni*cosni - sinni*sinni
-            sin2ni = 2*sinni*cosni
-            cos4ni = cos2ni*cos2ni - sin2ni*sin2ni
-            # newton solver iteration
-            for i in range(maxiter):
-                ratio = ri*rxi
-                fx = (p0*ratio**p3*cos2ni +
-                      p1*ratio**p4*cos4ni +
-                      p2*ratio**p5 + 1)*ri - ro  # f(x)
-                fxp = (p0*ratio**p3*cos2ni*(p3+1) +
-                       p1*ratio**p4*cos4ni*(p4+1) +
-                       p2*ratio**p5*(p5+1) + 1)  # f'(x)
-
-                delta = fx/fxp
-                ri = ri - delta
-                # convergence check for newton
-                if np.abs(delta) <= prec*np.abs(ri):
-                    break
-
-            xi = ri*cosni
-            yi = ri*sinni
-            out[el, 0] = xi
-            out[el, 1] = yi
-
-        return out
+        return solutions[:, None] * inputs * inverted_radii[:, None]
 
     @numba.njit(nogil=True, cache=True)
     def _ge_41rt_distortion(out, in_, rhoMax, params):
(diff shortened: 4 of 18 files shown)


ZackAttack614:
I'll work on fixing the `conda-package` workflow asap. It looks like the build container is missing some software dependencies to compile the new C++ code.

psavery:
@ZackAttack614 For the conda packaging, can you just try adding eigen, xsimd, and pybind11 [to the build dependencies](https://github.com/HEXRD/hexrd/blob/f7e2abda823596d37e5f9bf139453c6294bc5310/conda.recipe/meta.yaml#L16-L21) instead? It appears they are both available on conda-forge, and they both support all of the operating systems we want to support. I think we should strongly prefer doing that over having an extra step beforehand where we build/install dependencies. You might need to make sure that the `setup.py` file can find the headers of these libraries when they are installed from conda-forge.

The extra C++ dependencies might complicate [development installations](https://hexrdgui.readthedocs.io/en/latest/installing/#development). We want to keep development installations as simple as possible so that it is easy for others to contribute. We should try hard to completely avoid having a step beforehand where installing dependencies such as eigen and xsimd is required. For conda development installations, maybe we can do this by simply installing them from conda-forge (as mentioned in the paragraph above). 

Pip development installations will be more complicated. It would be ideal if eigen and xsimd had wheels available on [PyPI](https://pypi.org/). But I don't think they do. We could potentially look into automatically installing and compiling xsimd and eigen as a part of the setup.py file if we want to keep pip development installations simple. But we might also want to just stop supporting pip development installations for now. I think most users use conda anyways. But let's have a discussion about it first. The `test.yml` workflow in GitHub actions currently does a pip development installation, but we could convert that to a conda development installation instead if we chose to stop supporting pip development installations.

We should certainly add [tests](https://github.com/HEXRD/hexrd/tree/master/tests) for most of the functions modified here in this PR (especially the transforms functions). Since the transforms functions are used almost everywhere, we'll need to do a fair amount of testing before merging.

By the way, if you wanted to start using cmake, [scikit-build](https://scikit-build.readthedocs.io/en/latest/) is an excellent library for supporting python packages that include a cmake build step.

So here is a summary of things needed before this PR is merged:
- [ ] All of the CI passes
- [ ] Development installations do not require an extra install/compile step beforehand
- [ ] Decide whether to stop supporting pip development installations
- [ ] Add tests for most functions modified, especially the transforms functions
- [ ] I will test arm64 builds and installations (which are not a part of the CI currently)
- [ ] @donald-e-boyce should do a fair amount of testing for the HEDM side
- [ ] @saransh13 should do a fair amount of testing for the LLNL side

.github/workflows/container_build.sh
```diff
@@ -23,6 +28,7 @@ ${HOME}/miniconda3/bin/conda activate hexrd
 
 # Install conda build and create output directory
 ${HOME}/miniconda3/bin/conda install --override-channels -c conda-forge conda-build -y
+${HOME}/miniconda3/bin/conda install -c conda-forge pybind11 -y
```
psavery, line 31: Let's try to add these to the [build dependencies](https://github.com/HEXRD/hexrd/blob/f7e2abda823596d37e5f9bf139453c6294bc5310/conda.recipe/meta.yaml#L16-L21) instead of having an extra install/compile step beforehand.

.github/workflows/test.yml
```diff
@@ -60,6 +63,75 @@ jobs:
...
+          sudo apt-get update
+          sudo apt-get install -y cmake g++ wget unzip
+          wget -O xsimd.tar.gz https://github.com/xtensor-stack/xsimd/archive/refs/tags/7.4.10.tar.gz
+          tar -xf xsimd.tar.gz
+          cd xsimd-7.4.10
+          mkdir build
+          cd build
+          cmake ..
+          make -j$(nproc)
+          sudo make install
```
psavery, line 97: We might want to switch the `test.yml` workflow over to use a conda development installation instead of a pip development installation, especially if we decide to stop supporting pip development installations, so that the extra step of installing/compiling dependencies beforehand is not required.

hexrd/gridutil.py
```diff
@@ -62,31 +62,24 @@ def cellIndices(edges, points_1d):
...
-    # off_hi = idx >= len(edges) - 1
-    # if np.any(off_lo):
-    #     idx[off_lo] = np.nan
-    # if np.any(off_hi):
-    #     idx[off_hi] = np.nan
+    try:
+        return np.array(idx, dtype=int)
+    except Exception as e:
+        print(e)
+        pass
```
psavery, line 82: Let's remove this try/except before merging.

hexrd/imageseries/load/framecache.py
```diff
@@ -127,7 +127,7 @@ def shape(self):
         return self._shape
 
     def __getitem__(self, key):
-        return self._framelist[key].toarray()
+        return self._framelist[key]
```
psavery, line 130: Do we know why `.toarray()` was being used before?

hexrd/instrument/hedm_instrument.py
```diff
@@ -2721,15 +2706,14 @@ def _generate_ring_params(tthr, ptth, peta, eta_edges, delta_eta):
...
 def _run_histograms(rows, ims, tth_ranges, ring_maps, ring_params, threshold):
     for i_row in range(*rows):
         image = ims[i_row]
 
-        # handle threshold if specified
         if threshold is not None:
+            if sparse.issparse(image):
+                image = image.toarray()
             # !!! NaNs get preserved
-            image = np.array(image)
```
psavery, line 2732: I think this was being done to make a new copy of the array so that the original array doesn't get modified by the line below this one.

Are you sure it is okay to modify the original array here?

hexrd/transforms/cpp_sublibrary/Makefile
```diff
@@ -0,0 +1,26 @@
...
+$(TARGET_DIR)$(TARGET_NAME_TRANSFORMS)$(EXTENSION_SUFFIX): $(SRCS_TRANSFORMS)
+	$(CXX) $(CXXFLAGS) $(PYBIND_INCLUDES) $< -o $@ $(INCLUDES)
+
+$(TARGET_DIR)$(TARGET_NAME_INVERSE_DISTORTION)$(EXTENSION_SUFFIX): $(SRCS_INVERSE_DISTORTION)
+	$(CXX) $(CXXFLAGS) $(PYBIND_INCLUDES) $< -o $@ $(INCLUDES)
+
+.PHONY: clean transforms inverse_distortion
+clean:
+	rm -f $(TARGET_DIR)$(TARGET_NAME_TRANSFORMS)$(EXTENSION_SUFFIX) \
+				$(TARGET_DIR)$(TARGET_NAME_INVERSE_DISTORTION)$(EXTENSION_SUFFIX)
```
psavery, line 26: If you decide to use cmake instead, see [scikit-build](https://scikit-build.readthedocs.io/en/latest/)

hexrd/transforms/cpp_sublibrary/Makefile
```diff
@@ -0,0 +1,26 @@
+CXX = c++
+CXXFLAGS = -O3 -Wall -shared -std=c++20 -fPIC -funroll-loops -Wno-maybe-uninitialized
```
psavery, line 2: I'm not sure if we can require C++20. We should verify that it is available in all of the versions of the compilers we want to support.

hexrd/xrdutil/utils.py
```diff
@@ -1138,32 +1138,23 @@ def _project_on_detector_plane(allAngs,
...
-
-    det_xy = np.atleast_2d(tmp_xys[valid_mask, :])
-
-    # apply distortion if specified
     if distortion is not None:
-        det_xy = distortion.apply_inverse(det_xy)
+        tmp_xys = distortion.apply_inverse(tmp_xys)
 
-    return det_xy, rMat_ss, valid_mask
+    return tmp_xys, None, valid_mask
```
psavery, line 1157: The second return is now `None`? We should verify it isn't being used anywhere, and we should look into potentially removing it as well.

setup.py
```diff
@@ -62,6 +64,59 @@ def get_convolution_extensions():
...
+        if os.path.exists(full_path):
+            return full_path
+
+    raise Exception(f"The {library_name} library was not found in any known directory.")
+
+def get_cpp_extensions():
+    cpp_transform_pkgdir = Path('hexrd') / 'transforms/cpp_sublibrary'
+    src_files = [str(cpp_transform_pkgdir / 'src/transforms.cpp'), str(cpp_transform_pkgdir / 'src/inverse_distortion.cpp')]
+
+    extra_compile_args = ['-O3', '-Wall', '-shared', '-std=c++20', '-funroll-loops']
```
psavery, line 92: We should verify that C++20 is available on all of the versions of the compilers we want to support.

hexrd/instrument/hedm_instrument.py
ZackAttack614: In my testing, it seems that at this point `image` is actually a scipy sparse array, and casting it to a numpy array was important to do the thresholding. It's possible though that this function may be called with different argument types, in which case adding type constraints and checks would be important. I'd like to start adding type annotations to these functions to clear up any confusion about things like this, if that's a change we agree would be useful.

hexrd/transforms/cpp_sublibrary/Makefile
ZackAttack614: I agree with this fully. In fact, I think I can just flip this to C++11 without any difference in behavior. This must have been a holdover from my own local testing, I'm happy to remove it!

hexrd/xrdutil/utils.py
ZackAttack614: I see now that there is actually one place in hexrd which uses the second argument `rMat_ss`, [here](https://github.com/HEXRD/hexrd/blob/performance-tuning/hexrd/xrdutil/utils.py#L1549C17-L1549C23). I missed this in my assessment of this function earlier. I'll add this return value back in, excellent catch.

setup.py
ZackAttack614: Thanks for noting this. As mentioned in a previous comment, I'm happy to revert this to C++11.

hexrd/imageseries/load/framecache.py
ZackAttack614: Yes! This step actually converts the scipy sparse array into a numpy array. But when this array is used, the only values used are a small subset of the array itself. So, to save on memory usage and run time, I moved this logic to after the subarray is acquired before casting it from sparse array to numpy array.

Previous code:
```python
for i_frame in frame_indices:
    tmp = ome_imgser[i_frame][ijs[0], ijs[1]]
    contains_signal = contains_signal or np.any(
        tmp > threshold
    )
    patch_data_raw.append(tmp)
```

Modified code:
```python
for i_frame in frame_indices:
    tmp = ome_imgser[i_frame][ijs].toarray()
    contains_signal = contains_signal or np.any(
        tmp > threshold
    )
    patch_data_raw.append(tmp)```

File: hexrd/gridutil.py (hunk 3 above)
Comment (ZackAttack614) [reply hexrd/gridutil.py:82]: Yes, I'm not sure how this debugging step slipped in. Good catch, thank you!

File: .github/workflows/test.yml (hunk 2 above)
Comment (ZackAttack614) [reply .github/workflows/test.yml:97]: I'd love to discuss modifications to our current workflow in a call, if you're available!

**Comment by psavery** [comment]
@ZackAttack614 Looks like the current CI error is `ModuleNotFoundError: No module named 'pybind11'`. You might need to add that to the build dependencies too.

**Comment by psavery** [comment]
We are adding back HEXRD documentation in #555. The automatic readthedocs documentation requires installing HEXRD from a cloned repo, so we'll also need to make sure we can install it there without any issues as well.

We might need to change it to use a [conda environment instead of a pip environment](https://github.com/HEXRD/hexrd/pull/555/files#diff-cde814ef2f549dc093f5b8fc533b7e8f47e7b32a8081e0760e57d5c25a1139d9R6-R10). Or, we could alternative figure out a way to get it to skip the build step and create mock imports for the C++ modules. But using a conda environment and doing the actual build would probably be easier.

File: hexrd/imageseries/load/framecache.py (hunk 4 above)
Comment (psavery) [reply hexrd/imageseries/load/framecache.py:130]: Nice!

Comment (psavery) [reply hexrd/imageseries/load/framecache.py:130]: However, this changes the return type for one of the image series adapters' `__getitem__` function. I think all of the other image series adapters return a numpy array, is that right @donald-e-boyce? So, for instance, if you load data from an HDF5 file instead, it would be a numpy array, not a scipy sparse array.

Should we keep all of these consistent, @donald-e-boyce?

Comment (psavery) [reply hexrd/imageseries/load/framecache.py:130]: If we should make sure that all image series adapters return a numpy array, could we potentially just call it like this instead, @ZackAttack614?

```
tmp = ome_imgser[i_frame, ijs]
```

File: hexrd/instrument/hedm_instrument.py (hunk 5 above)
Comment (psavery) [reply hexrd/instrument/hedm_instrument.py:2732]: @ZackAttack614 I think scipy sparse arrays are actually only used for `.npz` files (and it might only be receiving a sparse array because of the change [here](https://github.com/HEXRD/hexrd/pull/554/files/32cb8418503c231852953a93e0c7214530b7e47c#r1292126145)). We use HDF5 files a lot in some other workflows that involve the GUI.

Joel did perform `np.array(x)` a lot for performing a copy, so I'm still wondering if he might have done this to avoid modifying the original array. I will try to look into this in more detail later.

File: setup.py [hunk 10]
```diff
@@ -70,12 +70,12 @@ def get_include_path(library_name):
         return env_var_hint
 
     possible_directories = [
+        sysconfig.get_path('include')+f'/../{library_name}'
```
psavery, line 73: I wonder if we should potentially find a more standard way for locating these, as this seems fragile. That might involve using CMake, although I'm not sure. But we can keep it if it is not likely to be fragile.

psavery: I wonder what other projects do that use the `xsimd` and `eigen` conda packages.

conda.recipe/meta.yaml
```diff
@@ -19,6 +19,8 @@ requirements:
     - python {{ python }}
     - setuptools
     - setuptools_scm
+    - eigen
+    - xsimd
```
psavery, line 23: I think we need to add `pybind11` here, and remove it from the other place where it is being installed.

.github/workflows/container_build.sh
```diff
@@ -2,7 +2,12 @@
 cd /github/workspace/
 
 # Install dependencies
-yum install -y wget git centos-release-scl
+yum install -y wget git centos-release-scl eigen3-devel
+
+# Install xsimd headers
+git clone https://github.com/xtensor-stack/xsimd.git
+mkdir -p /usr/include/xsimd
+cp -r xsimd/include/xsimd/* /usr/include/xsimd/
```
psavery, line 10: We'll get these from conda instead of installing them manually.

.github/workflows/test.yml
```diff
@@ -2,10 +2,13 @@ name: test
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, performance-tuning ]
```
psavery, line 5: We'll remove this change before merging.

.github/workflows/test.yml
```diff
@@ -2,10 +2,13 @@ name: test
...
   pull_request:
     branches: [ master ]
 
+env:
+  EIGEN3_INCLUDE_DIR: /eigen3/eigen-3.4.0
```
psavery, line 10: This will be found automatically in the conda libraries.

hexrd/transforms/cpp_sublibrary/src/inverse_distortion.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <pybind11/eigen.h>
+#include <Eigen/Core>
+#include <xsimd/xsimd.hpp>
+
+namespace py = pybind11;
+constexpr double FOUR_THIRDS_PI = 4.1887902;
+constexpr double N_THREE_HALVES_SQRT_3 = -2.59807621;
+constexpr double TWO_OVER_SQRT_THREE = 1.154700538;
+
+Eigen::ArrayXXd ge_41rt_inverse_distortion(const Eigen::ArrayXXd& inputs, const double rhoMax, const Eigen::ArrayXd& params) {
```
psavery, line 11: I believe that in some cases, we can avoid copying the arrays when going back and forth between numpy and Eigen. We should try to do that when possible. See [here](https://pybind11.readthedocs.io/en/stable/advanced/cast/eigen.html#eigen). We might want to consider passing arguments like this instead:
```
Eigen::Ref<Eigen::ArrayXXd> inputs
```

hexrd/distortion/ge_41rt.py
```diff
@@ -9,52 +9,22 @@
...
-                    break
-
-            xi = ri*cosni
-            yi = ri*sinni
-            out[el, 0] = xi
-            out[el, 1] = yi
-
-        return out
+    @numba.njit(nogil=True, cache=True, fastmath=True)
+    def _ge_41rt_inverse_distortion(inputs, rhoMax, params):
```
psavery, line 18: We will add one or two examples of inputs from the example you've been running to our testing suite for this function, and the other functions that have been modified.

hexrd/transforms/cpp_sublibrary/src/inverse_distortion.cpp
psavery: [Here](https://pypi.org/project/eigenpy/) is an interesting looking library that also claims to avoid copying when converting between numpy and eigen.

hexrd/xrdutil/utils.py
psavery: @ZackAttack614 and I discussed this, and since `rMat_ss` is not normally used, we will no longer compute it in this function, but we will add a new function where it gets computed.

ZackAttack614:
Closing this PR, since it's being broken into multiple PRs with siloed purpose.

PR #555, opened 2023-08-21
Revive the HEXRD readthedocs documentation

psavery:
A lot has changed since the documentation was last generated 6 years ago,
including sphinx and its extensions. Automatic documentation is now simpler,
so a lot of files previously present were no longer needed and are deleted in
this commit.

The process of generating the documentation is now quite simple. See the 
`docs/README.md` file for more info.

This overall seems to generate the documentation correctly.


psavery:
The documentation is now being built successfully on readthedocs and can be found [here](https://hexrd.readthedocs.io/en/revive-readthedocs/index.html).

bnmajor (approved):
:zombie: LGTM!

PR #557, opened 2023-09-13
Transfer calibration classes from HEXRDGUI to HEXRD

psavery:
These calibration classes are better located in HEXRD rather than in HEXRDGUI.

This also breaks up the `calibration.py` file into a package with several files, since it was getting large.


saransh13 (approved):
Tested it with some data. Calibration is running the same and results are correct.
LGTM!

PR #558, opened 2023-09-14
Backround refine

saransh13:
Ability to refine chebyshev background with other parameters after initial values are determined.
Needs additional changes in hexrdgui otherwise wppf package will be broken.


psavery (approved):
If it works well in your testing, then LGTM

PR #560, opened 2023-10-02
Ensure each invalid peak has a unique value

psavery:
Previously, all invalid peaks would have a value of `-999`. This would make them difficult to distinguish, however, in our spots viewer in HEXRDGUI.

Instead, now all invalid peaks still have a negative number (starting at `-100`), but they all have unique negative numbers. This allows us to ensure that the correct images are being displayed for whichever invalid peak the user wishes to see.


bnmajor (approved):
LGTM

Issue #556, opened 2023-09-13
Conda installation often takes too long


psavery:
Conda's dependency resolver sometimes takes a **very** long time to finish. Sometimes, it *never* finishes (several users have experienced this). I believe that even our GitHub Actions packaging CI takes a long time primarily because of the conda dependency resolver as well. We should come up with a plan for how to improve this. Here is the basic installation command (after you create a new environment):

```bash
conda install -c hexrd -c conda-forge python=3.10 hexrd
```

The one for HEXRDGUI is similar.

I wonder if there are certain things we can do to improve the performance of conda's dependency resolver?

An alternative solution would be to use mamba. Mamba is **extremely** fast in comparison, but the main negative is that you need to install and set up a mamba base environment separately beforehand.

psavery:
Apparently, mamba's solver is optionally available for use in conda, if you install it separately. It is actually [an officially supported solver](https://www.anaconda.com/blog/conda-is-fast-now).

If we wanted to get users to use it, we would have to add these commands, that they would run when the `hexrd` environment is active:

```bash
conda install -n base -c conda-forge conda-libmamba-solver
conda config --env --set solver libmamba
```

This would set the solver for the active environment to be `libmamba`. I tried this and it seems to run a lot faster this way.

We might also be able to speed it up even further if we do things like add a command to remove the `defaults` channel from the environment (we don't use anything in the defaults).

psavery:
Documentation was added for using the libmamba solver [here](https://hexrdgui.readthedocs.io/en/latest/installing/#mamba-solver).

PR #562, opened 2023-10-10
Try using libmamba solver for conda packaging

psavery:
It is [officially supported by conda](https://www.anaconda.com/blog/conda-is-fast-now), and it is significantly faster.

Before this commit, Windows packaging usually takes 1 hour and 15 minutes. Mac packaging usually takes about 45 minutes. Linux usually takes about 35 minutes.

We'll see if there are any noticeable differences.

Fixes: #556

### DIFF
--- a/.github/workflows/container_build.sh
+++ b/.github/workflows/container_build.sh
@@ -21,6 +21,10 @@ ${HOME}/miniconda3/bin/conda create --override-channels -c conda-forge -y -n hex
 ${HOME}/miniconda3/bin/activate hexrd
 ${HOME}/miniconda3/bin/conda activate hexrd
 
+# Install the libmamba solver (it is much faster)
+${HOME}/miniconda3/bin/conda install -n base -c conda-forge conda-libmamba-solver
+${HOME}/miniconda3/bin/conda config --set solver libmamba
+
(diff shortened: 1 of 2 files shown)


psavery:
I got libmamba working for the Linux packaging, and it reduced the time by about half to about 17 minutes! That's a great speedup.

However, I haven't figured out yet how to get it working for Windows and Mac. I am setting up the solver to be libmamba, but `conda build` is apparently still not using the libmamba solver.

psavery:
Well, for Mac and Windows, I had to install libmamba twice, and set it in the environment twice. I'm not sure why, but I've spent an hour or two trying to figure it out and was unable to figure it out. It has something to do with conda build's internal processes, which I do not know much about. I don't have time to continue exploring this, so we will leave it as is.

Anyways, it does make the packaging much faster for all operating systems! The packaging is nearly twice as fast for every OS. Windows took 40 minutes instead of 75 minutes. Mac took 18 minutes instead of 45 minutes. And Linux took 15 minutes instead of 35 minutes. This PR is ready to go! We should do something similar for HEXRDGUI.

bnmajor (approved):
Fantastic!! LGTM!

PR #564, opened 2023-10-11
make sure small rounding errors are dealt with correctly.

### LINKED ISSUE #563: invalid value in CalcAngle
RuntimeWarning: invalid value encountered in `arccos` in `hexrd.unitcell.CalcAngle`. Bounds need to be check to make sure the dot product is in the range `[-1, 1]`.

saransh13:
Fixes #563.


psavery (approved):
Note that this will also round some valid values to the bounds (i.e., 0.99999 will get rounded to 1). But if that is no problem, then this looks good to me!

saransh13:
Yep that is good!

PR #567, opened 2023-10-17
Add TBB as a dependency to HEXRD

psavery:
If TBB is installed, numba will use TBB instead of OpenMP.

In my brief testing, it appears that having TBB makes some of the workflows run a little faster. It also fixes some warnings we have been seeing about OpenMP.


psavery:
I confirmed that the output of `hexrd fit-grains` does not change whether TBB is used or OpenMP.

bnmajor (approved):
LGTM

PR #569, opened 2023-10-18
Build and recommend python3.11

psavery:
This removes support for python3.8 and adds support for python3.11, which is now recommended. python3.11 comes with several performance improvements that may benefit us.

python3.12 just came out a couple of weeks ago, but not all of our dependencies support it yet (it's not supported by: `fabio`, `numba`, and `silx` yet).


bnmajor (approved):
Nice!!

PR #573, opened 2023-10-27
Add option to set eta_step size for eta omega maps

psavery:
It was previously fixed to 0.25 degrees. However, the users may find it beneficial to change this.

Smaller eta_step sizes might lead to less overlap between neighboring peaks in the eta omega maps, for example.


psavery:
Although this works, I found an issue with modifying `eta_step`. Detectors that have a split in eta (i. e., contain `-180` next to `180`) don't appear correct for smaller eta values. I believe the issue is [here](https://github.com/HEXRD/hexrd/blob/380e7c40263471776eb1b2c789586e312337b72d/hexrd/instrument/hedm_instrument.py#L2691-L2727)

The problem is that for a detector split in eta, we need to remap the angles to be increasing with no gap. So, for instance, for dexelas, `ff2` needs to go from `98.5` degrees to `261.5` degrees.

Something in the logic in that function is breaking for small `eta_step` so that the eta values don't get ordered correctly, and that's making the histogram look wrong. We need to fix that issue.

psavery:
@donald-e-boyce This is working now.

bnmajor (approved):

psavery:
Unfortunately, removing the histogramming appears to have changed two things:

1. The places where eta doesn't lie on a detector used to be `np.nan`, but not after these changes (it is 0 instead)
2. The actual data appears different. The new data has fewer spots. I don't immediately know why.

psavery:
@donald-e-boyce helped me fix this.

The only potential issue is that, for a small eta step, there are some nans on the detector plates. However, indexing still worked fine. We will explore fixing this in the future, if needed.

PR #579, opened 2023-11-14
Only include "roi" in the config if it is not None

psavery:
The config is valid without the "roi". If the "roi" is None, we just should not include it at all.

This helps clean up the tree view a little bit in HEXRDGUI, so that there isn't an "roi" present when there is none.


bnmajor (approved):
:rocket:

PR #584, opened 2023-11-17
Inverse distortion

ZackAttack614:
This PR focuses on the GE inverse distortion function. It has been rewritten into C++ and uses Eigen to do the heavy lifting of the matrix math.

In the multi-ruby example analysis running on my laptop, this PR takes the total runtime from 82 seconds to 62 seconds, cutting out roughly 25% of the runtime.


tests/test_inverse_distortion.py
```diff
@@ -0,0 +1,102 @@
...
+                         [-52.77530392,  44.11918895],
+                         [-54.04540365,  42.37873249],
+                         [-74.88783308,  16.80210463],
+                         [-70.37458323, -25.96162026],
+                         [-38.38762237,  69.73166096],
+                         [-75.17679039,  15.06246417],
+                         [-89.99957766,  20.21009857],
+                         [-98.72150605, -10.57579081],
+                         [-82.31199676, -53.39597868],
+                         [-98.46565139, -12.31774635]])
```
psavery, line 66: I think we should also put these arrays in json files too.

psavery:
I should also test a couple of state files with these changes to verify that there are no issues. I believe this PR means that the transforms functions will now be compiled with a C++ compiler instead of a C compiler. In theory, this shouldn't cause any issues, but I will do a quick check to verify.

After the comments have been addressed and a few state files tested, it will be good to merge!

tests/test_inverse_distortion.py
ZackAttack614: I agree that they likely shouldn't be hard-coded. I added these so that we could consider including easily visible inputs and outputs for those trying to diagnose a problem. But I'm happy to exclude this.

tests/test_inverse_distortion.py
```diff
@@ -105,21 +34,19 @@ def test_large_input():
 
 
 def test_logged_data():
-    # Load logged data
     with open(
-        os.path.join(test_dir, "data", "inverse_distortion_in_out.pkl"), 'rb'
+        os.path.join(test_dir, 'data', 'inverse_distortion_in_out.json'),
+        encoding='utf-8'
```
psavery, line 39: I think I shouldn't have to specify the encoding - it uses `utf-8` by default.

psavery (approved):
I tried this out with the [`single_GE` example](https://github.com/HEXRD/examples/blob/master/NIST_ruby/single_GE/include/cstudy.yml), and the total time went from 2 minutes and 36 seconds to 2 minutes with the branch!

psavery:
I also tested a couple of state files and did not encounter any issues due to the fact that the compiler is C++ instead of C.

psavery:
At some point, we can verify that this produces a speed improvement for a real example with many grains, but it looks good, and we can verify that later.

PR #589, opened 2023-12-07
fixed bug in GrainData.load and added rotation_matrices attribute

donald-e-boyce:
This fixes a bug in the load() method; the keyword arguments needed a "**" in front. 

It also adds a method to compute the rotation matrices from the exponential map parameters. This is very useful since the orientations are always used as rotation matrices.  This was motivated by the recent discussions of near field processing.


psavery (approved):
Looks really great, Don. Thanks so much for adding a test, too!

PR #593, opened 2023-12-19
Add pressure and temperature calculations for lattice parameters

psavery:
The pressure and temperature parameters are set by reading in a JCPDS file, or by the user manually setting them.

These are then able to be used to compute lattice parameters at higher temperatures and pressures than at ambient.


saransh13 (approved):
LGTM!

Issue #595, opened 2024-01-04
Optimize findDuplicateVectors function


saransh13:
The `matirixutil.findDuplicateVectors` function needs to be `numba` accelerated for better interactivity in the GUI.

psavery:
Fixed by #596

PR #596, opened 2024-01-04
numba accelerated findDuplicateVectors

saransh13:
Fixes issue #595. The function is ~5-10x faster (after the first compile overhead).


psavery (approved):
LGTM! This produces a nice speed-up, and I verified that a few workflows produce the same results (including fit-grains).

PR #597, opened 2024-01-05
Remove transform section from these detectors

psavery:
The current translation of [0, 0, 0] doesn't work well with HEXRDGUI, since it indicates that the detector is on top of the sample.

We should make the transform section optional instead. If it is not provided, then HEXRDGUI will insert the default transform section. This will make the views work better...

If some detectors typically have a certain transform, it can still be provided and that will be used instead of the default.


bnmajor (approved):
Sounds good!

PR #598, opened 2024-01-09
Remove unused "calibration" directory

psavery:
This directory appears to be completely unused, and it is easy to confuse it with the real calibration directory in `fitting/calibration`.

Looking at the history, it doesn't look to me that they were ever used. Half of the files don't even import successfully.

Just remove the files.


saransh13 (approved):
LGTM

PR #602, opened 2024-01-10
Add initial pre-configured instrument

bnmajor:
This adds the "Dual Dexelas" instrument configuration

@psavery Any additional instruments we should include right now?


psavery:
I think this one is fine for now. But can we change it to use a yaml file instead of a `.hexrd` file? That is easier to read.

bnmajor:
> I think this one is fine for now. But can we change it to use a yaml file instead of a `.hexrd` file? That is easier to read.

Sure thing!

psavery (approved):
LGTM!

PR #603, opened 2024-01-12
ProcessedImageseries with frame_list automatically updates omega metadata

donald-e-boyce:
When a `ProcessedImageSeries` is instantiated with a `frame_list`, it now checks the metadata for an `omega` key. If `omega` is present, then the metadata is updated appropriately.


psavery (approved):
This looks good to me. Feel free to merge. I will verify it further when we have a real dataset.

PR #604, opened 2024-01-14
Imagefiles yaml text

donald-e-boyce:
This pull request allows you to load an imageseries of type `image-files` by passing the yaml text directly to the `open() `command instead of writing it to a file first and then loading the file.  This is just a convenience for writing raw data processing scripts.

I also have a test file that verifies this, but since it involves a large data file, I chose not to submit it here.  I suggest we add an imageseries directory to the hexrd-examples repository, where we can take advantage of existing large data files.

Here is the test script: 

```import numpy as np

import pytest


from hexrd import imageseries


@pytest.fixture
def image_files_fmt():
    return "image-files"


@pytest.fixture
def yaml_tmpl():
    return """image-files:
  directory: .
  files: RUBY_*
options:
  empty-frames: %(n_ef)d
  max-file-frames: %(maxff)d
  max-total-frames: %(maxtf)d
meta: {}
"""


def  test_yaml_text(image_files_fmt, yaml_tmpl, tmp_path):
    """Verify that yaml file and yaml text give same result"""
    d = dict(n_ef=1, maxff=10, maxtf=15)
    p = tmp_path / "image-files.yml"
    yaml_text = yaml_tmpl % d
    with open(p, "w") as f:
        f.write(yaml_text)
    ims_f = imageseries.open(p, image_files_fmt)
    ims_t = imageseries.open(yaml_text, image_files_fmt)

    assert len(ims_f) == len(ims_t)
    assert len(ims_f) == 15
    assert np.all(ims_f[0] == ims_t[0])
```

### DIFF
--- a/hexrd/imageseries/load/imagefiles.py
+++ b/hexrd/imageseries/load/imagefiles.py
@@ -30,10 +30,8 @@ class ImageFilesImageSeriesAdapter(ImageSeriesAdapter):
 
         Parameters
         ----------
-        fname: string | path
+        fname: string | Path
            name of YAML file or bytestring of YAML contents
-
-
         """
         self._fname = fname
         self._load_yml()
@@ -93,15 +91,18 @@ number of files: %s
         MAXFILF = 'max-file-frames'
         DTYPE = 'dtype'
         #
-        # Check whether fname is a filename or YAML content. If it has multiple
-        # lines, we consider it to be content, otherwise a file name.
+        # Check whether fname is a pathlib Path, a filename or YAML content.
+        # If it has multiple lines, we consider it to be YAML content,
+        # otherwise a file name.
         #
-        nlines = len(self.fname.splitlines())
+        is_str = isinstance(self.fname, str)
+        nlines = len(self.fname.splitlines()) if is_str else 1
         if nlines > 1:
             d = yaml.safe_load(self.fname)
         else:
             with open(self._fname, "r") as f:
                 d = yaml.safe_load(f)
+
         imgsd = d['image-files']
         dname = imgsd['directory']
         fglob = imgsd['files']
--- a/hexrd/imageseries/load/imagefiles.py
+++ b/hexrd/imageseries/load/imagefiles.py
@@ -25,13 +25,15 @@ class ImageFilesImageSeriesAdapter(ImageSeriesAdapter):
 
     format = 'image-files'
 
-    def __init__(self, fname, **kwargs):
+    def __init__(self, fname):
         """Constructor for image files image series
 
-        *fname* - should be yaml file with files and metadata sections
-        *kwargs* - keyword arguments
-                 . 'files' = a list of image files
-                 . 'metadata' = a dictionary
+        Parameters
+        ----------
+        fname: string | path
+           name of YAML file or bytestring of YAML contents
+
+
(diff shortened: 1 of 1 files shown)


hexrd/imageseries/load/imagefiles.py
```diff
@@ -81,13 +81,28 @@ def __str__(self):
...
-        with open(self._fname, "r") as f:
-            d = yaml.safe_load(f)
+        #
+        # Check whether fname is a pathlib Path, a filename or YAML content.
+        # If it has multiple lines, we consider it to be YAML content,
+        # otherwise a file name.
+        #
+        is_str = isinstance(self.fname, str)
+        nlines = len(self.fname.splitlines()) if is_str else 1
+        if nlines > 1:
```
psavery, line 100: Another alternative could be to check `os.path.exists(self.fname)`. If it's yaml, it wouldn't be an existing file path. But I think what is here works fine too!

psavery (approved):
Looks good to me! It's a great imporovement! 

In HEXRDGUI, we do create temporary yaml files in some places such as [here](https://github.com/HEXRD/hexrdgui/blob/0985fb5bdf28f1d7dbb002b6cfcd4eb15d8ad653/hexrdgui/image_file_manager.py#L95-L107). This PR will simplify that code so we don't have to make temporary files.

psavery:
@donald-e-boyce By the way, can we re-use the image series located [here](https://github.com/HEXRD/examples/tree/master/NIST_ruby/multiruby_dexelas/imageseries) for the tests? You could obtain them in the pytests via the `example_repo_path` fixture, like so:

```python
@pytest.fixture
def example_repo_imageseries_path(example_repo_path):
    return example_repo_path / 'NIST_RUBY/multiruby_dexelas/imageseries'

def test_name(example_repo_imageseries_path):
     # example_repo_imageseries_path will point to the directory containing the image series.
```

hexrd/imageseries/load/imagefiles.py
donald-e-boyce: I thought of checking whether the path exists first but chose not to do it because of the case when the user passes a filename but might have a mistake in it. In that case, the file would not exist, and so it would be treated as YAML text. Then I'm not sure what the error would be, probably some key not found in the YAML, when it really should be a `FileNotFoundError`.  That is why I first try to determine if is YAML text, then if it looks like a file name, it will try to load and return the `FileNotFoundError` if it does not exist.

psavery: That makes sense, Don

donald-e-boyce: Thanks also for the suggestion below with the pytest fixture for the data file in the examples repo. I didn't realize that was there and in the process learned about the pytest `conftest.py` file. That would be good to use, but at the moment, there are no image data files in the examples repo that would work because I am working on the 
`image-files` type of imageseries, and all the images in the examples are frame cache or HDF5.  

As I mentioned above, that gives me even more reason to set up some some examples of processing raw image data to make the frame caches.  Once I do that, then we could take advantage of it in the hexrd testing.

psavery: Sounds great, Don. Thank you!

PR #605, opened 2024-01-15
Frame cache writer

donald-e-boyce:
This allows you to write an imageseries in `frame-cache` format without specfying the `cache_file` keyword argument.

Before this change, you would pass the `filename` argument as well as the the `cache_file`  keyword arg, usually the same value in both places. Now, the cache_file is determined from the filename.  

The history of this is that the original format for `frame-cache` was a YAML file for the metadata, and the imagefiles were stored as sparse arrays in a numpy npz file, the cache file.  Currently the default format is to save all image data and metadata both in the same file, so a separate cache file is not necessary.  

For the YAML format, the cache file (if not an absolute path) is taken relative to the YAML file, so any directory  is prepended to the cache file.  For the single npz file format, this has the annoying affect that if you specify a directory in the YAML file and also in the cache file, the directory gets prepended again to the cache file, invalidating the path.  

I also added a `DeprecationWarning` for writing to the YAML format, since the single file format is concise, efficient and general enough for the metadata likely to be needed with the imageseries.

### DIFF
--- a/hexrd/imageseries/save.py
+++ b/hexrd/imageseries/save.py
@@ -85,8 +85,17 @@ class Writer(object, metaclass=_RegisterWriter):
         self._fname_base = tmp[0]
         self._fname_suff = tmp[1]
 
-    pass  # end class
+    @property
+    def fname(self):
+        return self._fname
 
+    @property
+    def fname_dir(self):
+        return self._fname_dir
+
+    @property
+    def opts(self):
+        return self._opts
 
 class WriteH5(Writer):
     fmt = 'hdf5'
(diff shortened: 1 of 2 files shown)


psavery (approved):
LGTM. This will also simplify HEXRDGUI code located [here](https://github.com/HEXRD/hexrdgui/blob/0985fb5bdf28f1d7dbb002b6cfcd4eb15d8ad653/hexrdgui/save_images_dialog.py#L83-L85) and [here](https://github.com/HEXRD/hexrdgui/blob/0985fb5bdf28f1d7dbb002b6cfcd4eb15d8ad653/hexrdgui/indexing/fit_grains_results_dialog.py#L718).

PR #606, opened 2024-01-16
Remove tbb requirement for non-intel machines

psavery:
tbb is not available on PyPI for non-intel machines. Only include it as a requirement in the setup.py file for intel machines.


bnmajor (approved):
LGTM

PR #607, opened 2024-01-19
Add xtensor support, remove Eigen

psavery:
We have decided to use xtensor instead of Eigen for the C++ code, for a few reasons:

1. Eigen doesn't really support tensors (3 or more dimensions) all that well, and we need tensor support. xtensor has better support.
2. Eigen uses Fortran-ordering, whereas xtensor uses C-ordering. Numpy also uses C-ordering, so we sometimes must make deep copies when converting to Eigen, but we do not need to for xtensor.
3. xtensor maintains xsimd, which we are already using anyways

Adding xtensor support is a great step towards accelerating other functions.

### DIFF
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -31,6 +31,7 @@ requirements:
     - pybind11
     - xsimd
     - xtensor
+    - xtensor-python
 
   run:
     - appdirs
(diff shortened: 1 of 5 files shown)


psavery:
@ZackAttack614 Ready for merging after you approve.

ZackAttack614 (approved):
This should be good for the context of this PR. The next PR related to the transforms module will include linear algebra operations, requiring [`xtensor-blas`](https://xtensor-blas.readthedocs.io/en/latest/usage.html) to be included in the conda installation. We can exclude that for now, to make sure we keep this prerelease as clean as possible for user testing.

Issue #608, opened 2024-01-22
Issue with Frame Cache Writing using Multiprocessing


darrencpagan:
I do not think the multiprocessor mapping function in imageseries framecache writing option is properly waiting for all processes to finish the operation that generates the sparse matrices from the data.

We are trying to write frame caches from raw-image data with 3600 frames, however the output sparse matrices only have three frames. When I modified the code to run sequentially on a single processor, everything was able to work. I looked online and there is conflicting information regarding the implementation of the thread executors as to whether it waits for all processes to finish.

psavery:
@darrencpagan I assume you mean [these lines](https://github.com/HEXRD/hexrd/blob/29e7c6f824544af0989292a5b9cbe28722dab174/hexrd/imageseries/save.py#L304-L305)?

The threads are supposed to finish before the `with` block is exited. I just tried this with a 1440 frame example, and it appeared that all frames were written. Are you able to provide me an example script/data, as we might be doing something slightly different?

darrencpagan:
Is there a best way to transfer a fairly large raw image (file) along with the scripts? I can provide on oneDrive or a cloud service. Maybe better for an email?

psavery:
@darrencpagan I can access it on `classe` if you put it there (if you do that, send me a message on slack with the filepath). Otherwise, I think a oneDrive/Google drive link will be fine (and share it with my email)!

darrencpagan:
I've shared a OneDrive with you both with the script and data that causes a problem. The frame-cache only saves 3 of 3601 frames.

Last missing piece of info is I'm working on a computer with 80 workers (40 cores operating 2 threads).

psavery:
Thank you for providing the example. I was able to reproduce the issue and determine the cause.

Because this is a raw image series, it [must be read in sequence](https://github.com/HEXRD/hexrd/blob/29e7c6f824544af0989292a5b9cbe28722dab174/hexrd/imageseries/load/rawimage.py#L115). An exception was being raised because the indices in the imageseries were not being accessed in order (via the threadpool).

However, because we were not evaluating the results of the `map()`, the exception was not being propagated - so you wouldn't see it.

Running it serially fixed the issue because that ensured that the indices were being accessed in order.

We will add some logic to fix this issue for writing a frame cache from a raw image series. And we will also modify the code to ensure that if an exception occurs, it will be propagated so that it will be visible to the user.

Thank you for reporting this issue, @darrencpagan!

darrencpagan:
[...]

PR #611, opened 2024-01-25
Make __getitem__ for raw imageseries threadsafe

psavery:
This ensures that for multithreaded situations (such as writing out a frame cache), `__getitem__` will be thread-safe. The change ensures that the frame will be obtained immediately after seeking, and no other thread can seek until the frame is obtained.

When I test the example Darren gave us (#608) on the master branch with no multithreading, it ran in 5m 37s. With this change, and using multithreading, it ran in 5m 13s. So the multithreading produced a minor speed increase.

Fixes: #608

### DIFF
--- a/hexrd/imageseries/load/rawimage.py
+++ b/hexrd/imageseries/load/rawimage.py
@@ -1,5 +1,6 @@
 """ Adapter class for raw image reader"""
 import os
+import threading
 
 import numpy as np
 import yaml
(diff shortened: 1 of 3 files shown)


psavery:
@donald-e-boyce Can you look over this? Feel free to run some of your tests to see if you notice a performance impact caused by the thread locking.

donald-e-boyce (approved):
This all looks good, and I tested it and it runs well.

PR #613, opened 2024-01-31
Add a test for gvec_to_xy

psavery:
Test data was taken from [here](https://github.com/ovillellas/xrd-transforms/blob/b94f8b2d7839d883829d00a2adc5bec9c80e0116/test_xrd_transforms/test_gvec_to_xy.py#L33).


ZackAttack614 (approved):
We can simplify some of the calculations here one day. Should we get around to it.

PR #614, opened 2024-02-02
Move projections into their own module

psavery:
xrdutil was getting too large, and we were running into circular import issues by referencing too many other parts of the code.

Moving projections into their own module helps prevent circular import issues since they will not be imported unless they are actually needed.


saransh13 (approved):
The refactor looks good. PolarView still needs to be made consistent with the API used hexrdgui though. That can be done in a separate PR.

PR #619, opened 2024-03-04
Add hedm_intensity to PlaneData

psavery:
This is computed nearly the same way as the powder intensity, except that it does not include multiplicity.

This PR also moves the lorentz and polarization functions to the crystallography module, where they are now used for computing the powder and HEDM intensities.


saransh13 (approved):
LGTM

psavery:
Finally got the Linux packaging fixed :slightly_smiling_face:

PR #624, opened 2024-03-29
Do not set values to 0 that are not referenced.

ZackAttack614:
Removed unnecessary setting of large arrays, roughly 6% estimated speedup for caking scripts.


psavery (approved):
Nice improvement! Looks good to me!

PR #625, opened 2024-04-02
Remove unnecessary variables, clean up warp_image

ZackAttack614:
Cleaned up some of the logic in `warp_image` directly. Expected speedup ~4-5%

### DIFF
--- a/hexrd/projections/polar.py
+++ b/hexrd/projections/polar.py
@@ -233,42 +233,35 @@ class PolarView:
 
         """
 
-        angpts = self.angular_grid
-        dummy_ome = np.zeros((self.ntth*self.neta))
+        gvec_angs = np.vstack([
+            self.angular_grid[1].flatten(),
+            self.angular_grid[0].flatten(),
+            np.zeros((self.ntth*self.neta))]).T
+        xypts = np.empty((len(gvec_angs), 2))
 
-        # lcount = 0
         img_dict = dict.fromkeys(self.detectors)
         for detector_id, panel in self.detectors.items():
             _project_on_detector = self._func_project_on_detector(panel)
-            img = image_dict[detector_id]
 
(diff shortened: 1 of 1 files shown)


hexrd/projections/polar.py
```diff
@@ -233,42 +233,35 @@ def warp_image(self, image_dict, pad_with_nans=False,
...
-                    dummy_ome]).T
+            args, kwargs = self._args_project_on_detector(gvec_angs, panel)
 
-            args, kwargs = self._args_project_on_detector(gvec_angs,
-                                                          panel)
-
-            xypts = np.nan*np.ones((len(gvec_angs), 2))
-            valid_xys, rmats_s, on_plane = _project_on_detector(*args,
-                                                                **kwargs)
+            valid_xys, _, on_plane = _project_on_detector(*args, **kwargs)
```
psavery, line 248: Could we just pass `valid_xys` into the interpolation function directly, rather than setting `xypts[on_plane, :] = valid_xys` below and passing `xypts`? It appears that the interpolation functions ignore invalid xys, and only operate on a subset of the valid xys (in their functions, they check which xys are on the panel, and ignore xys not on the panel, and they also apply the panel buffer mask to remove even more xys).

Can you verify that you get the same result that way? It might be faster also...

More thoughts: the interpolation functions call `clip_to_panel(xy, buffer_edges=True)`, which consists of two parts:
1. Removing any `xys` that are not on the panel
2. Applying the panel buffer mask to remove any masked `xys`

Since we already know that `valid_xys` are on the panel, it would be nice if we could just skip the check for whether the `xys` are on the panel. But we still need to apply the panel buffer mask (it hasn't been applied yet). Maybe we should make a separate way to apply the panel buffer mask so we can skip the extra check that `clip_to_panel()` performs... but only if we verify that the result stays the same.

PR #626, opened 2024-04-03
Stacking fault wppf

saransh13:
- Add stacking fault probability, $\alpha$ and twin probability, $\beta$ to the `wppf` module.
- Made the particle size and micro strain terms phase dependent.
- Removed anisotropic broadening, Xe and Ye for simplification.

### DIFF
--- a/hexrd/fitting/peakfunctions.py
+++ b/hexrd/fitting/peakfunctions.py
@@ -537,10 +537,10 @@ def _gaussian_pink_beam(p, x):
     g[~zmask] = \
         (0.5*(alpha*beta)/(alpha + beta)) * np.exp(u[~zmask])*t1[~zmask] \
         + np.exp(v[~zmask])*t2[~zmask]
-
     mask = np.isnan(g)
     g[mask] = 0.
-    g *= A
+    gmax = g.max()
+    g *= A/gmax
 
     return g
 
@@ -574,7 +574,8 @@ def _lorentzian_pink_beam(p, x):
 
     mask = np.isnan(y)
     y[mask] = 0.
-    y *= A
+    ymax = y.max()
+    y *= A/ymax
 
     return y
 
--- a/hexrd/fitting/peakfunctions.py
+++ b/hexrd/fitting/peakfunctions.py
@@ -539,8 +539,7 @@ def _gaussian_pink_beam(p, x):
         + np.exp(v[~zmask])*t2[~zmask]
     mask = np.isnan(g)
     g[mask] = 0.
-    gmax = g.max()
-    g *= A/gmax
+    g *= A / g.max()
 
     return g
 
--- a/hexrd/wppf/WPPF.py
+++ b/hexrd/wppf/WPPF.py
@@ -355,7 +355,6 @@ class LeBail:
                 t = t[allowed]
                 hkl = self.phases[p].hkls[allowed, :]
                 dsp = self.phases[p].dsp[allowed]
-                sf_f = sf_f[allowed]
                 tth_min = min(self.tth_min)
                 tth_max = max(self.tth_max)
                 limit = np.logical_and(t >= tth_min, t <= tth_max)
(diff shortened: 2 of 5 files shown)


hexrd/fitting/peakfunctions.py
```diff
@@ -537,10 +537,10 @@ def _gaussian_pink_beam(p, x):
     g[~zmask] = \
         (0.5*(alpha*beta)/(alpha + beta)) * np.exp(u[~zmask])*t1[~zmask] \
         + np.exp(v[~zmask])*t2[~zmask]
-
     mask = np.isnan(g)
     g[mask] = 0.
-    g *= A
+    gmax = g.max()
+    g *= A/gmax
```
psavery, line 543:
```suggestion
    g *= A / g.max()
```

This is shorter, but a not a necessary change.

hexrd/wppf/WPPF.py
```diff
@@ -340,22 +340,31 @@ def calctth(self):
...
                 hkl = self.phases[p].hkls[allowed, :]
                 dsp = self.phases[p].dsp[allowed]
+                sf_f = sf_f[allowed]
                 tth_min = min(self.tth_min)
                 tth_max = max(self.tth_max)
                 limit = np.logical_and(t >= tth_min, t <= tth_max)
                 self.tth[p][k] = t[limit]
                 self.hkls[p][k] = hkl[limit, :]
                 self.dsp[p][k] = dsp[limit]
+                if sf_f is not None and lfact_sf is not None:
```
psavery, line 365: I don't think `sf_f` could be `None`, since `sf_f = sf_f[allowed]` is above. But `sf_f` might have a `len()` of 0, or a `len()` less than or equal to the `limit` (which would cause an exception).

hexrd/wppf/WPPF.py
```diff
@@ -367,7 +376,7 @@ def initialize_Icalc(self):
 
         if self.intensity_init is None:
             if self.spectrum_expt._y.max() > 0:
-                n10 = np.floor(np.log10(self.spectrum_expt._y.max())) - 2
+                n10 = np.floor(np.log10(self.spectrum_expt._y.max())) - 1
```
psavery, line 379: Maybe the `-2` was a typo? :slightly_smiling_face:

hexrd/wppf/WPPF.py
```diff
@@ -434,77 +443,87 @@ def computespectrum(self):
...
                         tth,
                         dsp,
                         hkls,
-                        strain_direction_dot_product,
-                        is_in_sublattice,
                         tth_list,
                         Ic,
                         self.xn,
-                        self.wn,
+                        self.wn
```
psavery, line 494: I think comma at the end is good practice. You can keep it in the future.

hexrd/wppf/peakfunctions.py
```diff
@@ -417,9 +403,9 @@ def _gaussian_pink_beam(alpha,
     t2 = erfc(z)
     g = np.zeros(tth_list.shape)
     zmask = np.abs(del_tth) > 5.0
```
psavery, line 405: Looks like this `zmask` isn't actually used anymore? We can remove it if it is unused.

hexrd/wppf/phase.py
```diff
@@ -298,6 +342,31 @@ def CalcStar(self, v, space, applyLaue=False):
...
+        this function chooses a subset from a list 
+        of symmetrically equivalent reflections such
+        that there are no g and -g present.
+        """
+        klist = []
+        for i in range(ksym.shape[0]):
+            k = ksym[i,:]
+            kk = list(k)
+            nkk = list(-k)
+            if klist == []:
```
psavery, line 356:
```suggestion
            if not klist:
```

This is the standard way to check for an empty list.

psavery (approved):
Overall, code looks good to me. I'm assuming you tested it :slightly_smiling_face:

hexrd/wppf/WPPF.py
saransh13: -2 is intentional! This bit is trying to initialize the background and the starting assumption is that the background is $10^2$ times lower than the peak signal. This just seems to work better than $10^1$.

hexrd/wppf/peakfunctions.py
saransh13: Good catch. I think using the `zmask` makes the code a bit better behaved so I'll put it back.

hexrd/wppf/WPPF.py
saransh13: Fixed!

hexrd/wppf/WPPF.py
saransh13: Nice catch. Fixed now

PR #628, opened 2024-04-09
Refactor calibrators to work with lmfit

psavery:
Calibrators (instrument, powder, and laue) now utilize
lmfit for their calibration. This allows bound constraints to be
applied to the various parameters, and more complicated expression
constraints (such as engineering constraints) as well.

The instrument calibrator takes as arguments powder and laue calibrators.
The instrument calibrator creates its own lmfit parameters for the 
instrument. The powder and laue calibrators create their own lmfit
parameters for their respective parameters. All calibrators keep
track of the names of their own parameters, utilize these when performing
calibration, and update their parameter values.

The composite calibrator was completely removed, since it was almost
identical to the instrument calibrator (which is now preferred).


saransh13 (approved):
tested the changes. LGTM

PR #630, opened 2024-04-11
Use scipy's `convolve` for snip1d

psavery:
According to the [scipy documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve.html#scipy.signal.convolve), `convolve` will automatically pick which method to use depending on which method it thinks will be faster.

We were forcing it to use `fftconvolve` before, which, for our example, took about 1.588 seconds to run. Switching to `convolve` now takes about 0.468 seconds instead, so it is noticeably faster.

In our example, the final snip1d result had a max difference of `2.6e-13` compared to using `fftconvolve`, so it appears to have a minimal impact on the results.


saransh13 (approved):
LGTM

PR #632, opened 2024-04-16
Update ca certificates for centos Linux build

psavery:
Otherwise, wget is complaining that the certificates are expired when it tries to download miniconda from the anaconda website...


saransh13 (approved):
LGTM

PR #635, opened 2024-04-19
Invalidate sample layer past critical beta

psavery:
xy points on the detector past the critical beta have very high values and start appearing unrealistic. Just set any values to nan that are past the critical beta.

Note that we are not invalidating map values past the critical beta for now, because they actually appear okay when warping the polar view. This is because they move too far and disappear (which is actually what we want when warping to the polar view).


saransh13 (approved):
LGTM

PR #636, opened 2024-04-26
Downgrade packaging macos to 12

psavery:
This is the last version to use x86_64. The runners are starting to, by default, use arm64 (Apple silicon).


saransh13 (approved):
LGTM

PR #645, opened 2024-05-31
Rietveld phase fraction fix

saransh13:
The fix works provided all the phase fractions are refined simultaneously. The refinement of all phase fractions will be coupled in a future PR.


psavery (approved):
LGTM!

PR #647, opened 2024-06-19
Docstrings

kevindlewis23:
Doc strings added in xf_new_capi.  Also fixed a bug in [xy_to_gvec](https://github.com/Verdant-Evolution/hexrd/blob/441912346cbea972f86b3f979d412befd7e4e2a8/hexrd/transforms/new_capi/xf_new_capi.py#L339C1-L342C1) where a variable name was incorrect.

The [make_binary_rmat](https://github.com/Verdant-Evolution/hexrd/blob/5287b0a553892ba111731d3f1864dff0a56666c0/hexrd/transforms/new_capi/xf_new_capi.py#L449C1-L454C39) function does not have a doc string as I could not find any documentation about it.  Is it necessary to keep around, or is it unused?

 Also, in some testing, I found that the [unit_vector](https://github.com/Verdant-Evolution/hexrd/blob/5287b0a553892ba111731d3f1864dff0a56666c0/hexrd/transforms/new_capi/xf_new_capi.py#L343-L373) function is transposed as to what it was in the old xf api, which is what most of the functions use (it's nx3 instead of 3xn).  Is this intended?

### DIFF
--- a/hexrd/transforms/new_capi/xf_new_capi.py
+++ b/hexrd/transforms/new_capi/xf_new_capi.py
@@ -45,23 +45,23 @@ def angles_to_gvec(angs, beam_vec=None, eta_vec=None, chi=None, rmat_c=None):
     Parameters
     ----------
     angs : ndarray
-        The euler angles of diffraction. The last dimension must be 3.  In 
+        The euler angles of diffraction. The last dimension must be 3.  In
         (2*theta, eta, omega) format.
     beam_vec : ndarray, optional
-        Unit vector pointing towards the X-ray source in the lab frame.  
+        Unit vector pointing towards the X-ray source in the lab frame.
         Defaults to [0,0,-1]
     eta_vec : ndarray, optional
         Vector defining eta=0 in the lab frame.  Defaults to [1,0,0]
     chi : float, optional
         The inclination angle of the sample frame about the lab frame X.
     rmat_c : ndarray, optional
-        The change of basis matrix from the reciprocal frame to the crystal 
+        The change of basis matrix from the reciprocal frame to the crystal
         frame.  Defaults to the identity.
 
     Returns
     -------
     ndarray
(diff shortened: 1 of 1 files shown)


hexrd/transforms/new_capi/xf_new_capi.py
```diff
@@ -28,142 +29,364 @@
...
+    bMat,
+    wavelength,
+    vInv=None,
+    beamVec=cnst.beam_vec,
+    etaVec=cnst.eta_vec,
+):
     # this was oscillAnglesOfHKLs
     hkls = np.array(hkls, dtype=float, order='C')
     if vInv is None:
         vInv = np.ascontiguousarray(vInv_ref.flatten())
```
kevindlewis23, line 359: I think this is also a bug, vInv_ref doesn't seem to be defined.  But I don't think this funciton is used anyway,

hexrd/transforms/new_capi/xf_new_capi.py
```diff
@@ -28,142 +29,364 @@
...
+
+    Returns
+    -------
+    ndarray
+        (3,n) ndarray of n unit reciprocal lattice vectors (a.k.a. lattice
+        plane normals)
+
+    """
     assert hkl.shape[0] == 3, 'hkl input must be (3, n)'
     return unitVector(np.dot(bMat, hkl))
```
kevindlewis23, line 163: Undefined referenec here. Should be unit_vector, but I didn't change it because unit_vector works as the transpose of unitVector (so you either change unit_vector or do unitVector(np.dot(bMat, hkl).T).T). Also it's not part of the API

psavery (approved):
Thanks Kevin! This is really helpful. We need more docs like these...

kevindlewis23:
Recommendations incorporated, should be ready to push.

psavery (approved):

PR #649, opened 2024-06-21
Absorption correction

psavery:
This adds a physics package class along with code to perform absorption corrections due to components of the physics package or components on the detectors.


setup.py
```diff
@@ -12,6 +12,7 @@
 
 install_reqs = [
     'appdirs',
+    'chemparse',
```
psavery, line 15: We will need to add `chemparse` as a conda dependency as well.

saransh13 (commented):
Some minor issues with distortion which we will fix later. Otherwise LGTM

saransh13 (approved):

PR #650, opened 2024-06-21
fix error in fractional occupancy cases while reading cif file.

saransh13:
Same as description.

### DIFF
--- a/hexrd/material/material.py
+++ b/hexrd/material/material.py
@@ -718,7 +718,10 @@ class Material(object):
 
             chkstr = np.asarray([isinstance(x, str) for x in occ])
             occstr = np.array(occ)
-            occstr[chkstr] = 1.0
+            try:
+                occstr = occstr.astype(np.float64)
+            except:
+                occstr[chkstr] = 1.0
 
             atompos.append(np.asarray(occstr).astype(np.float64))
 


saransh13:
@psavery it seems like PEP8 is not happy with the bare except. Do you know of a better way to handle this part of the code?

psavery:
It wants you to specify the type of exception. In this case, it should only ever raise a `ValueError`, so you could do `except ValueError:`.

If you want to catch every possible exception, you can still do `except Exception`, which catches them all.

ZackAttack614:
[This](https://github.com/HEXRD/hexrd/blob/cif-fix-refactor/hexrd/material/material.py#L710-L716) may be a bit cleaner, if I am understanding the use case well enough:

```python
            occ = []
            for p in cifdata[occ_U[0]]:
                v = p.split('(')[0]

                occ.append(np.float64(v) if v.isnumeric() else 1.0)

            atompos.append(np.asarray(occ))
```

ZackAttack614:
@saransh13 I would like to close out of our dangling PRs before we can begin the v1.0 migration. Has there been activity on this? If not, I would like to propose closing it or wrapping up work to merge it in this week.

saransh13:
@ZackAttack614 I will update the PR later today.

saransh13:
@ZackAttack614 Your proposed changes fails for a simple test case with fractional occupancy. I prefer to keep the original code which was giving the right results. We can come back and simplify the code in a future PR.

ZackAttack614:
@saransh13 no problem at all, thanks so much for prioritizing this!

saransh13:
@ZackAttack614 Do you mind reviewing the PR? @psavery is on conference travel this week

ZackAttack614 (approved):
This looks all right, though going forward we'll be more rigid about making sure changes are covered by unit tests, indicated by the Codecov bot. @saransh13 if possible, could you add some unit test coverage for this function before merging? I'll approve this regardless, and if need be we can come back to that tomorrow when Kevin Lewis is around.

saransh13:
Yeah, I'd need some help from @kevindlewis23 to include it in the unit test. I have the data files and script that we need for that though. I'll merge this in the meantime since this is an important bug fix, and add the unit test as a future PR.

PR #651, opened 2024-06-24
Testcases

kevindlewis23:
Adding test cases for transforms/xf_new_capi


kevindlewis23:
validate_angle_ranges only returns true on all testcases.
gvec_to_xy  return NaN about half the time (which makes sense), but when testing against the old gvecToDetectorXY, the latter is always returning NaN.  Presumably there is something wrong with my testcases.

psavery:
> Check NaN cases in gvec_to_xy. Whenever both gvecToDetectorXYArray and gvec_to_xy return something, they return the same thing. However, NaN's do not match up

That's unfortunate :slightly_frowning_face:. This is part of the reason it has taken some time to integrate these.

ZackAttack614:
@donald-e-boyce I've heard from @kevindlewis23 that we're waiting on some analysis results from you to determine next steps on this. Can you comment? I would like to take care of our dangling PRs this week, so if needed please reach out to the team if assistance is needed to finish this up.

donald-e-boyce:
Zack, I've figured out what I need to,  but I need to put my stuff in test
format.  If you don't mind, I should have things mostly ready by Tuesday
meeting and done by the end of next week.

ZackAttack614:
No problem Don, thanks for the update! Definitely reach out if Kevin or I can help at all - Patrick's out until Monday, I believe.

kevindlewis23:
Moved to PR #696

PR #652, opened 2024-06-24
Add pinhole materials resource

psavery:
This contains all standard pinhole materials used in experiments. It is being transferred over from hexrdgui.

The list includes:

Mo
Pt
Ta
W
U6Nb

All of these were in the hexrdgui version except U6Nb.


saransh13 (approved):
LGTM

PR #654, opened 2024-06-27
Consolidate new transforms api

kevindlewis23:
Removed the xf-api decorators in xf_new_capi, as well as the transform_definitions.py file that they come from.  Also removed unused implementations of transform_definitions (numpy and numba).

Pros:
- Less code clutter
- Decorators don't affect functionality, could be confusing
- Decrease overhead (function signature checking for example)
- Changing parameter names for clarity no longer has to be done in multiple places
- With xf_new_capi, we had two full layers of interface before implementation, so this removes one layer.

Cons:
- For potential future implementations, having an interface completely separate from any implementation of it could be useful.
- Decorators had capability (unused) to check pre and post conditions of functions, separate from implementation.
- transform_definitions.py documented some unimplemented functions, this documentation is lost for potential future implementations (e.g. solve_omega. map_angle, quat_product_matrix)

Maybe deprecate other implementations instead of deleting for now


psavery (approved):
Looks good to me

kevindlewis23:
It sounds like the python stuff had a fewer newer algorithms,. It is possible people are using these, maybe deprecation makes sense?

psavery:
I think the Python versions Oscar was talking about are scattered throughout the code. I think these are probably not being used. They didn't exist in hexrd until I copied Oscar's repo over (and I don't think users would even know about them unless they went searching through the code).

psavery:
Since we can still retrieve these functions if we need them later, and since I think it's very unlikely they are being used by anyone, I'd so just go ahead and merge this PR. You can merge when ready.

Issue #571, opened 2023-10-19
Adding Coveralls or similar to auto-check new PR for unit tests


argerlt:
Is there a good reason for not implementing coveralls in the hexrd default PR actions?

For reference, coveralls gives the percentage of code called by unit tests, which is necessary for eventually ensuring actual code coverage, as opposed to assumed, and is super helpful for catching edge cases. 

Example from Orix, a different coding project I often work on:
https://github.com/pyxem/orix/issues/7

And example of where I made a pretty hefty commit:
https://github.com/pyxem/orix/pull/450/commits/462b513e568cd38033c3a04fef016000703234c1
and then coverall auto-caught an untested line for me, so I made a new unittest to address it:
https://github.com/pyxem/orix/pull/450/commits/15945593066500737f69f38bc6e03d18271a0356

Coveralls is also super easy to run on your own computer as a pre-flight check in command line:

https://orix.readthedocs.io/en/stable/dev/running_writing_tests.html

Its the sort of thing everyone should be doing before any PR, but often don't unless forced to. I can implement all this pretty quick, just want to check if there was a good reason not to.

Ideally, the end goal would be to completely cover hexrd, then make it so new PR cannot be merged to main unless they had unit tests, but short term, it would just be a report for users to self-check their code's completeness.

psavery:
I think that adding some sort of coverage would definitely be helpful. I'm personally more familiar with codecov.

But I'm also aware that we probably have pretty bad coverage, frankly. It's something we need to work on.

argerlt:
I believe codecov costs money, whereas coveralls is free as long as the repo is public. Other than that though, I believe they are functionally identical.

Also, semi-related, is there interest in adding auto-PEP-8 compliance checking, like python black? Again, since much of HEXRD is not currently compliant, wouldn't deny commits, would just give a line count for non-compliant lines.

I ask because it is easy to add both at the same time, since both can run as part of pytest

psavery:
> I believe codecov costs money, whereas coveralls is free as long as the repo is public. Other than that though, I believe they are functionally identical. (as in, if there is money and you would prefer codecov, I'm down to add that instead)

Codecov is actually free for open source projects. I'm not necessarily opposed to using coveralls, but I don't know much about it. I'm guessing it is pretty similar.

[...]

PR #656, opened 2024-06-27
Code coverage

kevindlewis23:
Codecov needs to be setup.  Get a token from [here](https://app.codecov.io/gh/Hexrd) and setup for hexrd.  Must be added as a workflow secret in settings.

Once it is, this PR adds a code coverage workflow and a status badge to the readme.

Maybe refactor to add the code coverage workflow to be part of the test workflow.  I did it separately because the test workflow runs on many OS's and that is unnecessary for coverage checks.  Also codecov needs an editable install of hexrd, and while I assume that doesn't change anything, the actual test cases should be as close to production as possible.

Fixes: #571


.github/workflows/codecov.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Run tests
+        env:
+            HEXRD_EXAMPLE_REPO_PATH: ${{ github.workspace }}/examples
+        run: |
+            pytest --cov hexrd --cov-report xml:coverage.xml tests
+        working-directory: hexrd
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
psavery, line 42: This token has been added.

psavery (commented):
What we have done on other projects is to include the codecov logic within the tests workflow, but only run it on one operating system (for instance, Linux). For example, for the `Upload coverage to Codecov` action, you can say `if: ${{ matrix.config.name == 'Linux'}}`, and there are things you can do to add the extra codecov arguments as well for only Linux (and not the other operating systems).

However, I'm not sure what to do about the editable install (`-e`) part. I guess we can keep this as a separate workflow if we want the main tests to not perform an editable install. The only issue is that, if we make workflow changes to the `test.yml`, they won't be reflected here. But we might need to just work with that anyways. What do you think?

kevindlewis23:
> However, I'm not sure what to do about the editable install (`-e`) part. I guess we can keep this as a separate workflow if we want the main tests to not perform an editable install. The only issue is that, if we make workflow changes to the `test.yml`, they won't be reflected here. But we might need to just work with that anyways. What do you think?

I think it makes more sense to move it actually.  I'm going to do that

psavery (approved):
Looks good to me

PR #659, opened 2024-07-01
Small coding style improvements

kevindlewis23:
No functional changes made here, mainly just removing unnecessary pass statements and putting underscores to mark unused variables.

### DIFF
--- a/hexrd/fitting/fitpeak.py
+++ b/hexrd/fitting/fitpeak.py
@@ -28,9 +28,7 @@
 import numpy as np
 # from numpy.polynomial import chebyshev
 
-from scipy import integrate
-from scipy import ndimage as imgproc
-from scipy import optimize
+from scipy import integrate, optimize, ndimage as imgproc
 
 from hexrd import constants
 from hexrd.imageutil import snip1d
@@ -58,7 +56,7 @@ def cnst_fit_obj(x, b):
     return np.ones_like(x)*b
 
 
-def cnst_fit_jac(x, b):
+def cnst_fit_jac(x, _b):
     return np.vstack([np.ones_like(x)]).T
 
 
@@ -66,7 +64,7 @@ def lin_fit_obj(x, m, b):
     return m*np.asarray(x) + b
 
 
-def lin_fit_jac(x, m, b):
+def lin_fit_jac(x, _m, _b):
     return np.vstack([x, np.ones_like(x)]).T
 
 
@@ -77,7 +75,6 @@ def quad_fit_obj(x, a, b, c):
 
 def quad_fit_jac(x, a, b, c):
     x = np.asarray(x)
-    return a*x**2 + b*x + c
     return np.vstack([x**2, x, np.ones_like(x)]).T
 
 
@@ -198,13 +195,13 @@ def fit_pk_parms_1d(p0, x, f, pktype='pvoigt'):
     weight = np.max(f)*10.  # hard coded should be changed
     fitArgs = (x, f, pktype)
     if pktype == 'gaussian':
-        p, outflag = optimize.leastsq(
+        p, _outflag = optimize.leastsq(
             fit_pk_obj_1d, p0,
             args=fitArgs, Dfun=eval_pk_deriv_1d,
             ftol=ftol, xtol=xtol
         )
     elif pktype == 'lorentzian':
-        p, outflag = optimize.leastsq(
+        p, _outflag = optimize.leastsq(
             fit_pk_obj_1d, p0,
             args=fitArgs, Dfun=eval_pk_deriv_1d,
             ftol=ftol, xtol=xtol
@@ -214,7 +211,7 @@ def fit_pk_parms_1d(p0, x, f, pktype='pvoigt'):
         ub = [p0[0]*2.0, np.max(x), 4.*p0[2], 1., 2.*p0[4], None]
 
         fitArgs = (x, f, pktype, weight, lb, ub)
-        p, outflag = optimize.leastsq(
+        p, _outflag = optimize.leastsq(
             fit_pk_obj_1d_bnded, p0,
             args=fitArgs,
             ftol=ftol, xtol=xtol
@@ -223,13 +220,13 @@ def fit_pk_parms_1d(p0, x, f, pktype='pvoigt'):
         lb = [p0[0]*0.5, np.min(x), 0., 0., 0., 0., 0., None]
         ub = [p0[0]*2.0, np.max(x), 4.*p0[2], 4.*p0[2], 1., 1., 2.*p0[4], None]
         fitArgs = (x, f, pktype, weight, lb, ub)
-        p, outflag = optimize.leastsq(
+        p, _outflag = optimize.leastsq(
             fit_pk_obj_1d_bnded, p0,
             args=fitArgs,
             ftol=ftol, xtol=xtol
         )
     elif pktype == 'tanh_stepdown':
-        p, outflag = optimize.leastsq(
+        p, _outflag = optimize.leastsq(
             fit_pk_obj_1d, p0,
             args=fitArgs,
             ftol=ftol, xtol=xtol)
@@ -404,6 +401,7 @@ def estimate_mpk_parms_1d(
 
     # case processing
     # !!! used to use (f[pt] - min_val) for ampl
+    
     if pktype == 'gaussian' or pktype == 'lorentzian':
         # x is just 2theta values
         # make guess for the initital parameters
--- a/hexrd/fitting/grains.py
+++ b/hexrd/fitting/grains.py
@@ -218,7 +218,6 @@ def objFuncFitGrain(gFit, gFull, gFlag,
             meas_omes = meas_xyo[:, 2]
             xy_unwarped = panel.distortion.apply(meas_xyo[:, :2])
             meas_xyo = np.vstack([xy_unwarped.T, meas_omes]).T
-            pass
 
         # append to meas_omes
         meas_xyo_all.append(meas_xyo)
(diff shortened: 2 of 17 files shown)


hexrd/fitting/fitpeak.py
```diff
@@ -77,7 +75,6 @@ def quad_fit_obj(x, a, b, c):
 
 def quad_fit_jac(x, a, b, c):
     x = np.asarray(x)
-    return a*x**2 + b*x + c
     return np.vstack([x**2, x, np.ones_like(x)]).T
```
kevindlewis23, line 78: Actual functionality change here, I assume the previous one was a bug

hexrd/fitting/fitpeak.py
```diff
@@ -58,15 +56,15 @@ def cnst_fit_obj(x, b):
     return np.ones_like(x)*b
 
 
-def cnst_fit_jac(x, b):
+def cnst_fit_jac(x, _b):
```
ZackAttack614, line 59: This function doesn't use `_b`. I'm not sure that changing the variable name is the right choice - if we're going to change this, we should consider fixing the signature to not have a second argument at all. Though, it may affect downstream user workflows unexpectedly.

hexrd/fitting/fitpeak.py
```diff
@@ -58,15 +56,15 @@ def cnst_fit_obj(x, b):
...
     return np.vstack([np.ones_like(x)]).T
 
 
 def lin_fit_obj(x, m, b):
     return m*np.asarray(x) + b
 
 
-def lin_fit_jac(x, m, b):
+def lin_fit_jac(x, _m, _b):
```
ZackAttack614, line 67: Same comment as prior - if `_m` and `_b` aren't being used by the function, we shouldn't have them.

hexrd/fitting/fitpeak.py
```diff
@@ -77,7 +75,6 @@ def quad_fit_obj(x, a, b, c):
 
 def quad_fit_jac(x, a, b, c):
     x = np.asarray(x)
-    return a*x**2 + b*x + c
```
ZackAttack614, line 80: Are we sure this is the incorrect return? The return that's staying doesn't seem to include `a`, `b`, or `c`.

hexrd/fitting/fitpeak.py
```diff
@@ -198,13 +195,13 @@ def fit_pk_parms_1d(p0, x, f, pktype='pvoigt'):
     weight = np.max(f)*10.  # hard coded should be changed
     fitArgs = (x, f, pktype)
     if pktype == 'gaussian':
-        p, outflag = optimize.leastsq(
+        p, _outflag = optimize.leastsq(
```
ZackAttack614, line 198: Is `_outflag` even used anywhere? If not, consider replacing all instances of it with `_`.

hexrd/fitting/grains.py
```diff
@@ -234,7 +233,7 @@ def objFuncFitGrain(gFit, gFull, gFlag,
         gHat_c = mutil.unitVector(np.dot(rMat_c.T, gVec_s))
 
         # !!!: check that this operates on UNWARPED xy
-        match_omes, calc_omes = matchOmegas(
+        _match_omes, calc_omes = matchOmegas(
```
ZackAttack614, line 236: If `_match_omes` is not being used, this should be replaced with `_`. Though, it is a bit strange that `matchOmegas` would be called, and the `_match_omes` result would not be used. Something smells off in the workflow.

hexrd/material/crystallography.py
```diff
@@ -1385,7 +1385,7 @@ def getHKLs(self, *hkl_ids, **kwargs):
         opts = dict(asStr=False, thisTTh=None, allHKLs=False)
         if len(kwargs) > 0:
```
ZackAttack614, line 1386: Could even remove the `if len(kwargs) > 0` check.

hexrd/nf_config/experiment.py
```diff
@@ -33,33 +33,38 @@ def max_tth(self):
     def comp_thresh(self):
         key = 'experiment:comp_thresh'
```
ZackAttack614, line 34: No need to set a variable for `key`, this can be removed - just refer to the string directly.

hexrd/nf_config/experiment.py
```diff
@@ -33,33 +33,38 @@ def max_tth(self):
...
         temp = self._cfg.get(key, None)
```
ZackAttack614, line 35: `{}.get(k)` defaults to `None` already if the key is not found. This can be `temp = self._cfg.get(key)`.

Or better yet, `temp = self._cfg.get('experiment:comp_thresh')`

hexrd/nf_config/experiment.py
```diff
@@ -33,33 +33,38 @@ def max_tth(self):
...
+            raise RuntimeError(
+                'comp_thresh must be None or a number between 0 and 1'
+            )
+
+        return temp
+
     @property
     def chi2_thresh(self):
         key = 'experiment:chi2_thresh'
         temp = self._cfg.get(key, None)
```
ZackAttack614, line 48: Same comments as prior - remove unnecessary `key` variable, remove `None` argument from `.get()`

hexrd/nf_config/experiment.py
```diff
@@ -33,33 +33,38 @@ def max_tth(self):
...
+                'chi2_thresh must be None or a number between 0 and 1'
+            )
+
+        return temp
 
 
     @property
     def misorientation(self):
         key = self._cfg.get('experiment:misorientation:use_misorientation')
         if key is True:
```
ZackAttack614, line 62: `if key is True` is killing me...

ZackAttack614: This should be typed to `bool` so that we can be sure `if key:` is safe to use instead. The danger is that it's a string, for example, and would evaluate to truthy. But we should just enforce that not to be possible.

ZackAttack614: There's also no need to have the `key` variable here either.

hexrd/projections/polar.py
```diff
@@ -251,7 +251,7 @@ def warp_image(self, image_dict, pad_with_nans=False,
                                                           panel)
 
             xypts = np.nan*np.ones((len(gvec_angs), 2))
-            valid_xys, rmats_s, on_plane = _project_on_detector(*args,
+            valid_xys, _rmats_s, on_plane = _project_on_detector(*args,
```
ZackAttack614, line 254: If `_rmats_s` is not used, this should be replaced with `_`.

hexrd/projections/spherical.py
```diff
@@ -105,7 +105,7 @@ def warp_eta_ome_map(self, eta_ome, map_ids=None, skip=10):
                 op.flatten()
             ]).T
 
-            ppts, nmask = zproject_sph_angles(
+            ppts, _nmask = zproject_sph_angles(
```
ZackAttack614, line 108: If `_nmask` is not used, this should be replaced with `_`.

hexrd/sampleOrientations/conversions.py
```diff
@@ -12,9 +12,7 @@
 
 @numba_njit_if_available(cache=True, nogil=True)
 def getPyramid(xyz):
-    x = xyz[0]
-    y = xyz[1]
-    z = xyz[2]
+    x, y, z = xyz
```
ZackAttack614, line 15: Please include a check that `len(xyz)` is 3 and is an unpackable object. Otherwise, this will raise a `ValueError`.

ZackAttack614 (changes requested):

donald-e-boyce:
I would hesitate to remove the Null class. In the base config file class,
where it is used, it sets the default get value to null, which is
explicitly not None.   I can imagine cases where using a default of None
has a specific meaning.  I don't have any in particular in mind, but I can
imagine such a case.

hexrd/fitting/fitpeak.py
kevindlewis23: I believe these functions are used by the optimizer, and this is the interface the optimizer uses

kevindlewis23:
Null change was reverted

hexrd/sampleOrientations/conversions.py
kevindlewis23: I assume it should raise an error anyway, but I'll make it a more meaningfully named error.

ZackAttack614:
I would like to consider removing the class. It seems its only use is as a substitute for `None`. I'm not convinced that it's useful.

kevindlewis23:
I did try to replace it with None but the test cases threw errors because the comparisons ended up not working.  I couldn't figure out why immediately, but I can try to fix it if you think it's worth it

kevindlewis23 (commented):
Accidental review, just supposed to be comments

hexrd/sampleOrientations/conversions.py
kevindlewis23: [Fix?](https://github.com/HEXRD/hexrd/blob/0d1a11fb0e71b1f4d3a3c5a125dca15bc5d9367b/hexrd/sampleOrientations/conversions.py#L15-L18)

hexrd/fitting/fitpeak.py
kevindlewis23: Same as above

donald-e-boyce:
> I would like to consider removing the class. It seems its only use is as a substitute for `None`. I'm not convinced that it's useful.

It's such a minor change, I don't think it's worth the effort.  I think the choice to not use None was intentional.

psavery:
Be sure to squash these commits, or cherry-pick the relevant ones.

ZackAttack614 (approved):

kevindlewis23:
I don't feel like resolving these merge conflicts

PR #660, opened 2024-07-01
Refactored two functions

kevindlewis23:
Refactor of computeIntersection and vecmvcobmatrix, as well as some small style changes.code fixes.  Both tested, one with permanent testcases written


psavery (approved):
Looks good to me!

psavery:
Looks like `mirrorlist.centos.org` is down, and that is causing the linux package to fail. Hopefully it will come back up soon...

bnmajor:
> Looks like `mirrorlist.centos.org` is down, and that is causing the linux package to fail. Hopefully it will come back up soon...

I think we'll need to bump the version: 
https://serverfault.com/questions/1161816/mirrorlist-centos-org-no-longer-resolve
https://www.redhat.com/en/topics/linux/centos-linux-eol

psavery:
Hmm... that's a shame. We could upgrade to centos 8, which would increase our glibc version to 2.28. But the problem is that there are still a lot of Linux computers out there that are still on glibc 2.17. It seems a lot of people (such as manylinux2014 python wheel standard) will still be using centos 7 for the time being for maximum compatibility on Linux systems.

It might be best to use an archive mirror instead, like it appears some people are doing.

psavery:
The packaging issue is now fixed in master. Feel free to rebase on master and push here, and it should fix the CI.

PR #662, opened 2024-07-01
Use archive mirror for CentOS 7

bnmajor:
CentOS 7 reached EOL 06/30/2024, and consequently mirrorlist.centos.org is no longer available. As detailed below, we would like to stay with CentOS 7 for now, so use the archive mirror instead.

### DIFF
--- a/.github/workflows/container_build.sh
+++ b/.github/workflows/container_build.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 cd /github/workspace/
 
+# Use archive mirror for CentOS 7 until we are ready to migrate to CentOS 8
+sed -i -e 's/mirrorlist/#mirrorlist/g' \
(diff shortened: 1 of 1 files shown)


psavery:
To add some info here: in order to make our packages as compatible as possible across as many Linux systems as possible, we need to build it with an older version of glibc. According to [these statistics](https://mayeut.github.io/manylinux-timeline/), nearly 26% of Linux users are still using glibc 2.17 or older. 

CentOS 7 uses glibc 2.17, and CentOS 8 uses glibc 2.28. Building on CentOS 7 will provide support to more Linux systems.

I think we ought to continue following the standards set by [manylinux](https://github.com/pypa/manylinux). For `manylinux2014`, they are continuing to use CentOS 7 with archived mirrors. We will do the same for now. We should follow them, and when a new manylinux standard comes out that uses a newer version of CentOS, we can upgrade as well.

psavery (approved):
LGTM

PR #682, opened 2024-07-15
Single grain fitting

### LINKED ISSUE #677: Cannot fit a single grain
Hello all,

Title says it all.  If you have a single grain in a grains.out and want to fit it we run into array shape issues.  I think this is something I saw even back in the early versions as well so a longer problem that may have a ticket already created (I didn't see one).  While it may not be something we do often, we should be able to fit a single grain at need.  Thoughts?  

Thanks - Sven

kevindlewis23:
Confirm this correctly understands what `grain_table` is.  This still passes the testcases, and if test_fitgrains is changed to remove the ndmin=2 (and making other necessary changes), it also passes, implying that this change does what it's supposed to do.

closes #677

### DIFF
--- a/hexrd/fitgrains.py
+++ b/hexrd/fitgrains.py
@@ -318,6 +318,7 @@ def fit_grains(cfg,
     even if there is only one grain or one process, so that it will be
     cancelable.
     """
+    grains_table = np.atleast_2d(grains_table)
 
     # grab imageseries dict
     imsd = cfg.image_series


gustafsonsven:
Was this unit tested against the multi-ruby datasets?  I don't see that specific test.  Just want to make sure this does not break a multi-grain grains table but changing the shape incorrectly.

gustafsonsven:
Ah I see.  Since it will not have any effect, and it allows a single grain to be fit correctly, then I don't see why we cannot push this through.  We should create an artificial mruby test case with a single grain just be eliminating the other two grains in the grains.out estimate.

psavery:
The only fit grains test we have is for the single ruby GE example, which actually only has one grain. However, it may be called in a different way such that a 2D grains_table is passed and avoids the error you were seeing @gustafsonsven. This change could potentially fix your issue while not affecting other cases. I'll approve and merge it.

PR #687, opened 2024-07-18
Rotations testcases

### LINKED ISSUE #684: rotations.py quatProduct doesn't work for (4,n) inputs, maybe.
If `rotations.py` `quatProduct` on (4,n) inputs is supposed to be element-wise composition, it does not work.  When running, it element-wise (seperately on (4,1) inputs), it matches running on (4,n) inputs on the first output quaternion only.

Can be reproduced with the following code:
```python
import numpy as np
from hexrd import rotations
def test_reproduce():
    n = 3
    # Generate random quaternions
    q1 = np.random.rand(4, n) * 2 - 1
    q2 = np.random.rand(4, n) * 2 - 1

    # Fix to unit rotation quaternions (this works and is tested)
    q1 = rotations.fixQuat(q1)
    q2 = rotations.fixQuat(q2)
    print("q1:")
    print(q1)
    print("q2:")
    print(q2)

    # Product using fixQuat()
    qp = rotations.quatProduct(q1, q2)
    print("qp:")
    print(qp)

    # Elementwise product using fixQuat()
    # Written as a regular loop for readability, can be done inline
    qp_el = None
    for i in range(n):
        a = np.atleast_2d(q1[:, i]).T
        b = np.atleast_2d(q2[:, i]).T
        prod = rotations.quatProduct(a, b)
        if qp_el is None:
            qp_el = prod
        else:
            qp_el = np.hstack((qp_el, prod))
    qp_el = np.array(qp_el)
    print("qp element-wise:")
    print(qp_el)
```

Example output is:
```
q1:
[[ 0.6639988   0.45485282  0.83234881]
 [-0.20387788 -0.44343047 -0.28927565]
 [ 0.55001834 -0.70874605 -0.24522852]
 [ 0.46370166  0.30685072 -0.40420048]]
q2:
[[ 0.40840558  0.72671763  0.36315318]
 [-0.30968977 -0.04341781 -0.40514601]
 [ 0.81406359 -0.34699685 -0.69307048]
 [-0.2731256   0.591261   -0.47289511]]
qp:
[[ 0.11305919  0.16252659  0.23675927]
 [-0.23880823 -0.26005055 -0.11490376]
 [-0.96445574  0.66319071  0.62761961]
 [-0.00365778 -0.6827422   0.73269079]]
qp element-wise:
[[ 0.11305919  0.11606473  0.17603383]
 [-0.23880823  0.02941978  0.27810212]
 [-0.96445574  0.9217511   0.69289508]
 [-0.00365778 -0.36883379  0.64153544]]

```

kevindlewis23:
Added some testcases for some of the math helpers in `rotations.py`.  Also fixed `quatProduct`, see [issue 684](https://github.com/HEXRD/hexrd/issues/684). After this PR confirms that the new testcases are working properly, I will rewrite some of these methods to use scipy.  I didn't want to change the methods at the same time as I added the testcases, because then there is no confirmation that the new implementations work.

closes #684

### DIFF
--- a/hexrd/rotations.py
+++ b/hexrd/rotations.py
@@ -276,20 +276,16 @@ def quatProduct(q1, q2):
             q2 = tile(q2, (1, n1))
             nq = n1
 
-    a = q2[0, ]
-    a3 = tile(a, (3, 1))
-    b = q1[0, ]
-    b3 = tile(b, (3, 1))
+    # Changed just to show that the testcases are correct, old implementation
+    # did not work when both inputs were (4, n)
+    # see https://github.com/HEXRD/hexrd/issues/684
+    # Will replace this with scipy rotation after showing that testcases
+    # are correct
 
-    avec = q2[1:, ]
-    bvec = q1[1:, ]
-
-    axb = zeros((3, nq))
+    qp = zeros((4, nq))
     for i in range(nq):
-        axb[:, i] = cross(avec[:, i], bvec[:, i])
-
-    qp = vstack([a*b - diag(dot(avec.T, bvec)),
-                 a3*bvec + b3*avec + axb])
+        q2_matrix = quatProductMatrix(q2[:, i].reshape(4, 1), mult='left')
+        qp[:, i] = dot(q2_matrix, q1[:, i])
 
     return fixQuat(qp)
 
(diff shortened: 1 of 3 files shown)


psavery (approved):
LGTM. Thanks Kevin!

PR #688, opened 2024-07-22
Refactor rotations.py to use scipy, much faster and cleaner

kevindlewis23:
Rewrote a lot of the rotation math in rotations.py to call the scipy methods that do similar things.  Benchamarking shows it's a lot faster and it cleans up the code a bit.

(I will definitely squash these changes before merging, there's so many because I was going back and forth between branches)


hexrd/rotations.py
```diff
@@ -178,18 +192,19 @@ def misorientation(q1, q2, *args):
 
     if len(args) == 0:
         # no symmetries; use identity
-        sym = (c_[1., 0, 0, 0].T, c_[1., 0, 0, 0].T)
+        sym = (np.c_[1.0, 0, 0, 0].T, np.c_[1.0, 0, 0, 0].T)
     else:
         sym = args[0]
```
ZackAttack614, line 197: We shouldn't have an `args` argument. Instead, replace it with `sym` with a default value of `None`. `args` adds a layer of unnecessary obfuscation.

kevindlewis23:
All of my actual functional changes were covered by tests by the way, what codecov is talking about is formatting changes from running Black, or the fact that I replaced the huge block of numpy function imports with just importing numpy, and then putting "np." in front of a bunch of thigngs.  I decreased the percentage of coverage for the project because I shortened code that was already covered by my previous PR.

hexrd/rotations.py
kevindlewis23: I could, I don't really want to mess with a function that I haven't written a testcase for.  My PR was focused on the mathematical rotation transformation methods, this change is only here for formatting reasons.

hexrd/rotations.py
```diff
@@ -401,33 +388,27 @@ def quatOfExpMap(expMaps):
...
-         tile(sin(0.5 * angs), (3, 1)) * axxs]
-    )
-    return quats
+    return _scipy_rotation_to_quat(R.from_matrix(r_mat))
 
 
 def quatAverageCluster(q_in, qsym):
-    """
-    """
+    """ """
```
ZackAttack614, line 411: This should be a real docstring if we're going to have a placeholder.

hexrd/rotations.py
```diff
@@ -536,159 +516,52 @@ def expMapOfQuat(quats):
...
-
-    rmat = C1*W + C2 * W2
-    rmat[:, 0, 0] += 1.
-    rmat[:, 1, 1] += 1.
-    rmat[:, 2, 2] += 1.
-
-    return rmat.squeeze()
-
-
-def rotMatOfExpMap_orig(expMap):
```
ZackAttack614, line 606: We can't technically be sure people aren't using this (why would they?) so let's deprecate this using the previous strategy instead of deleting it. We can remove the body of the function, and leave the signature with the deprecated decorator.

kevindlewis23: Fixed

psavery:
> I can also update rotMatOfQuat to use scipy and it will be faster (I ran some benchmarking tests). However, if you run it multiple times with the same input, the current implementation is much faster because of the Numba LRU cache. Will this be run with the same rotation matrices a lot? If not, I can change it, it won't matter that much for speed because nothing super slow is happening

The numba caching is actually caching the compiled code, and is thus a better thing to test against. Users will most often be using the pre-compiled code. The first time they run the application, it will be slower due to compiling, but since they will not be modifying the code, they will always be using the pre-compiled code after that.

kevindlewis23:
> The numba caching is actually caching the compiled code, and is thus a better thing to test against. Users will most often be using the pre-compiled code. The first time they run the application, it will be slower due to compiling, but since they will not be modifying the code, they will always be using the pre-compiled code after that.

Yeah thank you, I misunderstood what numba cache was.  I removed that from my original comment

hexrd/rotations.py
```diff
@@ -1096,48 +887,19 @@ def rmat(self, x):
...
-        None.
-
-        Notes
-        -----
-        1) This requires case-by-case implementations for all 24 possible
-           combinations of axes order and convention.
-        2) May be able to use SciPy to fill in some additional conventions.  As
-           for now, the api for a function that yields angles from simply takes
-           in a rotation matrix and yields the angles in radians.
+        None.WW
```
psavery, line 892:
```suggestion
        None
```

psavery (approved):

PR #691, opened 2024-07-23
Fix remove duplicate atoms

### LINKED ISSUE #686: Incorrect for Loop Implementation in unitcell.remove_duplicate_atoms
This for loop in [`remove_duplicate_atoms`](https://github.com/HEXRD/hexrd/blob/b6ae69fe20fca1f6cf48235cbeeb65d98bc8f6b5/hexrd/material/unitcell.py#L626-L629) appears to not be useful. It only does work on the first instance of the loop, then skips the rest.

My guess is that this is either supposed to be run with some kind of `else` that is missing, or this shouldn't be a for loop at all. Either way, this function is quite old and seems to me to be wrong.

Should this be fixed or removed?

saransh13:
The routine had some errors as pointed out by @ZackAttack614. More careful testing revealed some duplications and removal of sites with fractional occupancies. This PR is meant to fix these issues. There are a number of `break` statements in the function. So, stylistically the routine could be improved a little. In any case, this fixes #686.


hexrd/material/unitcell.py
```diff
@@ -616,7 +617,7 @@ def remove_duplicate_atoms(self,
         """
 
         if atom_pos is None:
-            atom_pos = self.atom_pos
+            atom_pos = copy.deepcopy(self.atom_pos)
```
ZackAttack614, line 620: Since you're not modifying `atom_pos`, I don't think this deep copy is necessary, is it?

ZackAttack614 (dismissed):
Before we merge this in, we need this to be covered by unit tests. @saransh13 if you could reach out to @kevindlewis23 about integrating more test coverage into the suite he's been extending, that would be great.

hexrd/material/unitcell.py
kevindlewis23: I saw this earlier as well.  At the end of this function you actually overwrite self.atom_pos anyway, so it shouldn't matter at all

kevindlewis23:
It sounds like you did some testing when you made this update, it shouldn't be very hard to incorporate those into unit tests.  If you add a file in the `tests` directory, pytest will automatically run any function whose name starts with `test_`, and make sure no errors are thrown (that all assert statements pass).  

You'll need to run `pip install -r tests/requirements-dev.txt` and set the environment variable "HEXRD_EXAMPLE_REPO_PATH" to poing to the hexrd examples repo.  Then you should be able to just run `pytest tests` or `pytest tests/[filename]` to run all the tests or your specific tests to make sure they pass.

It seems like you've already done some testing, and the ideal testing suite is always to have a few normal examples as well as some edge cases, but for cases where that is more difficult to do, I've randomly generated a bunch of testcases and written code to verify the results.  Let me know if you need any help with this.

ZackAttack614:
@saransh13 are you able to get to this soon? I'm happy to provide assistance if needed. We would like to get dangling issues resolved this week so that we can move forward with the hexrd v1.0 implementation.

hexrd/material/unitcell.py
saransh13: Thanks for catching this. I'll fix it.

saransh13:
@kevindlewis23 Here's the test script (there is an associated `.h5` file it reads data from that I can send you too. Some help in getting some test for this function would be very much appreciated!

```
from hexrd import material
from hexrd.material import load_materials_hdf5
import numpy as np
import h5py

f = h5py.File('testmat.h5')

mats = load_materials_hdf5(f)

print('=='*20)
print('original atomic positions')
print(mats['xtal1'].atom_pos)
mats['xtal1'].unitcell.remove_duplicate_atoms()
print('reduced atomic positions')
print(mats['xtal1'].atom_pos)

print('=='*20)
print('original atomic positions')
print(mats['xtal2'].atom_pos)
mats['xtal2'].unitcell.remove_duplicate_atoms()
print('reduced atomic positions')
print(mats['xtal2'].atom_pos)

print('=='*20)
print('original atomic positions')
print(mats['xtal3'].atom_pos)
mats['xtal3'].unitcell.remove_duplicate_atoms()
print('reduced atomic positions')
print(mats['xtal3'].atom_pos)
```

The output it produces is as follows:
```========================================
original atomic positions
[[0.  0.  0.  1. ]
 [0.5 0.  0.5 1. ]]
reduced atomic positions
[[0. 0. 0. 1.]]
========================================
original atomic positions
[[0.  0.  0.  0.5]
 [0.  0.  0.  0.5]]
reduced atomic positions
[[0.  0.  0.  0.5]
 [0.  0.  0.  0.5]]
========================================
original atomic positions
[[0.         0.         0.         0.33333334]
 [0.         0.         0.         0.33333334]
 [0.         0.         0.         0.33333334]
 [0.5        0.         0.         1.        ]
 [0.5        0.5        0.25       1.        ]
 [0.5        0.5        0.         1.        ]]
reduced atomic positions
[[0.         0.         0.         0.33333334]
 [0.         0.         0.         0.33333334]
 [0.         0.         0.         0.33333334]
 [0.5        0.         0.         1.        ]
 [0.5        0.5        0.25       1.        ]]```

**Comment by saransh13** [comment]
@ZackAttack614 Do you mind reviewing this (sans the unit test)? I'll work with @kevindlewis23 in the coming weeks to have this covered.

File: tests/test_material.py [hunk 2]
```diff
@@ -21,6 +21,10 @@ def default_material():
 def test_materials_file(example_repo_path):
     return example_repo_path / 'NIST_ruby/single_GE/include/materials.h5'
 
+@pytest.fixture
+def test_material_file_duplicate_atoms():
+    return f'{test_data_dir}/testmat.h5'
```
kevindlewis23, line 26: This will throw an error I think, `test_data_dir` needs to be a parameter in this function

saransh13: Fixed it.

kevindlewis23:
The third test does not pass on the old implementation, was it wrong before?

saransh13:
It was only handling the case for two disordered atoms on the same site correctly e.g. CuZn. I updated it to correctly handle an arbitrary number of atoms e.g. High entropy alloys.

kevindlewis23:
With the PEP8 fixes (you can run the Black formatter which should fix it automatically) this should be good, but I think Zack needs to review this to merge since he requested changes.  I'm a little bit worried this is missing a few potential testcases.  I think there's only one that actually removes any atoms, and you never remove more than one.  It would also be useful to add a testcase where two atoms are in the same place exactly but has to remove it (not a disordered alloy)

psavery (approved):
Thanks for adding a test!

saransh13:
@kevindlewis23 I think the exact same position case is the trivial case (subset of the first test case). In any case, I have changed the positions to check the trivial case. This is ready to be merged now.

PR #693, opened 2024-07-29
Remove duplicates

kevindlewis23:
Removal of duplicate methods and some refactoring of matrixutil.  Let me know if this needs to be two separate pull requests

Also ignore the commit messages at the beginning, I'll squash these all but they are from a different branch


kevindlewis23:
`_buildDict` in `spacegroup.py` has a small difference to `_buildDict` in `symbols.py`, where `hstr` takes everything before the colon instead of the entire string, like `nstr`.  So, my current solution is to use the same string in both places bet let the separate buildDicts be run

hexrd/matrixutil.py
```diff
@@ -742,39 +729,15 @@ def _findduplicatevectors(vec, tol, equivPM):
...
-    """
-    mag = np.sqrt(v[0]*v[0] + v[1]*v[1] + v[2]*v[2])
-    return mag
+normvec = normvec3 = np.linalg.norm
+cross = np.cross
+determinant3 = np.linalg.det
+trace3 = np.trace
 
 
 def normalized(v):
```
ZackAttack614, line 739: This is a silly thing to have as a function

hexrd/rotations.py
```diff
@@ -92,16 +92,8 @@ def arccosSafe(temp):
...
-
-    temp[gte1] = 1
-    temp[lte1] = -1
-
-    ang = np.arccos(temp)
-
-    return ang
-
+    temp = np.clip(temp, -1.0, 1.0)
+    return np.arccos(temp)
```
ZackAttack614, line 96: Can probably combine these lines

ZackAttack614: I also really don't like that the argument name is `temp`. Should be something more useful, like `cos_angle` maybe?

hexrd/matrixutil.py
kevindlewis23: I agree, I just don't want to remove any functionality in case people are importing this.

hexrd/rotations.py
```diff
@@ -191,11 +184,11 @@ def misorientation(q1, q2, *args):
             % (q1.shape,)
         )
 
-    if len(args) == 0:
+    if symmetries is None:
         # no symmetries; use identity
         sym = (np.c_[1.0, 0, 0, 0].T, np.c_[1.0, 0, 0, 0].T)
     else:
-        sym = args[0]
+        sym = symmetries
```
ZackAttack614, line 191: Can simplify this by not setting `sym = symmetries` but instead just replacing all of the `sym`s with `symmetries` in the function. That would also get rid of the `else` clause.

hexrd/rotations.py
kevindlewis23: Fixed

psavery (approved):
LGTM, although you'll probably have some merge conflicts after you merge the transforms one.

PR #694, opened 2024-07-29
Remove scripts folder

kevindlewis23:
I was told "no one uses these and that they should be removed.

Check if `Hydra_Feb19.yml` was moved to the right place


donald-e-boyce:
I would hold off on that for now ask Kelly what she thinks.  I don't think
the scripts are maintained and I'm not sure any actually fully work, but
these are the kind of things that people may use parts of. So we should
save this for discussion the next HEXRD meeting.

kevindlewis23:
> I would hold off on that for now ask Kelly what she thinks. I don't think the scripts are maintained and I'm not sure any actually fully work, but these are the kind of things that people may use parts of. So we should save this for discussion the next HEXRD meeting.

Kelly specifically said to remove it.  I'll add her as a reviewer so she can document the reasoning maybe

donald-e-boyce:
That's fine then!

psavery:
Saransh approved as well.

psavery (approved):

PR #700, opened 2024-08-09
Eigen transforms

kevindlewis23:
Faster implementations for some of transforms.


kevindlewis23:
The outputs will be fortran ordered, which only matters if you try to send it to C code without running np.ascontiguousarray first.

PR #701, opened 2024-08-09
Crystallography Refactored

kevindlewis23:
Refactored crystallography.  Some functions were removed or renamed, I aliased some stuff but not everything, as it would've been extremely long.


hexrd/instrument/detector.py
```diff
@@ -1131,14 +1131,6 @@ def make_powder_rings(
...
         # in case you want to give it tth angles directly
         if isinstance(pd, PlaneData):
-            # Okay, we have a PlaneData object
-            try:
-                pd = pd.makeNew()  # make a copy to munge
-            except TypeError:
-                # !!! have some other object here,
-                #     likely a dummy plane data object of sorts
-                raise
-
```
psavery, line 1141: I think this was being done before because they wanted a temporary version of the PlaneData object to modify (without modifying the PlaneData object that was passed in). See below where it is being modified. Are you sure this is safe to delete? It may have side effects.

tests/planedata/test_exclusion.py
```diff
@@ -24,10 +24,7 @@ def test_exclusion():
     assert pd.nHKLs == 2
     assert pd.getNhklRef() == 3
 
-    try:
-        # Exclude a range
-        pd.exclusions = [[0, 2]]
-        assert pd.nHKLs == 1
-        assert pd.getNhklRef() == 3
-    except NotImplementedError:
-        pass
```
psavery, line 33: @donald-e-boyce was this supposed to intentionally raise a `NotImplementedError`?

hexrd/instrument/detector.py
kevindlewis23: Not sure, I think Zack deleted that one.  It probably should be there

tests/planedata/test_exclusion.py
kevindlewis23: I wrote this test case.  It used to raise a NotImplementedError, but I implemented it so I removed the check in the test case

psavery: Sounds good.

hexrd/instrument/detector.py
kevindlewis23: I put it back.  I removed the useless try catch though.

psavery (approved):

PR #706, opened 2024-08-14
Resolve compiler flags for Windows

### LINKED ISSUE #612: Check if compile args are being used by Windows
These [compile arguments](https://github.com/HEXRD/hexrd/blob/81bc56155c4a8d754bad578841a13fab35a82db9/setup.py#L118-L119) are in a gcc-like format. But it looks like they are also being passed to the Visual Studio compiler. I have a feeling that Visual Studio might just be ignoring them. We should verify that Visual Studio is using these compiler arguments correctly.

ZackAttack614:
Resolves #612


psavery (approved):
LGTM

PR #702, opened 2024-08-12
Unitcell testcases

kevindlewis23:
Added testcases for some of unitcell.py, though I don't know enough about the rest to write testcases for it.  Also bugfix on TransSpace from @saransh13, and some minor reformatting.

PR #707, opened 2024-08-15
fix cartesian to reciprocal and direct space conversion

saransh13:
#702 revealed a bug in the cartesian --> reciprocal & direct space computation. Fixing the bug --all tests pass now.


saransh13:
This was already fixed in #702

PR #709, opened 2024-08-18
Fix sparsing from eiger stream files

psavery:
This fixes a few issues with creating sparse frame-cache arrays from eiger-stream-v1 files:

1. The dtype is now read correctly from the HDF5 file
2. Fixed treatment of arrays with empty shapes in the yaml dumper `NumpyToNativeDumper`
3. Save the nested eiger metadata as a yaml string, since frame caches do not yet support nested metadata


psavery:
This is now obsolete since #728 was merged.

PR #711, opened 2024-08-20
Origin choice for selected space groups

### LINKED ISSUE #559: Density of diamond cubic
Space group 227 i.e. diamond cubic materials have density twice the value it should be. The problem seems to be in the origin choice.

hexrd assumes origin choice 1. This means the Wyckoff position 8a with point group symmetry -43m is at (0,0,0).

Some sources on the internet assumes origin choice 2 (without noting it). This means the Wyckoff position 8a with symmetry -43m is at (1/8,1/8,1/8) and the position (0,0,0) has symmetry -3m.

saransh13:
Issue #559 pointed out that the density of diamond was 2x the theoretical density. It was traced back to the default choice of the origin. `hexrd` had origin choice 1 by default, while most CIF files available online assume origin choice 2.

The following changes have been made in this PR:
- All 24 space groups with two origin choices have been explicitly listed
- If user does not specify origin choice, then origin choice 2 is assumed by default for these 24 space groups.
- During `h5` file I/O, the origin choice can be explicitly passed as an argument `sgsetting`. This will override the value from the file.
- 
Fixes #559 on the hexrd side. The GUI still needs to implement an input for the origin choice.


hexrd/material/symmetry.py
```diff
@@ -113,32 +115,39 @@ def MakeGenerators(genstr, setting):
...
+            for i in range(1, genmat.shape[0]):
+                generator = np.dot(sym2, np.dot(
+                    np.squeeze(genmat[i, :, :]),
+                    sym))
+                frac = np.modf(generator[0:3, 3])[0]
+                frac[frac < 0.] += 1.
+                frac[np.abs(frac) < 1E-5] = 0.0
+                frac[np.abs(frac-1.0) < 1E-5] = 0.0
+                generator[0:3, 3] = frac
+                genmat[i, :, :] = generator
```
psavery, line 130: @saransh13 This is the only section with functional code changes, correct?

saransh13: This only affects materials with a different origin choice. What material is used in the failing test?

psavery (approved):

PR #715, opened 2024-09-23
Ignore Gimbal lock warnings from scipy

psavery:
We don't really need to see these, as they are not an issue for us. Hide the warnings.


bnmajor (approved):
LGTM!

PR #719, opened 2024-09-25
Remove point selection from WPPF

psavery:
This removes the need to import matplotlib from the code.
    
Now, it expects the background points to have been pre-selected elsewhere,
and the background points is the single argument for the `spline` background
method.

Corresponding branch in the GUI: hexrd/hexrdgui#1728


saransh13 (approved):
Tested spline point picking with some APS data. Looks very nice!

PR #720, opened 2024-09-28
Replace previously archived functions

ZackAttack614:
This PR will replace some previously archived functions from the `hexrd/material/crystallography.py` script in the current pre-release.

### DIFF
--- a/hexrd/material/crystallography.py
+++ b/hexrd/material/crystallography.py
@@ -27,6 +27,8 @@
 # =============================================================================
 import re
 import copy
+import csv
+import os
 from math import pi
 from typing import Optional, Union, Dict, List, Tuple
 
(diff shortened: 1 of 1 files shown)


psavery (approved):
Thanks Zack. I'm assuming the contents of the functions did not change at all? If not, looks good to me!

ZackAttack614:
Yep, it's identical to the code that was previously archived. Thanks, Patrick!

saransh13:
Any reason we are re-merging these archived functions? There are duplicates in the codebase now for the structure factor calculator in `hexrd.wppf.xtal._calcxrsf`.

ZackAttack614:
@saransh13 Thanks for pointing this out. This change was originally a part of a non-final change identifying code for removal. We have replaced it to prevent others' workflows from failing in pre-release.

This will be reevaluated when the whole community will weigh in on a duplicate code strategy. More to come on that soon.

PR #721, opened 2024-09-30
Add support for multiple x-ray sources

psavery:
This adds support for multiple x-ray sources, which may have different beam vectors, energies, and distances.

This also adds support for calibrating to diffraction patterns from multiple x-ray sources, including for powder, Laue, and also structureless.


saransh13 (approved):
Amazing job. LGTM!

PR #722, opened 2024-10-01
Add option to cache coordinate map in PolarView projection

psavery:
This provides a significant speedup when the same projection is being used for multiple images. This happens when many images are taken without changing the parameters (instrument parameters, polar resolution settings, etc.) that are relevant for generating the polar view.

By passing `cache_coordinate_map=True`, you acknowledge that you will not change these parameters. The cache will be generated on `PolarView` initialization, which is potentially time-consuming, and the cache will be re-used for all subsequent projections.

This was shown to provide a ~17x speedup for each projection after the cache has been generated.


saransh13 (approved):
LGTM

PR #723, opened 2024-10-15
Add Relative Constraints to Calibration

psavery:
This adds system relative constraints, where the entire detector system can be tilted/translated.

The center of rotation for the tilt is always about the mean center of the detectors.

This is an ideal setup if many subpanels are present, but their relative positions and tilts should remain fixed, while the entire system is allowed to calibrate.


saransh13 (approved):
LGTM

Issue #724, opened 2024-10-21
Do not apply pinhole effective area correction if there is no pinhole


saransh13:
Even if pinhole diameter and thickness is set to 0 (i.e. no pinhole), the correction is getting computed leading to `nans`.

PR #725, opened 2024-10-21
don't compute correction if no pinhole.

saransh13:
This fixes issue #724.

### DIFF
--- a/hexrd/instrument/detector.py
+++ b/hexrd/instrument/detector.py
@@ -1745,22 +1745,28 @@ class Detector:
     def calc_effective_pinhole_area(self, physics_package: AbstractPhysicsPackage) -> np.array:
         """get the effective pinhole area correction
         """
-        hod = (physics_package.pinhole_thickness /
-               physics_package.pinhole_diameter)
-        bvec = self.bvec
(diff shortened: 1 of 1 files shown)


hexrd/instrument/detector.py
```diff
@@ -1729,45 +1730,53 @@ def calc_physics_package_transmission(self, energy: np.floating,
...
-               physics_package.pinhole_diameter)
-        bvec = self.bvec
+        effective_pinhole_area = np.ones(self.shape)
 
-        tth, eta = self.pixel_angles()
-        angs = np.vstack((tth.flatten(), eta.flatten(),
-                  np.zeros(tth.flatten().shape))).T
-        dvecs = angles_to_dvec(angs, beam_vec=bvec)
+        if (physics_package.pinhole_diameter != 0.
+                and physics_package.pinhole_thickness != 0.):
```
psavery, line 1754:
```suggestion
        if (not np.isclose(physics_package.pinhole_diameter, 0)
            and not np.isclose(physics_package.pinhole_thickness, 0)):
```

I think we should use `np.isclose()` for float comparisons here too.

psavery (approved):
Looks good to me after using `np.isclose()`!

PR #727, opened 2024-11-07
Fix pinhole offset corrections

saransh13:
fix pinhole corrections for edge cases. All variables for PXRDIP and TARDIS should be same now. No special treatment of PXRDIP data needed.

### DIFF
--- a/hexrd/xrdutil/phutil.py
+++ b/hexrd/xrdutil/phutil.py
@@ -392,7 +392,8 @@ def calc_phi_x(bvec, eHat_l):
     """
     bv = np.array(bvec)
     bv[2] = 0.
-    if np.linalg.norm(bv) == 0.:
+    bv_norm = np.linalg.norm(bv)
+    if np.isclose(bv_norm, 0):
         return 0.
     else:
         bv = bv/np.linalg.norm(bv)
--- a/hexrd/xrdutil/phutil.py
+++ b/hexrd/xrdutil/phutil.py
@@ -396,7 +396,7 @@ def calc_phi_x(bvec, eHat_l):
     if np.isclose(bv_norm, 0):
         return 0.
     else:
-        bv = bv/np.linalg.norm(bv)
+        bv = bv / bv_norm
         return np.arccos(np.dot(bv, -eHat_l)).item()
 
 
--- a/hexrd/xrdutil/phutil.py
+++ b/hexrd/xrdutil/phutil.py
@@ -392,8 +392,11 @@ def calc_phi_x(bvec, eHat_l):
     """
     bv = np.array(bvec)
     bv[2] = 0.
-    bv = bv/np.linalg.norm(bv)
-    return np.arccos(np.dot(bv, -eHat_l)).item()
+    if np.linalg.norm(bv) == 0.:
+        return 0.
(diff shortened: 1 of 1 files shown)


saransh13:
@psavery This should solve our problem of having different eta angles for `PXRDIP` and `TARDIS` in the `rygg pinhole correction` functions. Can you please review?

hexrd/xrdutil/phutil.py
```diff
@@ -449,7 +452,7 @@ def _infer_eHat_l(panel):
 
     eHat_l_dict = {
         'TARDIS': -ct.lab_x.reshape((3, 1)),
-        'PXRDIP': ct.lab_y.reshape((3, 1))
+        'PXRDIP': -ct.lab_x.reshape((3, 1))
```
psavery, line 455: Maybe we can get rid of this function entirely if `PXRDIP` and `TARDIS` are the same.

hexrd/xrdutil/phutil.py
```diff
@@ -460,7 +463,7 @@ def _infer_eta_shift(panel):
 
     eta_shift_dict = {
         'TARDIS': -np.radians(180),
-        'PXRDIP': -np.radians(90),
+        'PXRDIP': -np.radians(180),
```
psavery, line 466: This one too. Maybe we can get rid of it also.

hexrd/xrdutil/phutil.py
```diff
@@ -392,8 +392,11 @@ def calc_phi_x(bvec, eHat_l):
     """
     bv = np.array(bvec)
     bv[2] = 0.
-    bv = bv/np.linalg.norm(bv)
-    return np.arccos(np.dot(bv, -eHat_l)).item()
+    if np.linalg.norm(bv) == 0.:
```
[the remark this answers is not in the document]
saransh13: I agree. I will accept your suggestions.

hexrd/xrdutil/phutil.py
saransh13: I think the FIDDLE instrument might have to be treated separately. I'm not 100% about that yet, but it's possible. So I'd keep the cases for now. We can remove in a future PR if the FIDDLE instrument doesn't need a separate case.

hexrd/xrdutil/phutil.py
saransh13: Same as above.

hexrd/xrdutil/phutil.py
```diff
@@ -392,8 +392,12 @@ def calc_phi_x(bvec, eHat_l):
     """
     bv = np.array(bvec)
     bv[2] = 0.
-    bv = bv/np.linalg.norm(bv)
-    return np.arccos(np.dot(bv, -eHat_l)).item()
+    bv_norm = np.linalg.norm(bv)
+    if np.isclose(bv_norm, 0):
+        return 0.
+    else:
+        bv = bv/np.linalg.norm(bv)
```
psavery, line 399:
```suggestion
        bv = bv / bv_norm
```

One last change.

psavery (approved):
I had one last change requested, but I'll go ahead and approve too.

PR #728, opened 2024-11-08
Add fch5 an hdf5-based format for storing framecaches

ChristosT:
The new format allows to store more complex datasets including  metadata at multiple levels.

The new fch5 file contains the following fields:
- `shape`,`nframes`,`dtype`  similar to before storing shape as tuple, nframes as integer and dtype of the imageseries as an encoded string.
- `data` an (m,1) array of type `dtype` holding the datavalues of all frames. `m` is evaluated on runtime based on `threshold` value
- `indices` an (m,2) array of type `np.uint16` holding the row and column indices of the datavalues for each frame.
- `frame_ids`: (2*nframes,1) holds the range that the i-th frame occupies in the above, i.e. the information of the i-th frame  can be accessed using:
```
data_i = data[frame_ids[2*i]:frame_ids[2*i+1]]
indices_i = indices[frame_ids[2*i]:frame_ids[2*i+1]]
```

Both the reader and the writer are multi-threaded and perform better in terms of space efficiency, speed and memory consumption during writes.

Improvements:

- memory consumption: During saving we write the framecache of each frame once it is ready instead of holding everything in memory. This reduces significantly the amount of memory required while processing a dataset for writing:

Maximum memory usage (as reported by GNU time) saving a raw 7.2G file: 
- npz  format: 18.63 Gb
- fch5 format: 1.8 Gb

- in terms of space efficiency, npz uses `zip`. For `fch5` we bring the [hdf5plugin](http://www.silx.org/doc/hdf5plugin/latest/index.html) package and chose the `blosc` compression filter with `zstd` as the underlying compression algorithm and `5` as the compression level. The choice of the compression configuration was based on trading off between size and speed while being competitive against the current `npz` writer. 

This results in the following sizes for the same raw file:

- npz: 816 Mb
- fch5: 661 Mb

and their corresponding saving time:

- npz: 177.43 s
- fch5: 106.45 s

 
 Note that similar to the npz case we store datavalues in their native format so they can be either integers or floats.

### DIFF
--- a/hexrd/imageseries/load/framecache.py
+++ b/hexrd/imageseries/load/framecache.py
@@ -6,10 +6,16 @@ from threading import Lock
 import numpy as np
 from scipy.sparse import csr_matrix
 import yaml
+import h5py
 
 from . import ImageSeriesAdapter
 from ..imageseriesiter import ImageSeriesIterator
 from .metadata import yamlmeta
+from hexrd.utils.hdf5 import unwrap_h5_to_dict
+
+import multiprocessing
+from concurrent.futures import ThreadPoolExecutor
+
 
 class FrameCacheImageSeriesAdapter(ImageSeriesAdapter):
     """collection of images in HDF5 format"""
@@ -26,13 +32,25 @@ class FrameCacheImageSeriesAdapter(ImageSeriesAdapter):
         self._framelist = []
         self._framelist_was_loaded = False
         self._load_framelist_lock = Lock()
+        # TODO extract style from filename ?
+        self._style = style.lower()
 
-        if style.lower() in ('yml', 'yaml', 'test'):
+        ncpus = multiprocessing.cpu_count()
+        self._max_workers = kwargs.get('max_workers', ncpus)
+
+        if self._style in ('yml', 'yaml', 'test'):
             self._from_yml = True
             self._load_yml()
-        else:
+        elif self._style == "npz":
             self._from_yml = False
             self._load_cache()
+        elif self._style == "fch5":
+            self._from_yml = False
+            self._load_cache()
+        else:
+            raise TypeError(f"Unknown style format for loading data: {style}."
+                            "Known style formats: 'npz', 'fch5' 'yml', ",
+                            "'yaml', 'test'")
 
     def _load_yml(self):
         with open(self._fname, "r") as f:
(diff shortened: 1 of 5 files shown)


ChristosT:
@psavery

hexrd/imageseries/load/framecache.py
```diff
@@ -26,13 +32,25 @@ def __init__(self, fname, style='npz', **kwargs):
...
             self._load_yml()
-        else:
+        elif self._style == "npz":
             self._from_yml = False
             self._load_cache()
+        elif self._style == "fch5":
+            self._from_yml = False
+            self._load_cache()
+        else:
+            raise TypeError("Unknown style format for loading data: {style}."
```
psavery, line 51:
```suggestion
            raise TypeError(f"Unknown style format for loading data: {style}."
```

Because of {style}, it looks like this is supposed to be an f-string, but it's not. Might need to add f before the start of the string.

hexrd/imageseries/load/framecache.py
```diff
@@ -45,6 +63,29 @@ def _load_yml(self):
...
+            self._load_cache_fch5()
+        else:
+            self._load_cache_npz()
+
+    def _load_cache_fch5(self):
+        with h5py.File(self._fname, "r") as file:
+            if 'HEXRD_FRAMECACHE_VERSION' not in file.attrs.keys():
+                raise NotImplementedError("Unsupported file."
+                                          "HEXRD_FRAMECACHE_VERSION"
+                                          "is missing!")
```
psavery, line 76:
```suggestion
                raise NotImplementedError("Unsupported file. "
                                          "HEXRD_FRAMECACHE_VERSION "
                                          "is missing!")
```

Need to add spaces, otherwise the message will look like "Unsupported file.HEXRD_FRAMECACHE_VERSIONis missing!".

hexrd/imageseries/load/framecache.py
```diff
@@ -45,6 +63,29 @@ def _load_yml(self):
...
+            version = file.attrs.get('HEXRD_FRAMECACHE_VERSION', 0)
+            if version != 1:
+                raise NotImplementedError("Framecache version is not"
```
psavery, line 79:
```suggestion
                raise NotImplementedError("Framecache version is not "
```

hexrd/imageseries/load/framecache.py
```diff
@@ -45,6 +63,29 @@ def _load_yml(self):
...
+                                          f"supported: {version}")
+
+            self._shape = file["shape"][()]
+            self._nframes = file["nframes"][()]
+            self._dtype = np.dtype(file["dtype"][()])
+            self._meta = dict()
```
psavery, line 85:
```suggestion
            self._meta = {}
```

I think this is more common.

setup.py
```diff
@@ -17,6 +17,7 @@
     'fast-histogram',
     'h5py<3.12',  # Currently, h5py 3.12 on Windows fails to import.
                   # We can remove this version pin when that is fixed.
+    'hdf5plugin',
```
psavery, line 20: We'll need to add this to the conda-forge dependencies too.

hexrd/imageseries/save.py
```diff
@@ -331,6 +363,117 @@ def extract_data(i):
...
+
+        # creating an array in memory will fail if data is too big or threshold
+        # too low, so we write to the file while iterating the frames
+        with h5py.File(self.cache, "w") as h5f:
+            h5f.attrs['HEXRD_FRAMECACHE_VERSION'] = 1
+            h5f["shape"] = shape
+            h5f["nframes"] = nframes
+            h5f["dtype"] = str(self._ims.dtype).encode()
+            metadata = h5f.create_group("metadata")
+            unwrap_dict_to_h5(metadata, self._process_meta())
```
psavery, line 403:
```suggestion
            unwrap_dict_to_h5(metadata, self._meta)
```

Can you just do this instead? I think we don't need `self._process_meta()` for the fch5 format.

We should also verify that this saves nested metadata correctly!

ChristosT:
This is tested with the gui and works as expected.

PR #729, opened 2024-11-15
cast only non-nan numbers to integers. 

saransh13:
Suppresses an annoying error that kept popping up.

`/Users/soderlind3/hexrd/hexrd/gridutil.py:90: RuntimeWarning: invalid value encountered in cast
  return np.array(idx, dtype=int)`

### DIFF
--- a/hexrd/gridutil.py
+++ b/hexrd/gridutil.py
@@ -87,7 +87,9 @@ def cellIndices(edges, points_1d):
     #     idx[off_lo] = np.nan
     # if np.any(off_hi):
     #     idx[off_hi] = np.nan
-    return np.array(idx, dtype=int)
+    idx = np.array(idx)
+    idx[np.isnan(idx)] = 0
+    return np.array(idx).astype(int)
 
 
 @numba.njit(nogil=True, cache=True)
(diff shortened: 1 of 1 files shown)


hexrd/gridutil.py
```diff
@@ -87,7 +87,9 @@ def cellIndices(edges, points_1d):
     #     idx[off_lo] = np.nan
     # if np.any(off_hi):
     #     idx[off_hi] = np.nan
-    return np.array(idx, dtype=int)
+    idx = np.array(idx)
+    idx[np.isnan(idx)] = 0
+    return np.array(idx).astype(int)
```
psavery, line 92:
```suggestion
    idx[np.isnan(idx)] = -1
    return idx.astype(int)
```

Since this function returns indices, and `0` is a valid index, I think we should potentially convert it to something else, right? It looks like if you have an array of `float64` and convert it to `int`, then nans get converted to `-9223372036854775808`, so that is what the behavior was before. But I think converting to `-1` might suffice.

Also, I don't think we need to make copies (by doing `np.array()`), since this variable was made locally, so I think we can skip that.

saransh13: The cast operation was converting `nans` to `0`. I was just preserving the behavior. I'm unsure if changing to `-1` will break anything.

psavery: Looks like different operating systems do different things...

psavery (approved):

PR #733, opened 2024-12-03
File placement improvements

donald-e-boyce:
The intention of this pull request is to make the output file placement in `find-orientations` and `fit-grains` cleaner and simpler while cleaning up computational code that has hardwired file names and file handling that would be better done elsewhere, _i. e._ the config module.

Roughly, this is what has been done:
- put output files in the `analysis_dir` instead of in the `working_dir` with long names based on the `analysis_name`
- allow for slashes in `analysis_name` and automatically create subdirectories as needed for the `analysis_dir`
    - this is particularly useful when the config file has multiple (YAML) documents
- move path manipulations and hardwired file names from computational modules to the config module
- use pathlib for handling file and directory names
- update docstrings where useful (for affected modules/functions)

You activate the new file placement with the `new_file_placement` keyword in the root level of the config file, e.g.

```
analysis_name: ruby/scan-0  

# working_dir:
new_file_placement: on

```
If the `new_file_placement`  is False (the default), then nothing is changed; the old file placement is used.

Originally, this was intended mainly for `find-orientations`, but `fit-grains` is slightly affected since it looks for a file written by `find-orientations`, _i.e._ the accepted orientations.

Running this on the multiruby and single_GE NIST_Ruby examples, I get the exact same results using the new or the old file placement.

### DIFF
--- a/hexrd/cli/find_orientations.py
+++ b/hexrd/cli/find_orientations.py
@@ -51,14 +51,8 @@ def configure_parser(sub_parsers):
 
 
 def write_scored_orientations(results, cfg):
-    # grab working directory from config
-    wdir = cfg.working_dir
-
-    scored_quats_filename = os.path.join(
-        wdir, '_'.join(['scored_orientations', cfg.analysis_id])
-    )
     np.savez_compressed(
-        scored_quats_filename,
+        cfg.find_orientations.orientation_maps.scored_orientations_file,
         **results['scored_orientations']
     )
 
--- a/hexrd/cli/find_orientations.py
+++ b/hexrd/cli/find_orientations.py
@@ -134,14 +134,9 @@ def execute(args, parser):
             quats_f
         )
         sys.exit()
-    if not os.path.exists(cfg.working_dir):
-        os.makedirs(cfg.working_dir)
 
     # configure logging to file
-    logfile = os.path.join(
-        cfg.working_dir,
-        'find-orientations_%s.log' % cfg.analysis_id
-        )
+    logfile = cfg.find_orientations.logfile
     fh = logging.FileHandler(logfile, mode='w')
     fh.setLevel(log_level)
     fh.setFormatter(
--- a/hexrd/cli/find_orientations.py
+++ b/hexrd/cli/find_orientations.py
@@ -50,33 +50,20 @@ def configure_parser(sub_parsers):
     p.set_defaults(func=execute)
 
 
-def write_scored_orientations(results, cfg):
+def write_results(results, cfg):
+    # Write scored orientations.
     np.savez_compressed(
         cfg.find_orientations.orientation_maps.scored_orientations_file,
         **results['scored_orientations']
     )
 
-
-def write_results(results, cfg):
-    # Write out the data
-    write_scored_orientations(results, cfg)
-
-    # grab working directory from config
-    wdir = cfg.working_dir
-
-    if not os.path.exists(cfg.analysis_dir):
-        os.makedirs(cfg.analysis_dir)
-    qbar_filename = os.path.join(
-        wdir,
-        'accepted_orientations_' + cfg.analysis_id + '.dat'
-    )
+    # Write accepted orientations.
+    qbar_filename = str(cfg.find_orientations.accepted_orientations_file())
     np.savetxt(qbar_filename, results['qbar'].T,
                fmt='%.18e', delimiter='\t')
 
-    # ??? do we want to do this by default?
-    gw = instrument.GrainDataWriter(
-        os.path.join(cfg.analysis_dir, 'grains.out')
-    )
+    # Write grains.out.
+    gw = instrument.GrainDataWriter(cfg.find_orientations.grains_file)
     for gid, q in enumerate(results['qbar'].T):
         phi = 2*np.arccos(q[0])
         n = xfcapi.unit_vector(q[1:])
@@ -118,16 +105,17 @@ def execute(args, parser):
     cfg = config.open(args.yml)[0]
 
     # prepare the analysis directory
-    quats_f = os.path.join(
-        cfg.working_dir,
-        'accepted_orientations_%s.dat' % cfg.analysis_id
-        )
-    if os.path.exists(quats_f) and not (args.force or args.clean):
+    quats_f = cfg.find_orientations.accepted_orientations_file(to_load=True)
+
+    if (quats_f is not None) and not (args.force or args.clean):
         logger.error(
             '%s already exists. Change yml file or specify "force" or "clean"',
             quats_f
         )
         sys.exit()
+
+    # Create analysis directory and any intermediates.
+    cfg.analysis_dir.mkdir(parents=True, exist_ok=True)
 
     # configure logging to file
     logfile = cfg.find_orientations.logfile
--- a/hexrd/cli/find_orientations.py
+++ b/hexrd/cli/find_orientations.py
@@ -1,11 +1,16 @@
 from __future__ import print_function, division, absolute_import
 
 import os
+import logging
+import sys
+
 import numpy as np
 
 from hexrd import constants as const
+from hexrd import config
 from hexrd import instrument
 from hexrd.transforms import xfcapi
+from hexrd.findorientations import find_orientations, write_scored_orientations
 
 
 descr = 'Process rotation image series to find grain orientations'
@@ -52,10 +57,7 @@ def configure_parser(sub_parsers):
 
 def write_results(results, cfg):
     # Write scored orientations.
-    np.savez_compressed(
-        cfg.find_orientations.orientation_maps.scored_orientations_file,
-        **results['scored_orientations']
-    )
+    write_scored_orientations(results, cfg)
 
     # Write accepted orientations.
     qbar_filename = str(cfg.find_orientations.accepted_orientations_file)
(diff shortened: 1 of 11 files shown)


hexrd/config/findorientations.py
```diff
@@ -231,12 +270,47 @@ def eta_step(self):
...
+
+        return mapf
+
+    @property
+    def scored_orientations_file(self):
+        root = self.parent
+        if root.new_file_placement:
+            adir = root.analysis_dir
+            actmat = root.find_orientations._active_material_str()
+            sof = Path(adir) / f'scored-orientations-{actmat}.npz'
```
psavery, line 308:
```suggestion
            sof = adir / f'scored-orientations-{actmat}.npz'
```

The analysis dir is guaranteed to be a `Path` object now, right?

hexrd/findorientations.py
```diff
@@ -388,56 +388,48 @@ def load_eta_ome_maps(cfg, pd, image_series, hkls=None, clean=False):
...
+                f"recomputing eta/ome orientation maps"
+            )
             res = generate_eta_ome_maps(cfg, hkls=hkls)
-    else:
-        logger.info('clean option specified; '
-                    + 'recomputing eta/ome orientation maps')
-        res = generate_eta_ome_maps(cfg, hkls=hkls)
-    filter_maps_if_requested(res, cfg)
+
+            filter_maps_if_requested(res, cfg)
```
psavery, line 432:
```suggestion
    filter_maps_if_requested(res, cfg)
```

It looks like this filtering occurred outside of the `if` block, right?

hexrd/findorientations.py
```diff
@@ -388,56 +388,48 @@ def load_eta_ome_maps(cfg, pd, image_series, hkls=None, clean=False):
...
-    if fn.split('.')[-1] != 'npz':
-        fn = fn + '.npz'
-
-    if not clean:
         try:
-            res = EtaOmeMaps(fn)
+            res = EtaOmeMaps(str(fn))
             pd = res.planeData
-            logger.info('loaded eta/ome orientation maps from %s', fn)
+            logger.info(f'loaded eta/ome orientation maps from {str(fn)}')
```
psavery, line 418:
```suggestion
            logger.info(f'loaded eta/ome orientation maps from {fn}')
```

It should be converted to a string automatically here.

hexrd/cli/fit_grains.py
```diff
@@ -195,25 +183,36 @@ def execute(args, parser):
                 raise(RuntimeError,
                       "error loading indexing results '%s'" % quats_f)
         else:
+            quats_f = cfg.find_orientations.accepted_orientations_file
```
psavery, line 186:
```suggestion
```

This was already defined above, right?

hexrd/config/root.py
```diff
@@ -16,8 +17,42 @@
...
+        This directory is certain to exist. If the specified directory does
+        not exist, it defaults to the current working directory.
+        """
+        try:
+            temp = Path(self.get('working_dir'))
+            if not temp.exists():
+                raise IOError(f'"working_dir": {temp} does not exist')
+            return temp
+
+        except RuntimeError:
```
psavery, line 33: What causes the `RuntimeError` to occur? Looks like the above `IOError` is not a `RuntimeError`.

Oh, maybe it occurs if `self.get('working_dir')` returns `None`? If that's the case, it might be clearer to just check for that in an `if/else` statement rather than catch a `RuntimeError`.

hexrd/config/root.py
```diff
@@ -26,7 +61,32 @@ def analysis_name(self, val):
 
     @property
     def analysis_dir(self):
-        return os.path.join(self.working_dir, self.analysis_name)
+        """Analysis directory, where output files go
+
+        The name is derived from `working_dir` and `analysis_name`. This
+        propetry returns a Path object. The directory and any intermediate
```
psavery, line 67:
```suggestion
        property returns a Path object. The directory and any intermediate
```

psavery (commented):
Overall, I think this is a huge improvement! It is much better to contain the logic for files/directories within the config, rather than within the `find-orientations` and `fit-grains` functions. I left some comments - they are only minor things.

Can you make sure that the tests pass? It looks like some of the tests are failing, because apparently comparing `Path` objects with strings always fails (for example, `cfg.working_dir == '/some/path'` always fails now, because `Path('/some/path') != '/some/path'`).

I think you just need to get the tests to use a `Path` object now in the comparison, or convert the `Path` object to `str`, whichever you prefer.

There are apparently some packaging issues too that I will look over.

hexrd/config/findorientations.py
```diff
@@ -23,6 +24,44 @@
 
 class FindOrientationsConfig(Config):
 
+    _find_ori = "find-orientations"
+
+    def _active_material_str(self):
+        return self.parent.material.active.strip().replace(' ', '-')
```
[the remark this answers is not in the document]
donald-e-boyce: My thinking here was that a property is part of the external API for the module. Since I only plan to use this internally, then making it property might be confusing.

hexrd/findorientations.py
```diff
@@ -388,56 +388,48 @@ def load_eta_ome_maps(cfg, pd, image_series, hkls=None, clean=False):
...
-    clean : TYPE, optional
-        DESCRIPTION. The default is False.
+    cfg: Config instance
+        root config file for this problem
+    pd: PlaneData instance
+        crystallographic properties
+    image_series: ImageSeries instance
+        stack of images
+    hkls: list, default = None
+        list of HKLs used in the eta-omega maps.
```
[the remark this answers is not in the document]
donald-e-boyce: I'll update these docs. First, it will try to load the file if they have already been computed. If not, it will use the HKJs in the config file unless the HKLs are specifically passed here; then the passed HKLs override.

hexrd/cli/find_orientations.py
```diff
@@ -50,39 +50,20 @@ def configure_parser(sub_parsers):
     p.set_defaults(func=execute)
 
 
-def write_scored_orientations(results, cfg):
```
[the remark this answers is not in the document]
donald-e-boyce: I agree it makes more sense to put it somewhere else, maybe in hexrd/findorientations.py.  I can do that if you don't mind updating the GUI.  Or I could put it there and keep a reference here, and you could change it at your leisure.

hexrd/config/findorientations.py
```diff
@@ -231,12 +270,47 @@ def eta_step(self):
...
+            mapf = root.analysis_dir / dflt
+        else:
+            fname = '_'.join([root.analysis_id, "eta-ome_maps.npz"])
+            mapf = root.working_dir / fname
+
+        # Now check the YAML.
+        temp = self._cfg.get(
+            'find_orientations:orientation_maps:file',
+            default=None
+        )
```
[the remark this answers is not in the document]
donald-e-boyce: Actually not. If it is not `None`, it can still be a relative path, and so the final result still depends on the `new_file_placement` flag.

hexrd/config/findorientations.py
donald-e-boyce: Yes, the `analysis_dir` and the `working_dir` are both guaranteed to be a Path, regardless of whether new or old file placement is being used.

hexrd/config/root.py
donald-e-boyce: Got it.

hexrd/findorientations.py
donald-e-boyce: Got it.  Had to see for myself !

hexrd/findorientations.py
donald-e-boyce: Good catch.

hexrd/config/findorientations.py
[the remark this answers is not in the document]
psavery: Okay that sounds fine!

hexrd/findorientations.py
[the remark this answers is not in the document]
psavery: Okay, thanks Don!

hexrd/config/findorientations.py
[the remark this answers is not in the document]
psavery: It looks like if `temp` is not `None`, then `mapf` is overridden below (and it doesn't rely on any of the code above). But this isn't a big deal...

hexrd/cli/find_orientations.py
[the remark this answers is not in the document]
psavery: That sounds great! Thanks Don. Yeah and the import you left is indeed leaving a reference to it! I'll update the GUI after this PR is merged.

hexrd/config/root.py
psavery: Thanks Don! That makes it easier to read 🙂 . It just wasn't clear before what was causing the `RuntimeError` that was being caught.

hexrd/findorientations.py
psavery: Yeah! When you put a variable within the f-string braces `{}`, it will implicitly convert to a string (to be combined with the rest of the string).

psavery:
Thanks Don! The changes look good so far! Now we just need to fix the tests (as mentioned [here](https://github.com/HEXRD/hexrd/pull/733#pullrequestreview-2486651655)).

psavery (approved):
Looks good to me! I fixed the packaging issues in master (so they won't be an issue after merging).

PR #735, opened 2024-12-11
Add ability to specify center of rotation

psavery:
For relative constraints, when tilts are applied to the detectors, the center of rotation is an additional setting that can be varied. Specifying a center of rotation that better matches what is expected in the experiment will result in calibrations with greater accuracy and speed.

For example, for FIDDLE, if there is any tilt whatsoever, then the detectors must have rotated about the lab origin (0, 0, 0). Requiring the detectors to be rotated about the origin ensures that the calibration will match what is expected from the experiment.

For other detector setups, such as the Eiger, the mean center of the instrument better matches what is expected if a tilt must be applied.


saransh13 (approved):
LGTM

PR #737, opened 2024-12-20
Add beam vector to simulateGVecs arguments

psavery:
It was assumed to be the default beam vector before, but we should allow it to be specified.


psavery:
We need to also add the beam vector as an argument to `HEDMInstrument.simulate_rotation_series()` and `Detector.simulate_rotation_series()` as well.

bnmajor (approved):
LGTM!

PR #738, opened 2024-12-20
Only pass one rMat_s to angularPixelSize

psavery:
We were previously passing an array of matrices, and only the first matrix was being used in the old transforms code. In the new transforms code, an error is being raised because it expects only a single matrix (and not an array of matrices). We should explicitly only pass this first matrix to `angularPixelSize()`.


psavery:
This PR maintains the previous behavior when we were using the old transforms.

However, I think it's possible that the `rMat_s` for the angular pixel size doesn't actually matter, and we can just set it to identity.

bnmajor (approved):

PR #739, opened 2024-12-30
simple fixes 

donald-e-boyce:
This fixes some problems introduced in pull request #733. Fixes include:

1. default eta-omega maps filename
2. fixes argument name in `write_scored_orientations` function
3. cleans up the `working_dir` property in config.root


psavery (approved):
The changes look good to me! We'll verify the CI passes next

PR #744, opened 2025-01-10
Eigen transforms

ZackAttack614:


hexrd/transforms/cpp_sublibrary/src/transforms.cpp
```diff
@@ -0,0 +1,401 @@
...
+  auto mat02 = rMat_c(0, 2);
+  auto mat10 = rMat_c(1, 0);
+  auto mat11 = rMat_c(1, 1);
+  auto mat12 = rMat_c(1, 2);
+  auto mat20 = rMat_c(2, 0);
+  auto mat21 = rMat_c(2, 1);
+  auto mat22 = rMat_c(2, 2);
+
+  // SIMD loop for full batches
+  for (size_t i = 0; i < num_full_batches * batch_size; i += batch_size) {
```
kpwelsh, line 92: Can we use xsimd::transform here to avoid writing this loop body twice?

hexrd/transforms/cpp_sublibrary/src/transforms.cpp
kpwelsh, line 1: Theres some odd indentation all around this file. 
Since we don't have a lot of cpp in this repo, I think its a good idea to start using clang-format here.

tests/transforms/test_make_binary_rmat.py
```diff
@@ -0,0 +1,25 @@
...
+
+def test_make_binary_rmat():
+    np.random.seed(0)
+    for _ in range(100):
+        axis = random_unit_vectors()
+        rmat = make_binary_rmat(axis)
+
+        # Two binary rmats should be the identity
+        assert np.allclose(rmat @ rmat, np.eye(3))
+
```
kpwelsh, line 21:
```suggestion
        assert np.allclose(rmat.T @ rmat, np.eye(3)), "It is orthogonal"
        assert np.all((np.abs(rmat) - 1 < 1e-10) | (np.abs(rmat) < 1e-10)), "It is binary"
```


These seem like reasonable tests to have as well,

kpwelsh (changes requested):
A couple of minor things should be changed.
A couple of minor things could be changed.

Not 100% sure if there is a  better way to do the xsimd or not.

hexrd/transforms/cpp_sublibrary/src/transforms.cpp
```diff
@@ -0,0 +1,401 @@
...
+}
+
+PYBIND11_MODULE(transforms, m)
+{
+  m.doc() = "HEXRD transforms plugin";
+
+  m.def("make_binary_rot_mat", &makeBinaryRotMat, "Function that computes a rotation matrix from a binary vector");
+  m.def("make_rot_mat_of_exp_map", &makeRotMatOfExpMap, "Function that computes a rotation matrix from an exponential map");
+  m.def("makeOscillRotMat", &makeOscillRotMat, "Function that generates a collection of rotation matrices from two angles (chi, ome)");
+  m.def("anglesToGvecToDetectorXYFromAngles", &anglesToGvecToDetectorXYFromAngles, "I hate this.");
```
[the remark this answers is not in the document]
psavery: This function could just be called `anglesToDetectorXY` instead.

setup.py
```diff
@@ -113,7 +114,20 @@ def get_pybind11_include_path():
...
+        str(cpp_transform_pkgdir / 'src/transforms.cpp'),
+        str(cpp_transform_pkgdir / 'src/inverse_distortion.cpp'),
+    ]
+
+    extra_compile_args = [
+        '-O3',
+        '-Wall',
+        '-shared',
+        '-std=c++14',
+        '-funroll-loops',
```
psavery, line 127: I think Windows is just going to straight up ignore these. Maybe we should write alternative flags to use for Windows, so it will be optimized as well? Like we did above on line 47. We do have Windows users.

psavery:
I ran a quick performance check on running `hexrd fit-grains ./mruby_config_composite.yml`, and after these changes, it took 139 seconds instead of 141.5 seconds before. It's about the same speed - which is good!

hexrd/transforms/cpp_sublibrary/src/transforms.cpp
kpwelsh: Lets not bother with this. Apparently xsimd::transform was removed recently and is currently in limbo.
https://github.com/xtensor-stack/xsimd/issues/1045

hexrd/transforms/cpp_sublibrary/src/transforms.cpp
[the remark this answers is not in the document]
kpwelsh:
```suggestion
  m.def("anglesToDetectorXY", &anglesToDetectorXY, "Converts angles to detector xy coordinates.");
```

Seems fine to me.

tests/transforms/test_gvec_to_xy_from_file.py
```diff
@@ -15,6 +15,7 @@ def test_gvec_to_xy_from_file(test_data_dir):
         allow_pickle=True
     )
     for obj in arr:
+        print(1)
```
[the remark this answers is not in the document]
kpwelsh:
```suggestion
```

setup.py
ZackAttack614: This is a good point, I'll come back to this in a future PR

kpwelsh (approved):

ZackAttack614:
@psavery Just making sure I have your approval to merge this in before clicking the button

psavery (approved):
Looks good to me :slightly_smiling_face:

PR #745, opened 2025-01-10
Take into account radius for calibration params

psavery:
For cylindrical detectors, we need to take into account the radius for calibration flags and parameters.

This uses the old, non-lmfit style for defining calibration flags and parameters. We will be able to eliminate all of this when we upgrade the HEDM calibration to use the new lmfit calibration style...

But for now, we need to maintain this.


saransh13 (approved):
LGTM

PR #746, opened 2025-01-11
Add instrument templates for dynamic compression

psavery:


saransh13 (changes requested):
Two changes:
1. Can you remove `.yml` from the instrument names
2. Capitalize first letters of Rigaku and Varex.
3. Capitalize DCS.
I'll approve after this.

psavery:
@saransh13 and I discussed this separately, and agreed that the current filenames are fine: we just need to make sure the names look fine in the GUI (which is done separately).

PR #757, opened 2025-02-11
Grain data update

donald-e-boyce:
This adds some new features to the GrainData class, including saving/reading grains.out files, quaternions attribute, strain attribute and a `select` method.  This also adds user documentation in the docs section.

### DIFF
--- a/hexrd/cli/fit_grains.py
+++ b/hexrd/cli/fit_grains.py
@@ -75,13 +75,15 @@ class GrainData(_BaseGrainData):
 
     def write_grains_out(self, fname):
         """Write a file in grains.out format"""
-        gw = GrainDataWriter(filename=fname)
+        gw = instrument.GrainDataWriter(filename=fname)
         n = len(self.id)
         for i in range(n):
             gparams = np.hstack(
                 (self.expmap[i], self.centroid[i], self.inv_Vs[i])
             )
-            gw.dump_grain(self.id, self.completeness, self.chisq, gparams)
+            gw.dump_grain(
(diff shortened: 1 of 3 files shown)


hexrd/cli/fit_grains.py
```diff
@@ -54,6 +55,11 @@ def load(cls, fname):
         """
         return cls(**np.load(fname))
 
+    @classmethod
+    def from_grains_out(cls, fname):
+        """Read hexrd grains output file"""
+        return cls.from_array(np.loadtxt(fname))
```
psavery, line 61: Nice, Don! This class is looking like it is becoming very useful :slightly_smiling_face:

tests/test_graindata.py
```diff
@@ -7,41 +7,86 @@
...
-exp90 = (np.pi/2) * np.identity(3)
-rmats90 = np.array([
-    [[1, 0, 0], [0, 0,- 1], [0, 1, 0]],
-    [[0, 0, 1], [0, 1, 0], [-1, 0, 0]],
-    [[0, -1, 0], [1, 0, 0], [0, 0, 1]]
-])
+@pytest.fixture
+def quats90():
+    c45, s45 = np.cos(np.pi / 4), np.sin(np.pi /4)
+    return [[c45, s45, 0, 0], [c45, 0, s45, 0], [c45, 0, 0, s45]]
```
psavery, line 18: Constants could also just be global variables like `QUATS_90`. pytest fixtures often use other things like environment variables or data on disk to make them. But it doesn't hurt to do it this way, and you can keep it.

PR #762, opened 2025-02-12
Grain data update

donald-e-boyce:
This adds some new features to the GrainData class, including saving/reading grains.out files, quaternions attribute, strain attribute and a select method. This also adds user documentation in the docs section.


psavery (approved):
LGTM

PR #748, opened 2025-01-20
Reorganization

kpwelsh:
# Reorganization Plan
## 1 - File Movement
The reorganization consists primarily of creating sub-folders under `hexrd` and moving most files into one of them. The goal of the first PR will be to move the files into this new location while making the minimum changes necessary to maintain functionality. The reorganization plan suggests that some files / classes be split up across workflows. During this phase, we will instead *duplicate* the files across workflows.

## 2 - Reconciliation
During this phase, we will consider in smaller chunks code changes necessary to split up the classes and files across workflows. This should be limited to approximately the following code areas:
* hexrd/materials/
* * crystallography
* * unitcell
* hexrd/instrument/
* * hedm_instrument
* * physics_package
* * detector
* hexrd/xrdutil
* hexrd/config
* * instrument

## 3 - Migration Tooling
During this phase, we will consider what the best options are for making code migration for users as painless as possible. 

----

# Phase 1 File Migration

All of the mapping was done according to the workflow labeling spreadsheet. A complete list of the file mappings can be found in `file_map.pkl` . It is being kept temporarily to support migration workflows.

## File Mapping

The files were mapped automatically using a new `package_remapper` tool. This works in a couple of stages to correctly move the files while keeping imports coherent. This tool was used to accomplish the bulk of the work and help ensure that no imports are missed, but all code changes made by the tool were reviewed and verified by hand.

### File Mappings

The package_remapper requires a 1-to-many map of files from one repository to the other. It will copy the files in the `source` to everywhere listed in the `destination` . The file mapping used was generated from the spreadsheet by rolling up the workflows to the file level. Things like `hexrd/valunits.py` clearly map to one workflow, `core` . However, things like `hexrd/config/instruments.py` need to be split into multiple workflow sub-folders (`core` and `hedm` in this case. During the file mapping stage, we will just copy the whole file in both locations and log it for a human (me) to disentangle later. It is far too complicated to safely do this disentanglement automatically.

Example:
```
hexrd/config/instrument.py -> hexrd/core/config/instrument.py
                           -> hexrd/hedm/config/instrument.py

```

It is not necessary that all files in the source directory have an explicit destination. If files are omitted for some reason, they will be brought over with the containing module. If there is no containing module (e.g. `setup.py`) they will be copied over in the same location relative to the package source. When generating the list of files, the relevant `.gitignore` is used to filter out files we don’t care about if any are present.

### Quick Git Commit

[...]

PR #763, opened 2025-02-12
Reorganization rebase

kpwelsh:
This is an unfortunate continuation of #748 . 

Had to make a new branch to get the commit history looking OK.


kpwelsh:
This is no longer necessary. Figured out how to overwrite the old reorganization branch

PR #767, opened 2025-02-15
mask out diffracted beams which travel parallel to physics package surface or which are backscattered.

saransh13:
Some configurations have points on the detector which can intercept backscattered x-rays. These points should be masked out since no signal will be detected due to self-absorption of the physics package.

### DIFF
--- a/hexrd/instrument/detector.py
+++ b/hexrd/instrument/detector.py
@@ -1720,7 +1720,14 @@ class Detector:
 
         dvecs = angles_to_dvec(angs, beam_vec=bvec)
 
-        secb = np.abs(1./np.dot(dvecs, sample_normal).reshape(self.shape))
+        secb = 1./np.dot(dvecs, sample_normal)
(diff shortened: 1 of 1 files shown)


hexrd/instrument/detector.py
```diff
@@ -1720,7 +1720,17 @@ def calc_physics_package_transmission(self, energy: np.floating,
...
-        secb = np.abs(1./np.dot(dvecs, sample_normal).reshape(self.shape))
+        cosb = np.dot(dvecs, sample_normal)
+        '''angles for which secb <= 0 or close are diffracted beams
+        almost parallel to the sample surface or backscattered, we
+        can mask out these values by setting secb to nan
+        '''
+        mask = np.logical_or(cosb < 0, 
+                             np.isclose(
+                            cosb, 0.,
+                            atol=5E-2))
```
psavery, line 1731: This could alternatively just be:

```python
mask = cosb <= 5e-2
```

psavery (approved):
LGTM

Issue #769, opened 2025-02-20
hexrd import fails with hexrd 3.12


donald-e-boyce:
In python 3.12, `SafeConfigParser` has been eliminated.  See [python 3.12 changes](https://docs.python.org/3/whatsnew/3.12.html#configparser).

```(fenicsx-0.8) (MacBook-Pro-2023: pkg) 1262. ipython 
Python 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:35:20) [Clang 16.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.25.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import hexrd
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[1], line 1
----> 1 import hexrd

File ~/Source/hexrd/pkg/hexrd/__init__.py:4
      1 import importlib
      2 import sys
----> 4 from .material import crystallography
      5 from .material import jcpds
      6 from .material import mksupport

File ~/Source/hexrd/pkg/hexrd/material/__init__.py:1
----> 1 from .material import (
      2     _angstroms,
      3     _kev,
      4     load_materials_hdf5,
      5     Material,
      6     save_materials_hdf5,
      7 )

File ~/Source/hexrd/pkg/hexrd/material/material.py:34
      1 # -*- coding: utf-8 -*-
      2 # =============================================================================
      3 # Copyright (c) 2012, Lawrence Livermore National Security, LLC.
   (...)
     26 # Boston, MA 02111-1307 USA or visit <http://www.gnu.org/licenses/>.
     27 # =============================================================================
     28 """
     29 Module for XRD material class
     30 
     31 Use the Material class directly for new materials.  Known
     32 materials are defined by name in materialDict.
     33 """
---> 34 from configparser import SafeConfigParser as Parser
     35 import numpy as np
     37 from hexrd.material.crystallography import PlaneData as PData

ImportError: cannot import name 'SafeConfigParser' from 'configparser' (/Users/deboyce/miniconda3/envs/fenicsx-0.8/lib/python3.12/configparser.py)

In [2]: 
```

PR #771, opened 2025-02-21
Use ConfigParser instead of deprecated SafeConfigParser

ZackAttack614:
Resolves issue #769. `SafeConfigParser` is removed in Python 3.12.

See source code here: https://github.com/python/cpython/blob/v3.11.1/Lib/configparser.py#L1242-L1252

Fixes: #769

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     name: ${{ matrix.config.name }}
     runs-on: ${{ matrix.config.os }}
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


ZackAttack614:
The change to `ConfigParser` somehow negatively affects MacOS, but the tests pass in Linux and Windows. I'm not sure why this would be, will look into it. It's a bit cumbersome since I don't have access to a Mac.

psavery:
The error here was just caused by a new version of `PyCIFRW`, and is fixed in #772

psavery (approved):

psavery:
I just rebased this one on master. If the CI passes, we'll merge it.

PR #777, opened 2025-02-27
Parallelize rows in snip1d

psavery:
Snip1d was taking a very long time to compute for high resolution polar images. Since it is computed on a row-by-row basis, however, I found that parallelizing over the rows using multiprocessing provided a big performance improvement (3x faster for my example).

Some other things I tried that did not result in a speed improvement:

* Using the latest `convolve` from astropy
* Using the latest `convolve_fft` from astropy
* Performing multithreading instead of multiprocessing
* Parallelizing at a finer grain level in the snip1d loops

This also includes a test that verifies that the snip1d output does not change.


saransh13 (approved):
We should add same parallelism to the FastSNIP1D kernel in a future PR.

PR #778, opened 2025-03-05
Remove an unnecessary copy in ProcessedImageSeries

psavery:
When a frame is requested, all imageseries adapters appear to be designed to return a new copy of the frame. Additionally, all `ProcessedImageSeries` operations appear to return a new copy of the frame.

This means that the copy in `ProcessedImageSeries._process_frame` is unnecessary.

Copying the frames can prove to be very computationally expensive as well. Profiling `pull_spots()` on some Eiger data that used a `ProcessedImageSeries` as input showed that 99% of the time was spent making this copy. Thus, removing this unnecessary copy brought the time down in my example from ~70 minutes to ~51 seconds (an 82x speed increase).

This will definitely impact the GUI, since we often use `ProcessedImageSeries` there. It will only affect `pull_spots()` calls in scripts, however, when a `ProcessedImageSeries` is used.


donald-e-boyce (approved):
This looks great!

PR #781, opened 2025-03-11
Only write spots files if indicated to do so

psavery:
In the `fit_grains()` function, the parameter `write_spots_files` is used to determine whether to write spots files or not.

Changes in 62c07b5 and 5a88225 appear to have accidentally caused it to be ignored, however. Switch back to the previous variable, which is correctly updated according to whether the spots files should be written or not.


donald-e-boyce (approved):
Looks good. Thanks for catching this.

PR #783, opened 2025-03-13
 Improve rectangle operation performance

ChristosT:
# Overview

Provide optimized implementation for when applying the rectangle operator in an imageseries as its first operation. 

ImageSeriesAdapter class has a new method
```
def get_region(self, frame_idx: int, region: RegionType) -> np.ndarray:
    r = region
    return self[frame_idx][r[0][0]:r[0][1], r[1][0]:r[1][1]]
```
Derived classes can offer optimized implementations. 
only ```FrameCacheImageSeriesAdapter```  and ``HDF5ImageSeriesAdapter`` take advantage of this for now.

# Affected Workflows

workflows utilizing `ProcessedImageSeries` should see a  speedup if they use the rectangle operation as their first operation.
Results running locally on a 1441 frame dataset
| format            | before (s) | after(s) | % improvement |
|-------------------|------------|----------|---------------|
| framecache (npz)  | 1020       | 59       | 94%           |
| framecache (fch5) | 935        | 40.3     | 95%           | 
| raw hdf5                | 6020       | 6070     | -0.8 %        | 

# Documentation Changes
no documentation changes

### DIFF
--- a/hexrd/imageseries/process.py
+++ b/hexrd/imageseries/process.py
@@ -8,7 +8,7 @@ from .baseclass import ImageSeries
 
 
 class ProcessedImageSeries(ImageSeries):
-    """imsageseries based on existing one with image processing options
+    """imageseries based on existing one with image processing options
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


hexrd/imageseries/process.py
```diff
@@ -56,9 +56,23 @@ def __iter__(self):
...
+       # note: key refers to original imageseries
+        oplist = self.oplist
+
+        # when rectangle is the first operatation we can try to call the
+        # optimized version.  If the adaptor provides one it should be
+        # signifcantly faster if not it will fallback to the same
+        # implementation that _rectangle provides.
+        if oplist and  oplist[0][0] == self.RECT:
+            region = oplist[0][1]
+            img = self._rectangle_optimized(key,region)
```
psavery, line 68:
```suggestion
        # when rectangle is the first operation we can try to call the
        # optimized version.  If the adapter provides one it should be
        # signifcantly faster if not it will fallback to the same
        # implementation that _rectangle provides.
        if oplist and  oplist[0][0] == self.RECT:
            region = oplist[0][1]
            img = self._rectangle_optimized(key, region)
```

hexrd/imageseries/process.py
```diff
@@ -81,6 +95,9 @@ def _subtract_dark(self, img, dark):
         ret[ret < 0] = 0
         return ret
 
+    def _rectangle_optimized(self, img_key, r):
+        return self._imser._adapter.get_region(img_key,r)
```
psavery, line 99:
```suggestion
        return self._imser._adapter.get_region(img_key, r)
```

psavery (approved):
This is a great improvement, thanks Christos!

PR #785, opened 2025-03-24
Unpin pycifrw

psavery:
It seems that the [issue](https://github.com/jamesrhester/pycifrw/issues/14) that caused us to pin it has been fixed.


bnmajor (approved):
Awesome!

PR #787, opened 2025-03-27
Handle recursion in imageseries get_region()

psavery:
Sometimes (as is the case for `OmegaImageSeries`), `self._adapter` is itself an imageseries, rather than an imageseries adapter.

To handle this recursion, use a forwarding function on the `ImageSeries` base class to forward the call to its adapter (which, if the adapter is itself an imageseries, will then forward the call to its own adapter).

This fixes running [this example](https://github.com/HEXRD/examples/blob/master/NIST_ruby/multiruby_dexelas/include/mruby_config_composite_roi.yml).

### DIFF
--- a/hexrd/imageseries/baseclass.py
+++ b/hexrd/imageseries/baseclass.py
@@ -1,6 +1,9 @@
 """Base class for imageseries
 """
-from .imageseriesabc import ImageSeriesABC
+import numpy as np
+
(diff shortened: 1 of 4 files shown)


ChristosT (approved):
Looks great! By the way how come we didn't catch that in the CI ? Is this example skipped from the tests ?

psavery:
Yeah, not every example is ran in the tests. Maybe we should add that one.

PR #788, opened 2025-04-01
Modified pinhole area correction

saransh13:
# Overview

An error was discovered in the pinhole area correction based on the CeO2 data recorded on TARDIS. An extra `sec(beta)` factor was added to the correction published in Rygg et al., Rev. Sci. Instrum. 91, 043902 (2020). This PR makes that change. 

Another small modification was made to use the existing functionality in the `transforms` module to compute the `beta` angles. This streamlines the function.


psavery (approved):
LGTM!

PR #791, opened 2025-04-10
Remove instrument detection from phutil

psavery:
We originally had instrument-specific behavior for some angles, but that was fixed, and we no longer need that.

We are in the process of adding the FIDDLE instrument, which ought to be the same.


psavery:
This is being superseded by #792. We unfortunately have to keep the instrument detection, because there is not enough information in the instrument config to determine the shifts. These shifts depend on the relationship of Lab X, Y, Z to the assumed axes in the pinhole code.

PR #792, opened 2025-04-11
Add required x-axis and eta shifts for Fiddle instrument

psavery:
These are necessary for correct pinhole corrections for FIDDLE.


saransh13 (approved):
LGTM.

PR #793, opened 2025-04-16
Fix some bugs with the detector class

saransh13:
Two bug fixes in this PR:

1. fix missing bvec argument from cylindrical detectors, same as the planar detectors.
2. fix runtime warning for pinhole effective area correction due to `sqrt` of a negative number.

### DIFF
--- a/hexrd/instrument/cylindrical_detector.py
+++ b/hexrd/instrument/cylindrical_detector.py
@@ -210,11 +210,25 @@ class CylindricalDetector(Detector):
             'cols': self.cols,
         }
 
-    def pixel_tth_gradient(self, origin=ct.zeros_3):
-        return _pixel_tth_gradient(origin=origin, **self._pixel_angle_kwargs)
+    def pixel_tth_gradient(self, origin=ct.zeros_3, bvec: np.ndarray | None = None):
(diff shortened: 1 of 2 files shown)


hexrd/instrument/cylindrical_detector.py
```diff
@@ -117,7 +117,9 @@ def cart_to_dvecs(self,
                                                tVec_c=tvec_c,
                                                normalize=False)
 
-    def pixel_angles(self, origin=ct.zeros_3):
+    def pixel_angles(self, origin=ct.zeros_3, bvec: np.ndarray | None = None):
+        if bvec is None:
+            bvec = self.bvec
```
psavery, line 122: We need to pass this `bvec` to the kwargs for `_pixel_angles` somehow. I'd suggest removing the bvec from the `self._pixel_angle_kwargs` and calling `_pixel_angles` like this:

```python
        return _pixel_angles(origin=origin, bvec=bvec, **self._pixel_angle_kwargs) 
```

psavery: Done.

psavery (approved):

PR #794, opened 2025-04-29
WPPF fixes

saransh13:
WPPF module had a few edge case issues that showed up in the Rietveld and LeBail methods. This was leading to errors when the spline background method was being used. This PR fixes those issues.

There was an additional bug when more reflections were selected in the material than allowed by the X-ray wavelength or the detector coverage. This issue has also been fixed.

### DIFF
--- a/hexrd/wppf/WPPF.py
+++ b/hexrd/wppf/WPPF.py
@@ -209,7 +209,6 @@ class LeBail:
                 self._background.append(Spectrum(x=tth, y=np.zeros(tth.shape)))
 
         elif "spline" in self.bkgmethod.keys():
-            self._background = []
 
             points = self.bkgmethod["spline"]
             if not points:
@@ -217,11 +216,7 @@ class LeBail:
                     "Background points must be set to use spline"
                 )
 
-            for i, pts in enumerate(np.asarray(points)):
-                tth = self._spectrum_expt[i]._x
-                x = pts[:, 0]
-                y = pts[:, 1]
-                self._background.append(self.splinefit(x, y, tth))
+            self.splinefit(np.squeeze(np.array(points)))
 
         elif "chebyshev" in self.bkgmethod.keys():
             self.chebyshevfit()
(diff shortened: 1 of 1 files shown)


hexrd/wppf/WPPF.py
```diff
@@ -307,15 +302,19 @@ def chebyshevfit(self):
...
+        self._background = []
+        x = points[:,0]
+        y = points[:,1]
         cs = CubicSpline(x, y)
-        bkg = cs(tth)
-        return Spectrum(x=tth, y=bkg)
+        for i, s in enumerate(self._spectrum_expt):
+            tth = self._tth_list[i]
+            bkg = cs(tth)
+            self._background.append(Spectrum(x=tth, y=cs(tth)))
```
psavery, line 317:
```suggestion
            self._background.append(Spectrum(x=tth, y=bkg))
```

Can it be this instead?

hexrd/wppf/WPPF.py
```diff
@@ -1831,14 +1825,19 @@ def chebyshevfit(self):
...
         03/08/2021 SS adding tth as input. this is the
         list of points for which background is estimated
         """
+        self._background = []
+        x = points[:,0]
+        y = points[:,1]
         cs = CubicSpline(x, y)
-        bkg = cs(tth)
-        return Spectrum(x=tth, y=bkg)
+        for i, s in enumerate(self._spectrum_expt):
```
psavery, line 1837: `s` is not used. Should it be used?

hexrd/wppf/WPPF.py
```diff
@@ -1831,14 +1825,19 @@ def chebyshevfit(self):
...
+            tth = self._tth_list[i]
+            bkg = cs(tth)
+            self._background.append(Spectrum(x=tth, y=cs(tth)))
```
psavery, line 1840:
```suggestion
            self._background.append(Spectrum(x=tth, y=bkg))
```

hexrd/wppf/WPPF.py
```diff
@@ -1831,14 +1825,19 @@ def chebyshevfit(self):
...
         list of points for which background is estimated
         """
+        self._background = []
+        x = points[:,0]
+        y = points[:,1]
         cs = CubicSpline(x, y)
-        bkg = cs(tth)
-        return Spectrum(x=tth, y=bkg)
+        for i, s in enumerate(self._spectrum_expt):
+            tth = self._tth_list[i]
```
psavery, line 1838: It was `tth = self._spectrum_expt[i]._x` before. Is `self._tth_list[i]` the same, or better?

hexrd/wppf/WPPF.py
```diff
@@ -307,15 +302,19 @@ def chebyshevfit(self):
...
```
psavery, line 315: `self._tth_list[i]` here is more preferable to `self._spectrum_expt[i]._x`, right?

hexrd/wppf/WPPF.py
saransh13: Line 1062 is 
`self._tth_list = [s._x for s in self._spectrum_expt]`

So they should be the same.

hexrd/wppf/WPPF.py
psavery: Okay, the reason I asked is if we don't want to use it, it might be better to just change it to `for i in range(len(self._spectrum_expt))` rather than `for i, s in enumerate(self._spectrum_expt)`

psavery (approved):

PR #798, opened 2025-05-06
Update nfutil.py

darrencpagan:
Small change to gen_trial_exp_data function which fixes a bug that caused the max_thh variable to be ignored when the detector is placed far away from the sample.

# Overview

[Summarize the changes that you made. Make sure to call out changes or decisions that were not included in the original discussion. This should include:

-   APIs you are changing (i.e. function signatures)
-   Who might need to know about the change
-   Which test cases you will need to add / change
]

# Affected Workflows

[Describe who is affected by this and how.]

# Documentation Changes

[Enumerate the places where documentation needs to change. If the documentation lives in this repo, it should be changed in this PR.]

### DIFF
--- a/hexrd/grainmap/nfutil.py
+++ b/hexrd/grainmap/nfutil.py
@@ -1262,7 +1262,13 @@ def gen_trial_exp_data(grain_out_file,det_file,mat_file, mat_name, max_tth, comp
     # crystallography data
     beam_energy = valunits.valWUnit("beam_energy", "energy", instr.beam_energy, "keV")
     beam_wavelength = constants.keVToAngstrom(beam_energy.getVal('keV'))
-    dmin = valunits.valWUnit("dmin", "length",
+    if max_tth is not None:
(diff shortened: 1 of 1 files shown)


hexrd/grainmap/nfutil.py
```diff
@@ -1262,7 +1262,13 @@ def gen_trial_exp_data(grain_out_file,det_file,mat_file, mat_name, max_tth, comp
     # crystallography data
     beam_energy = valunits.valWUnit("beam_energy", "energy", instr.beam_energy, "keV")
     beam_wavelength = constants.keVToAngstrom(beam_energy.getVal('keV'))
-    dmin = valunits.valWUnit("dmin", "length",
+    if max_tth is not None:
+        print('ping')
```
psavery, line 1266: Maybe you should remove this debug statement? Or you can leave it if you prefer. It's up to you.

psavery (approved):

PR #800, opened 2025-05-09
Add chemical formula to layer properties

psavery:
For both detector coating layers and physics package layers, allow a chemical formula to be provided.

This fixes an issue where you previously always had to provide the chemical formula for the material name. Now, you can provide other names such as `diamond` and the chemical formula can be provided separately (for example, `C8`) to be used for absorption correction calculations.


psavery:
Thanks Saransh for https://github.com/HEXRD/hexrd/pull/796 !

saransh13 (approved):
LGTM

PR #804, opened 2025-05-29
Solid angle speedup

saransh13:
Simplify the solid angle calculation using the first principle definition. For a surface of area, $\Delta$ at a location, $\mathbf{r}$ from the diffracting volume, the solid angle, $\mathrm{d}\Omega$ is given by:

$\mathrm{d}\Omega = \left(\frac{\Delta}{\vert \mathbf{r}\vert^{3}}\right)\mathbf{r}\cdot\hat{\mathbf{n}}$. 

$\hat{\mathbf{n}}$ is the local normal of the surface. Using this definition instead of summing over the areas of the triangle improves the speed by 2 orders of magnitude.

### DIFF
--- a/hexrd/instrument/cylindrical_detector.py
+++ b/hexrd/instrument/cylindrical_detector.py
@@ -231,6 +231,12 @@ class CylindricalDetector(Detector):
         )
 
     @property
+    def pixel_normal(self):
+        if not hasattr(self, '_pixel_normal'):
+            self._pixel_normal = self.local_normal()
+        return self._pixel_normal
+
+    @property
     def caxis(self):
         # returns the cylinder axis
         return np.dot(self.rmat, ct.lab_y)
--- a/hexrd/instrument/cylindrical_detector.py
+++ b/hexrd/instrument/cylindrical_detector.py
@@ -7,7 +7,7 @@ from hexrd import xrdutil
 from hexrd.utils.decorators import memoize
 
 from .detector import (
-    Detector, _solid_angle_of_triangle, _row_edge_vec, _col_edge_vec
+    Detector, _row_edge_vec, _col_edge_vec
 )
 
 from functools import partial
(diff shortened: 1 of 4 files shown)


hexrd/instrument/cylindrical_detector.py
```diff
@@ -230,6 +230,12 @@ def pixel_eta_gradient(self, origin=ct.zeros_3, bvec: np.ndarray | None = None):
             **self._pixel_angle_kwargs,
         )
 
+    @property
+    def pixel_normal(self):
+        if not hasattr(self, '_pixel_normal'):
+            self._pixel_normal = self.local_normal()
+        return self._pixel_normal
```
psavery: Setting `self._pixel_normal` here is a form of memoization, but we have to make sure that `self._pixel_normal` is invalidated if any parameters used in `self.local_normal()` are modified. The easier memoization pattern to use is to make a global function like `_local_normal()` that takes all required parameters as arguments, and decorate it with `@memoize`. That does automatic invalidation when any arguments are changed/modified.

However, the calculation for `local_normal()` seems simple enough that it's probably better to not memoize it at all. Memoization does come with a cost: all input arrays must be hashed every time it is called. If a function is simple enough, the cost may exceed the benefit. I think that it may be better to just skip memoization entirely in these `pixel_normal` and `pixel_solid_angle` functions. I'm going to do that in a commit that I'm adding.

hexrd/instrument/cylindrical_detector.py
```diff
@@ -283,34 +289,6 @@ def angle_extent(self):
...
-        Detector.update_memoization_sizes(all_panels)
-
-        num_matches = sum(isinstance(x, CylindricalDetector) for x in all_panels)
-        funcs = [
-            _pixel_angles,
-            _pixel_tth_gradient,
-            _pixel_eta_gradient,
-            _pixel_solid_angles,
-        ]
-        Detector.increase_memoization_sizes(funcs, num_matches)
```
psavery, line 312: We still need to keep this function, and just remove `_pixel_solid_angles` from the list. I'm doing that in a commit I am adding.

psavery (commented):
This is an amazing improvement, and a lot of code cleanup! I simplified it a little bit to remove the memoization (as it seems unnecessary - see the other comments I left), and added back in the `update_memoization_sizes()` function, which is needed for the other memoized functions.

PR #810, opened 2025-05-31
Skip nested metadata when writing out NPZ files

psavery:
NPZ files do not support nested metadata. When we write them out, they produce an NPZ file that requires `allow_pickle=True` to open, and we do not want to perform that within hexrd or the GUI for security reasons.

While saving an NPZ file, if we encounter a nested metadata item, print a warning that it will be skipped, and skip over it.


psavery:
Merging as this is somewhat urgent...

PR #815, opened 2025-06-13
Liquid model wppf

saransh13:
# Overview

The WPPF module lacked a model to deal with cases where diffraction from both crystalline and amorphous diffraction are present. This PR addresses this deficiency by including 3 amorphous diffraction models:

1. An asymmetric split-gaussian peak shape
2. An asymmetric split-pseudo flight peak shape
3. a liquid profile which is experimentally measured

The first two peak shapes are implemented as `numba` accelerated function in `WPPF.peakshapes` module. The set of refinement parameters are different for the three models. Additional changes have been made to `WPPF.wppfsupport` module to deal with this. An example fit for Ti is shown below

Screenshot 2025-06-13 at 2 05 45 PM

# Affected Workflows

None

### DIFF
--- a/hexrd/wppf/WPPF.py
+++ b/hexrd/wppf/WPPF.py
@@ -120,9 +120,9 @@ class LeBail:
 
         self.phases = phases
 
-        self.params = params
-
         self.amorphous_model = amorphous_model
+
+        self.params = params
 
         self.initialize_Icalc()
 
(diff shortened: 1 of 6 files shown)


psavery (approved):
We'll definitely need to cleanup WPPF code at some point, but this works and there are tests included!

PR #816, opened 2025-06-16
Significantly improve fit-grains performance

psavery:
Profiling showed that the list comprehension in `objFuncFitGrain` was taking a significant amount of time. This is because that function is called repeatedly during the least squares fit, and repeatedly performing the list comprehension would take a substantial amount of time.

Performing the list comprehension outside of `objFuncFitGrain` was shown to provide a substantial speedup.

This is something I discovered when I was working on the spot tracking code. I'm simply cherry-picking that commit here.

This only affects the fitting part, and not the `pull_spots()` part.

### DIFF
--- a/hexrd/fitting/grains.py
+++ b/hexrd/fitting/grains.py
@@ -78,7 +78,33 @@ def fitGrain(gFull, instrument, reflections_dict,
 
     gFit = gFull[gFlag]
 
-    fitArgs = (gFull, gFlag, instrument, reflections_dict,
+    # objFuncFitGrain can run *significantly* faster if we convert the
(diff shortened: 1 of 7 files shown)


psavery:
The second commit on this PR adds support for indexing into an imageseries like this: `ome_imgser[i_frame, ijs[0], ijs[1]]`, instead of the previous method of `ome_imgser[i_frame][ijs[0], ijs[1]]`.

This means that instead of creating a whole dense array, and then performing fancy indexing into it, we can perform fancy indexing on the sparse array directly, and then create the array from that. This has the potential to speed up `pull_spots()` significantly. For one case I just studied recently, it sped up `pull_spots()` by around 5x.

Issue #811, opened 2025-06-06
problem no longer working


donald-e-boyce:
### Describe the bug

I was trying to rerun a problem that was running perfectly as of 12/28/2024, but I get this error now.  The error arises in hexrd/grains.py in the matchOmegas() function, and it has to do with `nan`s being returned by `xfcapi.oscill_angles_of_hkls`.   

I am using the current version as of today: commit bf9b1d86271c6496bc115a0aa171bc5f22bece75 (May 31). 

```
25-06-06 11:12:39 - 	starting fit on 8 processes with chunksize 58
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/Users/deboyce/miniconda3/envs/hexrd-py311/lib/python3.11/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/deboyce/miniconda3/envs/hexrd-py311/lib/python3.11/multiprocessing/pool.py", line 48, in mapstar
    return list(map(*args))
           ^^^^^^^^^^^^^^^^
  File "/Users/deboyce/Source/hexrd/pkg/hexrd/fitgrains.py", line 217, in fit_grain_FF_reduced
    grain_params = fitGrain(
                   ^^^^^^^^^
  File "/Users/deboyce/Source/hexrd/pkg/hexrd/fitting/grains.py", line 83, in fitGrain
    results = optimize.leastsq(objFuncFitGrain, gFit, args=fitArgs,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/deboyce/miniconda3/envs/hexrd-py311/lib/python3.11/site-packages/scipy/optimize/_minpack_py.py", line 438, in leastsq
    retval = _minpack._lmdif(func, x0, args, full_output, ftol, xtol,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/deboyce/Source/hexrd/pkg/hexrd/fitting/grains.py", line 238, in objFuncFitGrain
    match_omes, calc_omes = matchOmegas(
                            ^^^^^^^^^^^^
  File "/Users/deboyce/Source/hexrd/pkg/hexrd/fitting/grains.py", line 335, in matchOmegas
[...]

PR #817, opened 2025-06-16
Add start of omega period to measured omegas

psavery:
This fixes an issue introduced in #786.

We were missing the start of the omega period. This was not an issue if the start of the omega period was 0, but it was an issue otherwise, and likely caused some bad results.

Fixes: #811


psavery:
Merging, as this is urgent...

PR #820, opened 2025-06-22
Provide omega period to fitGrain

psavery:
This does change the results of fit-grains slightly. We are currently assessing the changes.

The chi-squared remained nearly the same. For one grain, it increased very slightly, and for another grain, it decreased very slightly.


psavery:
@donald-e-boyce This is what it would look like if we provided the omega period to `fitGrain()`.

Feel free to examine the differences in the fit-grains results from this branch compared to master. They are slightly different results.

psavery:
I personally tried this using [this example](https://github.com/HEXRD/examples/blob/master/NIST_ruby/multiruby_dexelas/include/mruby_config_composite.yml).

PR #827, opened 2025-08-08
Sanitize material names in WPPF

psavery:
The material names cannot contain '-' when being used as lmfit parameters, or else an error occurs.

Partially fixes: hexrd/hexrdgui#1891


saransh13 (approved):
LGTM

PR #828, opened 2025-08-08
Remove tth_distortion properties from calibrator

psavery:
It doesn't make sense for an instrument calibrator to have tth_distortion properties, because each modal calibrator inside might have a different `tth_distortion` dict.

To avoid confusion, just remove this property.


saransh13 (approved):
lgtm

PR #831, opened 2025-08-14
Remove WPPF Parameters class

psavery:
We can just use lmfit parameters for everything. We have to ultimately use lmfit anyways.


hexrd/wppf/WPPF.py
```diff
@@ -705,7 +693,12 @@ def Refine(self):
                      variables which are allowed to be varied.
         """
 
-        params = self.initialize_lmfit_parameters()
+        # FIXME: why is this necessary? We get the wrong answer if we
+        # don't do this.
+        params = lmfit.Parameters()
+        for name, param in self.params.items():
+            if param.vary:
+                params[name] = param
```
psavery, line 701: @saransh13 This is relevant for the expression issue.

Issue #834, opened 2025-08-25
matrixutil.vecMVCOBMatrix is incorrect


donald-e-boyce:
### Describe the bug

The `matrixutil.vecMVCOBMatrix` is incorrect.  This is not a major issue since I do not see it used anywhere else in hexrd, but it should be removed as it is not close to being correct.  

### Impact

Not sure if its an issue

### Reproduction

Here is a script that demonstrates the error. It works only if the input rotatio `R` is a 90 degree rotation about a coordinate axis. Otherwise, it is just wrong. The script uses the [HEXRD polycrystal package](https://github.com/HEXRD/polycrystal).

```
"""Test HEXRD vecMVCOBMatrix routine

This works for 90 degree rotations about the coordinate axes, but fails for
more general rotations.

We test this generating rotation matrices from the axial vector by
exponentiating the corresponding skew matrices. The axial vector is parallel
to the axis of rotation and has length the angle of rotation.

We also use the HEXRD polycrystal package.
"""

import numpy as np
from scipy.linalg import expm

from hexrd.matrixutil import symmToVecMV, vecMVToSymm, vecMVCOBMatrix
from polycrystal.utils.tensor_data.voigt_system import VoigtSystem


NL = "\n"
deg90, deg45 = np.pi / 2, np.pi / 4

w = np.array([
    (deg90, 0,  0),
    (0,  deg90, 0.0),
    (0,  0, deg90),
    (0,  0, deg45),
    (1, 2, 3),
    (0.45, 0.87, 1.3),
])
Ws = VoigtSystem.from_parts(skew=w)
Rs = expm(Ws.matrices)

A9 = np.arange(9).reshape((3, 3))
A = 0.5 * (A9 + A9.T)
print("A = \n", A)

# This shows that hexrd and polycrystal give the same result for Voigt system.
v6 = symmToVecMV(A, scale=False)
v6_vs = VoigtSystem(A).symm
assert np.allclose(v6, v6_vs)

for R in Rs:
    print("===== R =\n", R)
    L = vecMVCOBMatrix(R)

    # This the the Voigt 6-vector from HEXRD COB matrix function.
    Lv6 = L @ v6

    # This is the Voigt 6-vector of the rotated matrix.
    v6_rot = symmToVecMV(R @ A @ R.T, scale=False)

    # Now we compare them.
    print(NL, Lv6, NL, v6_rot)

    if np.allclose(Lv6, v6_rot):
        print("Vectors match!")
    else:
[...]

PR #835, opened 2025-08-25
removing incorrect function matrixutil.vecMVCOBMatrix

donald-e-boyce:
# Overview

Removing function `vecMVCOBMatrix` from `matrixutil` as it is incorrect. See #834 .
# Affected Workflows

No workflows are affected.


psavery:
In fact, you can just remove that whole test file because it looks like that is the only test being performed in the file.

psavery (approved):
LGTM, pending CI

PR #856, opened 2025-12-03
Add cif writer

bnmajor:
Added CIF export support to Material:
- `to_cif_block()`: Build a new CifBlock for the material
- `write_cif(filepath)`: Write CIF file for the material

Tests:
- Added `hexrd/tests/material/test_cif_writer.py`


psavery (approved):
If CI passes, then it looks good to me!

PR #864, opened 2025-12-13
Unit test coverage

ZackAttack614:
# Overview

This PR adds unit test coverage for broad swaths of the HEXRD core section. It does not offer complete coverage, but is a stepping stone in the path to 100% library coverage. Upon merge, this will bring total coverage up to 67%.

The only changes to the library code itself amount to fixing which Exceptions are thrown in error cases. Previously, there were assert checks and incorrect RuntimeErrors thrown, which have been fixed for production use to ValueError and TypeError handling instead.

# Affected Workflows

This PR does not affect end users, but does increase the amount of unit test coverage the HEXRD library has for developer use.

# Documentation Changes

No changes to documentation, as the places where there were changes to the library itself were not documented prior. This amounts to only fixes for the Exception types that are thrown.

### DIFF
--- a/tests/core/distortion/test_distortion_dexela_2923.py
+++ b/tests/core/distortion/test_distortion_dexela_2923.py
@@ -9,29 +9,23 @@ def test_dexela_2923_distortion_class():
     params = [0.0] * 8
     distortion_instance = Dexela_2923(params)
 
-    # Test properties
     assert distortion_instance.maptype == "Dexela_2923"
     assert distortion_instance.is_trivial
 
-    # Test invalid length raises
     with pytest.raises(AssertionError):
         distortion_instance.params = [0.0] * 5  # Invalid length
 
-    # Test setting correct length
     distortion_instance.params = np.array([1.0] * 8)
     assert np.array_equal(distortion_instance.params, np.array([1.0] * 8))
     distortion_instance.params = np.array([0.0] * 8)  # Reset
 
     xy_in = np.array([[1.0, 2.0], [3.0, 4.0]])
-
-    # Trivial apply / apply_inverse
     xy_out = distortion_instance.apply(xy_in)
     np.testing.assert_array_equal(xy_out, xy_in)
 
     xy_out_inv = distortion_instance.apply_inverse(xy_in)
     np.testing.assert_array_equal(xy_out_inv, xy_in)
 
-    # Non-trivial apply / apply_inverse
     distortion_instance.params = np.array([1.0] * 8)
     xy_out_nontrivial = distortion_instance.apply(xy_in)
     assert not np.array_equal(xy_out_nontrivial, xy_in)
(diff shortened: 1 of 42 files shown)


hexrd/core/convolution/convolve.py
```diff
@@ -116,6 +116,7 @@ def _copy_input_if_needed(
     return output
 
 
+# TODO: This function does not work when array or kernel are 32-bit float types.
```
psavery: This function was copied from `astropy`. It does tend to be one of the more time-consuming functions that run for things like SNIP algorithms. If `astropy` ever makes significant performance improvements to their code, we may want to copy it again from. [Here](https://github.com/astropy/astropy/blob/8407999759909e5e24cc45fa69bfc5c936564d75/astropy/convolution/convolve.py#L124) is the latest version.

We might should avoid from deviating too far from their code, since we may rely on them for updates.

tests/core/distortion/test_distortion_dexela_2923_quad.py
```diff
@@ -0,0 +1,110 @@
...
+        monkeypatch.setattr(
+            dexela_2923_quad,
+            "_dexela_2923_quad_distortion",
+            dexela_2923_quad._dexela_2923_quad_distortion.py_func,
+        )
+        monkeypatch.setattr(
+            dexela_2923_quad,
+            "_dexela_2923_quad_inverse_distortion",
+            dexela_2923_quad._dexela_2923_quad_inverse_distortion.py_func
+        )
```
psavery: Instead of having functions like this, do you think we should just use the environment variable `NUMBA_DISABLE_JIT=1` when running coverage?

tests/core/distortion/test_distortion_dexela_2923.py
```diff
@@ -0,0 +1,114 @@
...
+    xy_out_inv = distortion_instance.apply_inverse(xy_in)
+    np.testing.assert_array_equal(xy_out_inv, xy_in)
+
+    # Non-trivial apply / apply_inverse
+    distortion_instance.params = np.array([1.0] * 8)
+    xy_out_nontrivial = distortion_instance.apply(xy_in)
+    assert not np.array_equal(xy_out_nontrivial, xy_in)
+
+    xy_out_inv_nontrivial = distortion_instance.apply_inverse(xy_in)
+    assert not np.array_equal(xy_out_inv_nontrivial, xy_in)
```
psavery: I wonder for tests like this if we should also verify that the output is correct? This currently just verifies that the numbers changed.

psavery: We'll do it another time.

tests/core/instrument/test_hedm_instrument.py
```diff
@@ -0,0 +1,1306 @@
...
+        """
+        Ensure that when check_only=False and omega_to_frame returns valid frame
+        the reflection is processed and completeness is computed.
+        """
+        
+        # --- Instrument Setup ---
+        det = instr.detectors['det1']
+        det.tvec = np.array([0.0, 0.0, -1000.0])
+        instr.tvec = np.zeros(3)
+        type(det).pixel_area = PropertyMock(return_value=1.0)
```
psavery: The detector's `pixel_area` attribute is being modified here.

psavery (approved):
LGTM

PR #872, opened 2026-01-07
Logging

ZackAttack614:
# Overview

This PR changes most `print` statements into useful logging statements. The goal is to simplify and quiet the terminal to make its output more useful to the end user.

### DIFF
--- a/hexrd/core/material/symbols.py
+++ b/hexrd/core/material/symbols.py
@@ -232,13 +232,13 @@ def PrintPossibleSG(xtal_sys):
         sgmax = 2
         sgmin = 1
 
-    pstr = ''
     for i in range(sgmin, sgmax + 1):
         j = i - sgmin + 1
-        pstr += f'{i}:{pstr_spacegroup[i - 1]}\t'
+        pstr = f'{i}:{pstr_spacegroup[i - 1]}\t'
         if j % 4 == 0 or j == sgmax:
-            break
-    logger.info(pstr)
+            print(pstr)
+        else:
+            print(pstr, end='')
 
     return sgmin, sgmax
 


hexrd/core/material/material.py
```diff
@@ -587,7 +589,7 @@ def _readCif(self, fcif=DFLT_NAME + '.cif'):
         try:
             cif = ReadCif(fcif)
         except RuntimeError:
-            print("file not found")
+            logging.warning("File not found")
```
psavery: The rest of the code will fail after this. I know that was the behavior before, but maybe you can call `logger.exception()` instead of `logger.warning()`, and then add a line with `raise` to re-raise the exception?

hexrd/core/material/mksupport.py
```diff
@@ -15,9 +17,7 @@
 
 
 def mk(filename, xtalname):
-
-    # print some initial information for the user
-    print(pstr_mkxtal)
+    logging.debug(pstr_mkxtal)
```
psavery:
```suggestion
    logging.info(pstr_mkxtal)
```

I have a feeling this should be `info` instead.

hexrd/hedm/findorientations.py
```diff
@@ -569,14 +560,14 @@ def generate_eta_ome_maps(cfg, hkls=None, save=True):
...
-    logger.info("\t\t...took %f seconds", timeit.default_timer() - start)
+    logger.debug("\t\t...took %f seconds", timeit.default_timer() - start)
 
     if save:
         # save maps
         fn = cfg.find_orientations.orientation_maps.file
         eta_ome.save(fn)
 
-        logger.info(f'saved eta/ome orientation maps to "{fn}"')
+        logger.debug(f'saved eta/ome orientation maps to "{fn}"')
```
psavery:
```suggestion
        logger.info(f'saved eta/ome orientation maps to "{fn}"')
```

Maybe this should be kept as info for confirmation of the output file location.

hexrd/hedm/findorientations.py
```diff
@@ -850,10 +841,10 @@ def find_orientations(
             doMultiProc=ncpus > 1,
             nCPUs=ncpus,
         )
-        logger.info("\t\t...took %f seconds", timeit.default_timer() - start)
+        logger.debug("\t\t...took %f seconds", timeit.default_timer() - start)
```
psavery:
```suggestion
        logger.info("\t\t...took %f seconds", timeit.default_timer() - start)
```

hexrd/hedm/findorientations.py
```diff
@@ -850,10 +841,10 @@ def find_orientations(
...
     completeness = np.array(completeness)
 
-    logger.info(
+    logger.debug(
```
psavery:
```suggestion
    logger.info(
```

hexrd/hedm/grainmap/nfutil.py
```diff
@@ -112,11 +112,8 @@ def finish(self, name):
         entry = self.timing.pop()
         assert name == entry[0]
         total = t - entry[2]
-        logging.info(
-            "%s took %8.3fs (%8.6fs per item).",
-            entry[0],
-            total,
-            total / entry[1],
+        logging.debug(
```
psavery:
```suggestion
        logging.info(
```

hexrd/powder/wppf/WPPF.py
```diff
@@ -2103,24 +2101,22 @@ def Refine(self):
             self.Rwplist = np.append(self.Rwplist, self.Rwp)
             self.gofFlist = np.append(self.gofFlist, self.gofF)
 
-            msg = (
-                f"Finished iteration. Rwp: "
+            logging.debug("Finished iteration. Rwp: "
```
psavery:
```suggestion
            logging.info("Finished iteration. Rwp: "
```

This info is essential for the user.

psavery (commented):
Looks good for the most part!

I think we're gonna change the logger calls to use local loggers, like so:

```python
logger = logging.getLogger(__name__) 
```

Plus there are a few suggested changes.

PR #884, opened 2026-01-29
Documentation & Typing Part 1

ZackAttack614:
# Overview

This PR is the first of many which will add docstrings and typing to the HEXRD library.

# Affected Workflows

No workflows are affected by this PR.

# Documentation Changes

This PR adds explicit documentation for functions in a few places, including argument types and descriptors. It also adds static type identifiers to a number of variables, parameters, and return types for clarity.


hexrd/core/instrument/detector.py
```diff
@@ -1210,65 +1205,95 @@ def _generate_bilinear_interp_dict(
...
-        """
-        Generate points on Debye_Scherrer rings over the detector.
-
-        !!! it is assuming that rmat_s is built from (chi, ome) as it the case
-            for HEDM!
+        pd: PlaneData | np.ndarray, # TODO: Make a different function for array input
+        merge_hkls: bool = False,
+        delta_tth: Optional[float] = None,
+        delta_eta: float = 10.0,
+        eta_period: list[float] | np.ndarray = [-np.pi, np.pi],
```
psavery: Making default values be mutable types, such as a `list`, can produce very hard-to-track-down bugs. It's better to either set it to `None`, use a `tuple`, or a constant numpy array (which should ideally have the `writeable` flag set to `False`.

hexrd/core/instrument/hedm_instrument.py
```diff
@@ -587,32 +578,38 @@ def __init__(
                 self.physics_package = instrument_config['physics_package']
 
             xrs_config = instrument_config['beam']
-            is_single_beam = 'energy' in xrs_config and 'vector' in xrs_config
-            if is_single_beam:
```
psavery: Having simple variables like this can improve read-ability, since you know we are interpreting those conditions as having a "single beam" (and can help you reason through changes if needed).

hexrd/hedm/findorientations.py
```diff
@@ -419,24 +425,17 @@ def load_eta_ome_maps(cfg, pd, image_series, hkls=None, clean=False):
...
     #  if True, only do median subtraction
     #  if scalar, do median + LoG filter with that many pixels std dev
-    if filter_maps:
-        if not isinstance(filter_maps, bool):
-            sigm = const.fwhm_to_sigma * filter_maps
-            logger.info("filtering eta/ome maps incl LoG with %.2f std dev", sigm)
-            _filter_eta_ome_maps(eta_ome, filter_stdev=sigm)
-        else:
-            logger.info("filtering eta/ome maps")
-            _filter_eta_ome_maps(eta_ome)
```
psavery: Isn't this changing the behavior? If `filter_maps` is a float in the yaml file, it gets interpreted as `fwhm` to use as input for the next function. If it's a bool, then it uses the default `fwhm`.

This could break some input files.

hexrd/core/instrument/hedm_instrument.py
ZackAttack614: Would this variable not be better as a comment? I'm not a fan of single-use variables (and neither is pylint)

hexrd/core/utils/compatibility.py
```diff
@@ -1,13 +1,16 @@
 from importlib.metadata import version
 
+import numpy as np
+from numpy.typing import NDArray
+
 import h5py
 
 
-def h5py_read_string(dataset):
+def h5py_read_string(dataset: h5py.Dataset) -> NDArray[np.bytes_] | NDArray[np.str_]:
```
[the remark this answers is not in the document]
ZackAttack614: Possibly, but the documentation for h5py is a bit frustrating to parse. Do you have a proposed change?

hexrd/core/imageseries/baseclass.py
```diff
@@ -49,10 +48,6 @@ def get_region(self, frame_idx: int, region: RegionType) -> np.ndarray:
         return self._adapter.get_region(frame_idx, region)
 
     def set_option(self, key: str, value: Any):
-        if not hasattr(self._adapter, 'set_option'):
-            msg = f'"{type(self._adapter)}" has not implemented "set_option"'
-            raise NotImplementedError(msg)
```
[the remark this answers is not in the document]
ZackAttack614: If my typing is correct, `self._adapter` is of type `ImageSeriesAdapter`. That class already raises a `NotImplementedError` when `set_option` is called without implementation. Indeed, it's hard for me to imagine this code ever entering this if statement. However, the typing could be wrong.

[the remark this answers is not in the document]
ZackAttack614: In this PR, the change is implemented:

https://github.com/HEXRD/hexrd/blob/documentation/hexrd/core/imageseries/load/__init__.py

hexrd/core/imageseries/load/eiger_stream_v1.py
```diff
@@ -67,35 +66,26 @@ def __len__(self):
         return len(self._data_group)
 
     def __getstate__(self):
-        # Remove any non-pickleable attributes
-        to_remove = [
-            '__h5file',
-        ]
+        # Remove any non-pickleable attributes. Prefix them with the private prefix
+        to_remove = [f'_{self.__class__.__name__}_h5file']
```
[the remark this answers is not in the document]
ZackAttack614: I'm not sure that I feel comfortable personally testing this change. If you would like, please go ahead and give it a shot yourself, I'm happy to incorporate that here if appropriate.

hexrd/core/instrument/hedm_instrument.py
psavery: Sure!

hexrd/core/utils/panel_buffer.py
```diff
@@ -76,12 +76,14 @@ def panel_buffer_from_str(name: str, panel: Detector) -> np.ndarray:
     return roi_buffer
 
 
+# TODO: Broken function - dir_path.glob is not defined
```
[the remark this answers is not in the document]
psavery:
```python
>>> from hexrd.core.utils.panel_buffer import valid_panel_buffer_names
>>> valid_panel_buffer_names()
['chess-eiger-stream-v2', 'chess-eiger-stream-v1']
```

I put a breakpoint in there. The `dir_path` object is a `PosixPath`, which does have `glob` defined on it. Maybe the type checking is incorrect about what type is being returned from the `importlib` library?

hexrd/core/utils/panel_buffer.py
```diff
@@ -76,12 +76,14 @@ def panel_buffer_from_str(name: str, panel: Detector) -> np.ndarray:
...
 def valid_panel_buffer_names() -> list[str]:
     dir_path = importlib.resources.files(hexrd.core.resources.panel_buffers)
     return [path.stem for path in dir_path.glob('*.npz')]
 
 
 # Cache this so we only read from disk once
+# TODO: Broken function - path.exists is not defined
```
psavery: I think this also works. We are actively using it.

hexrd/core/material/crystallography.py
```diff
@@ -1440,19 +1481,18 @@ def getHKLID(
...
-            if isinstance(hkl, np.ndarray):
-                # if is ndarray, assume is 3xN
-                return [self._getHKLID(x, master=master) for x in hkl.T]
-            else:
-                return [self._getHKLID(x, master=master) for x in hkl]
+        if isinstance(hkl, np.ndarray):
+            # if is ndarray, assume is 3xN
+            return [self._getHKLID(x, master=master) for x in hkl.T]
+        elif isinstance(hkl, tuple) or isinstance(hkl, list):
+            return [self._getHKLID(x, master=master) for x in hkl]
```
psavery: Is this behavior different than before? It looks like the code `[self._getHKLID(x, master=master) for x in hkl]` used to not run if it was a tuple, because of the `hasattr(hkl, '__setitem__')` check.

hexrd/core/material/crystallography.py
```diff
@@ -2134,12 +2202,12 @@ def lorentz_factor(tth: np.ndarray) -> np.ndarray:
 
 
 def polarization_factor(
-    tth: np.ndarray,
+    tth: NDArray[np.float64],
+    eta: Optional[NDArray[np.float64]] = None,
```
psavery: I think we shouldn't re-order this arguments. Can you move `eta` back down below `unpolarized`?

hexrd/core/instrument/hedm_instrument.py
```diff
@@ -305,18 +307,20 @@ def _parse_imgser_dict(imgser_dict, det_key, roi=None):
             )
             if isinstance(images_in, OmegaImageSeries):
                 # if it was an OmegaImageSeries, must re-cast
+                assert isinstance(ims, ProcessedImageSeries)
                 ims = OmegaImageSeries(ims)
         elif isinstance(images_in, np.ndarray):
             # 2- or 3-d array of images
             ndim = images_in.ndim
+            images_in = np.asarray(images_in)
```
psavery: Is this needed? Didn't it just verify above that it's an `ndarray`?

hexrd/core/fitting/calibration/lmfit_param_handling.py
```diff
@@ -18,17 +19,16 @@
...
 from hexrd.core.fitting.calibration.relative_constraints import (
-    RelativeConstraints,
     RelativeConstraintsType,
 )
 
 
 # First is the axes_order, second is extrinsic
 DEFAULT_EULER_CONVENTION = ('zxz', False)
 
-EULER_CONVENTION_TYPES = dict | tuple | None
```
psavery: Conceptually, `None` is a valid euler convention here. 

It means there is no euler convention, and we use the exponential map parameters instead.

hexrd/core/material/crystallography.py
ZackAttack614: This is true. I've pushed a change which tries to resolve this with clear typing.

psavery (approved):
LGTM

PR #886, opened 2026-02-03
Add ability to preprocess eiger-stream-v2 files

psavery:
Previously, we only allowed processing of eiger-stream-v1 files. This adds automatic detection so that `eiger-stream-v1` or `eiger-stream-v2` is selected automatically.

It also adds a couple of command-line options specific for `eiger-stream-v2`, namely setting the threshold option and the multiplier.

A test was also added that exercises the full preprocessing.

This will be used potentially next week at CHESS.

Example of running this:

```bash
hexrd preprocess eiger -n 1 --eiger_stream_v2_threshold 'man_diff' --eiger_stream_v2_multiplier 0.75 ./eiger_stream_v2_test_dataset.h5
```

That is on the example file [here](https://github.com/HEXRD/examples/tree/master/eiger/eiger_stream_v2).

It should output a file named `test-cachefile.npz`.

### DIFF
--- a/hexrd/hedm/preprocess/preprocessors.py
+++ b/hexrd/hedm/preprocess/preprocessors.py
@@ -1,26 +1,27 @@
 import logging
+import os
+import time
+from typing import Any, Optional, Union, Sequence, cast
 
+import h5py
+from numpy.typing import NDArray
+from numpy import float32
+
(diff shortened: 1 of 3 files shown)


bnmajor (approved):
LGTM!

PR #887, opened 2026-02-04
Transforms cleanup

ZackAttack614:
# Overview

This PR removes a lot of unused and dead code from the transforms module of HEXRD. We aniticpate that this may affect users who have very old scripts (5 years or more) or others who have used the `xf` or `old_xfcapi` components of the library. As this code has not been supported in many years, it is time for this code to be removed from the library. We will support the transition to the newer code as needed, on a case-by-case basis.

Some documentation about transitioning from the `old_xfcapi` code can be found in the Wiki:

https://github.com/HEXRD/hexrd/wiki/old_xfcapi-Conversion-Guide

# Affected Workflows

Users who have scripts which use the `old_xfcapi` or `xf` components of HEXRD's transforms module will need to move off of this and into the new API, using the conversion guide above.

# Documentation Changes

Documentation regarding transition has been written, and uploaded into the Github Wiki. Documentation covering the removed code has been deleted.


tests/test_preprocess.py
```diff
@@ -50,38 +41,14 @@ def test_save_load_eiger(ceria_examples_path):
...
-    with pytest.raises(RuntimeError):  # required argument 'absolute_path' is missing
+    with pytest.raises(
+        RuntimeError
+    ):  # required argument 'absolute_path' is missing
         args.validate_arguments()
     args.absolute_path = str(ceria_examples_path)
     args.validate_arguments()
 
 
-def test_save_load_eiger_stream_v2(eiger_stream_v2_examples_path: Path):
```
psavery: Same here - new preprocessor features were deleted by accident.

hexrd/hedm/config/findorientations.py
```diff
@@ -226,16 +225,14 @@ def fiber_ndiv(self):
...
+        hkls: int | str | None = self._cfg.get(
             'find_orientations:orientation_maps:active_hkls', default='all'
         )
         if isinstance(hkls, int):
             return [hkls]
-        elif hkls == 'all' or hkls is None:
+        if hkls == 'all':
             return None
-
-        assert isinstance(hkls, list)
```
psavery: Were the changes in this file intentional? If so, that's fine! I'm just making sure they were intentional.

PR #889, opened 2026-02-06
Remove all used print() statements.

### LINKED ISSUE #350: Replace any print statements with appropriate logging calls

ZackAttack614:
Resolves #350

# Overview

This PR converts all useful print statements into logger statements, and removes the rest.

# Affected Workflows

No workflows are affected by this PR.

# Documentation Changes

No documentation changes are required for this PR.

### DIFF
--- a/hexrd/core/material/symbols.py
+++ b/hexrd/core/material/symbols.py
@@ -220,6 +220,7 @@ def PrintPossibleSG(xtal_sys):
         sgmax = 2
         sgmin = 1
 
+    line = ""
     for i in range(sgmin, sgmax + 1):
         j = i - sgmin + 1
         pstr = f"{i}:{pstr_spacegroup[i - 1]}\t"


tests/fit_grains_check.py
```diff
@@ -138,7 +138,3 @@ def compare_grain_fits(
         ctol=ctol,
         vtol=vtol,
     )
-    if cresult:
-        print("test passed")
-    else:
-        print("test failed")
```
psavery: If we're keeping this `__main__` section, isn't this final printing the most important part of it? Otherwise, if you ran it, you wouldn't know if it passed or failed.

hexrd/core/instrument/hedm_instrument.py
```diff
@@ -2245,7 +2245,7 @@ def __init__(self, filename):
             self.fid = filename
         else:
             self.fid = open(filename, 'w')
-        print(self._header, file=self.fid)
+        logger.info(self._header, file=self.fid)
```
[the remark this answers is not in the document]
psavery: Maybe using the `print()` statement before implicitly added a `'\n'` after each call too, so you might need to add `+ '\n'` to the writes.

hexrd/core/material/symbols.py
```diff
@@ -225,7 +225,7 @@ def PrintPossibleSG(xtal_sys):
         pstr = f"{i}:{pstr_spacegroup[i - 1]}\t"
         line += pstr
 
-        if j % 4 == 0 or i == sgmax:
+        if j % 4 == 0 or j == sgmax:
```
saransh13: `j` only runs from `1` to `sgmax-sgmin`. the original logic is correct. The logic ensures that only 4 symbols are printed on each line. For e.g. `PrintPossibleSG(4)`, with the old logic yields

```168: P 6       	169: P 61      	170: P 65      	171: P 62      
172: P 64      	173: P 63      	174: P -6      	175: P 6/m     
176: P 63/m    	177: P 6 2 2   	178: P 61 2 2  	179: P 65 2 2  
180: P 62 2 2  	181: P 64 2 2  	182: P 63 2 2  	183: P 6 m m   
184: P 6 c c   	185: P 63 c m  	186: P 63 m c  	187: P -6 m 2  
188: P -6 c 2  	189: P -6 2 m  	190: P -6 2 c  	191: P 6/m m m 
192: P 6/m c c 	193: P 63/m c m	194: P 63/m m c	```

File: tests/fit_grains_check.py (hunk 1 above)
Comment (ZackAttack614) [reply tests/fit_grains_check.py:144]: This is a really good point. Upon closer inspection, this fit_grains_check.py file doesn't actually need to exist. I've moved `compare_fit_grains` to where it was called in `tests/test_fit-grains.py` and removed the script entirely. It seems this pytest file was created to reproduce a lot of the testing done in that old `__main__` section in a pytest format, so this should be the proper way to handle things.

File: hexrd/core/instrument/hedm_instrument.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (ZackAttack614) [reply hexrd/core/instrument/hedm_instrument.py]: Good catch, thanks!

File: hexrd/core/instrument/hedm_instrument.py [hunk 4]
```diff
@@ -2294,7 +2294,7 @@ def dump_patch(
                 self._delim.join(np.tile('{:<23.16e}', 10)).format(*res[7:]),
             ]
         )
-        print(output_str, file=self.fid)
+        logger.info(output_str, file=self.fid)
```
psavery: Shouldn't this one be `self.fid.write(output_str + '\n')` like the other changes?
```suggestion
        self.fid.write(output_str + '\n')
```

hexrd/core/material/symbols.py
```diff
@@ -222,11 +222,12 @@ def PrintPossibleSG(xtal_sys):
 
     for i in range(sgmin, sgmax + 1):
         j = i - sgmin + 1
-        pstr = f'{i}:{pstr_spacegroup[i - 1]}\t'
+        pstr = f"{i}:{pstr_spacegroup[i - 1]}\t"
+        line += pstr
```
psavery, line 227: Do we need to add a `line = ""` outside of this `for` loop? It looks like it. Otherwise `line` would be uninitialized.

PR #908, opened 2026-03-26
Indexer cleanup

donald-e-boyce:
# Overview
This cleans up the documentation for `hexrd.hedm.indexer.py` and adds python type hints.

This update was mostly done by Claude, with some minor adjustments by me. The documentation is way better, and the type hints are very good too.  

I've run several problems on both the `master` and the `indexer-cleanup` branches, and all of the indexer output is identical for all of the problems, including `accepted-orientations-ruby.dat`,  `eta-ome-maps-ruby.npz`, `grains.out`, and `scored-orientations-ruby.npz`.  

Comparing the log files, howevers, shows a peculiar difference.   If we strip the log files of their timestamp and diff, ignoring elapsed time differences, we get:

First, comparing new branch to current a recent master (2 or 3 days ago):  
```
(hexrd-py-3.12) (MacBook-Pro-2023: single_GE) 1288. diff ruby/indexer/find-orientations-ruby.log.strip ruby/master.bak/
11c11
< hexrd.hedm.findorientations - 	mean reflections per grain: 486
---
> hexrd.hedm.findorientations - 	mean reflections per grain: 485
```

And also comparing the latest master branch (today) to the recent master: 
```
(hexrd-py-3.12) (MacBook-Pro-2023: single_GE) 1289. diff ruby/master/find-orientations-ruby.log.strip ruby/master.bak/
< hexrd.hedm.findorientations - 	mean reflections per grain: 487
---
> hexrd.hedm.findorientations - 	mean reflections per grain: 485
15c15
```
The difference does not seem to be related to the changes in `indexer.py`, as the current master branch gives different results from a recent one. Also, the mean reflections per grain is calculated after the indexer is called, in the `find_orientations.create_clustering_parameters()` function.

# Affected Workflows

This is part of the find-orientations workflow.

# Documentation Changes

Only internal documentation has changed. No external documentation needs to be changed.

PR #909, opened 2026-03-26
Indexer cleanup

donald-e-boyce:
# Overview

See #908, unintentionally closed.  This cleans up documentation and typing in indexer.py.  The latest changes the `ParamMP` variable to a TypedDict.

### DIFF
--- a/hexrd/hedm/indexer.py
+++ b/hexrd/hedm/indexer.py
@@ -26,134 +26,159 @@
 # the Free Software Foundation, Inc., 59 Temple Place, Suite 330,
 # Boston, MA 02111-1307 USA or visit <http://www.gnu.org/licenses/>.
 # =============================================================================
+"""Spherical map-based orientation indexing for HEDM (paintGrid algorithm).
+
+This module implements the paintGrid algorithm, which scores a set of trial
+crystal orientations against measured eta-omega intensity maps produced by
+High Energy X-Ray Diffraction (HEXRD) experiments.
+
+For each trial orientation, the algorithm:
+
+1. Computes the predicted diffraction angles (tth, eta, omega) for every
+   symmetry-equivalent HKL reflection using the orientation matrix and the
+   reciprocal-lattice B-matrix.
+2. Filters predictions that fall outside the specified valid eta/omega ranges.
+3. Checks, with a pixel-dilation tolerance, whether each surviving prediction
+   coincides with measured intensity above a per-HKL threshold in the
+   corresponding eta-omega map.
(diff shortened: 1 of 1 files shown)


hexrd/hedm/indexer.py
```diff
@@ -26,134 +26,192 @@
...
+    before any worker calls :func:`paintGridThis`.
+    """
 
-paramMP = None
-nCPUs_DFLT = multiprocessing.cpu_count()
-logger = logging.getLogger(__name__)
+    symHKLs: NDArray
+    symHKLs_ix: NDArray
+    wavelength: float
+    hklList: list[Any]
```
ZackAttack614: We should really avoid `Any` if at all possible when typing. Is it safe to assume that `hklList` is a list of integers, for example? If it can be multiple types, we can address that too, e.g. `list[int | float]`.

ZackAttack614 (commented):
This is definitely better! I'll propose that we add in typing for the elements of numpy arrays, as well. We've done that in most places, e.g. [here](https://github.com/HEXRD/hexrd/blob/master/hexrd/core/material/crystallography.py#L67-L72). That allows us to statically analyze the script and know whether the data type is an int, float64, bool, or str - all of which are matrix element types we've worked with in the HEXRD ecosystem.

donald-e-boyce:
So, in your file,  what is the difference between `np.integer` and `np.signedinteger`? And is `np.floating`preferred for any float that is not explicitly something more specific?  In other words, is `np.floating`, `np.integer`, and `np.bool` enough?  And why are there numpy types for non-NDArrays?  e.g. `tTheta: np.floating` -- couldn't that just be `float`?
  
```
class HKLData(TypedDict):
    hklID: int
    hkl: NDArray[np.integer]
    tTheta: np.floating
    dSpacings: np.floating
    tThetaLo: np.floating
    tThetaHi: np.floating
    latPlnNrmls: NDArray[np.floating]
    symHKLs: NDArray[np.signedinteger]
    centrosym: bool
```

donald-e-boyce:
So, technically, it looks like it's wrong to use numpy datatypes for regular floats and ints.

```
In [2]: x = 10.0

In [3]: isinstance(x, np.floating)
Out[3]: False
```

donald-e-boyce:
OH, I see. It's a typedict. I thought they were type hints.
never mind

donald-e-boyce:
OK, I added the numpy scalar types and pushed everything up.  Should be good now.

ZackAttack614 (approved):
Looking good, thanks very much for adding to the documentation and typing effort for HEXRD, Don!

PR #910, opened 2026-03-26
Correctly save a `tThWidth` of `None`

psavery:
# Overview

Before, if the `tThWidth` was set to `None`, which is a valid setting, saving the material and reloading it would reset it to the default value, rather than restoring the value of `None`.

This PR fixes the bug so that `None` is properly saved and restored, as it should be.

# Affected Workflows

All workflows, but it probably doesn't affect most users (only a select few would set it to `None`).

# Documentation Changes

None


saransh13 (approved):
LGTM!

PR #911, opened 2026-03-26
Simulate laue pattern

ZackAttack614:
# Overview

The `simulate_laue_pattern` functions for `Detector` and `HEDMInstrument` have been moved into the laue workflow, from the core directory. This is a necessary step in ensuring that the new architecture keeps workflow-specific functions and classes in the correct workflow directories.

- `hexrd.core.instrument.detector.Detector.simulate_laue_pattern` has been removed from the context of a class method, and made into a standalone function in the `laue` workflow here: `hexrd.laue.simulate.simulate_laue_pattern_on_panel`
- `hexrd.core.instrument.hedm_instrument.HEDMInstrument.simulate_laue_pattern` has been removed from the context of a class method, and made into a standalone function in the `laue` workflow here: `hexrd.laue.simulate.simulate_laue_pattern_on_instrument`

# Affected Workflows

This change affects users of the Laue workflow simulating laue patterns.

# Documentation Changes

- [ ] Migration instructions must be written in the library wiki before this PR can be accepted.


hexrd/laue/fitting/calibration/laue.py
```diff
@@ -15,6 +15,7 @@
 from hexrd.core.rotations import angleAxisOfRotMat, RotMatEuler
 from hexrd.core.transforms import xfcapi
 from hexrd.core.utils.hkl import hkl_to_str, str_to_hkl
+from hexrd.laue.simulation import simulate_laue_pattern_on_panel
```
psavery: This file uses `simulate_laue_pattern_on_instrument` too, so that should also be imported.

Maybe we should add automatic linting to the PRs to check for things like this?

ZackAttack614: Fixed, thanks! I agree in principle that we should add auto-linting, but there are so many linter issues with the current codebase that it wouldn't add much practicality just yet. I'd love to get us to that point soon with this rearchitecture!

psavery (approved):

PR #915, opened 2026-04-02
March dollase texture

saransh13:
adding the march-dollase texture model to wppf. The model can account for fiber texture related intensity modulation in the azimuthally averaged lineout.


psavery (approved):
LGTM

PR #917, opened 2026-04-03
fixes bug in unitVector

### LINKED ISSUE #916: bug in unitVector
### Describe the bug

There is a bug in the `unitVector` function when the input is 1D.  It then returns a matrix with the unit vector repeated in rows.

### Impact

Will eventually cause someone a problem

### Reproduction

Here is an example of the error.
```
"""Test unit vector"""

import numpy as np

from hexrd.core.matrixutil import unitVector

x = np.array([1.0, 2.0, 3.0])
u = unitVector(x)
print(u)
```
And running it:

```
> python test_uv.py
[[0.26726124 0.53452248 0.80178373]
 [0.26726124 0.53452248 0.80178373]
 [0.26726124 0.53452248 0.80178373]]
```

### Version

```shell
Latest version.
```

### Affected Workflows

- [x] core
- [ ] EDD
- [ ] HEDM
- [ ] Powder
- [ ] Laue

donald-e-boyce:
# Overview

This fixes #916, improves the docstring and adds two new tests.

### DIFF
--- a/hexrd/core/matrixutil.py
+++ b/hexrd/core/matrixutil.py
@@ -74,13 +74,25 @@ def rowNorm(a):
 
 
 def unitVector(a):
+    """ normalize array of column vectors (hstacked, axis = 0)
+
+    Parameters
+    ----------
+    a: numpy array
+       array to normalize; shape should be (m, n) or (m,)
+
+    Returns
+    -------
+    anrm: numpy array
+       array with shape (m, n)  with each column normalized to magnitude 1.0
     """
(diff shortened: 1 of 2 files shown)


psavery:
It looks good, but a pre-existing test is failing [here](https://github.com/HEXRD/hexrd/blob/b109e03221482fd28b758136b13c035f283dd930/tests/matrix_util/test_norms.py#L38). I think you need to update it to expect the new shape that gets returned.

PR #926, opened 2026-04-22
Add de la Vallée Poussin kernel on SO(3)

bnmajor:
# Overview

Adds a de la Vallée Poussin kernel implementation on SO(3) for constructing smooth orientation distribution functions.

**New modules:**
- SO3Kernel
- DeLaValleePoussinKernel

**New public API:**
- DeLaValleePoussinKernel(halfwidth: float) - constructor, halfwidth in radians
- .eval(R1, R2) - evaluate kernel between rotation matrices
- .eval_centered(R, center) - convenience wrapper for unimodal distributions
- .misorientation_angle(R1, R2) - compute misorientation angle from 3×3 rotation matrices

**Mathematical basis:**
The kernel is defined as K(ω) = C · cos(ω/2)^(2κ), where C is a Beta-function normalization constant and κ is derived analytically from the halfwidth via κ = ln(0.5) / (2·ln(cos(h/2))).

# Affected Workflows

N/A

# Documentation Changes

User-facing documentation for the texture module will be added in the module integration PR once the full API is finalized.


psavery:
To do: 

* move the `texture` directory inside the `phase_transition` directory
* Add a simple test for running the `__str__` and `__repr__` functions
* Keep in mind that for `misorientation_angle`, we may need to take into account crystal symmetry (depending on how it will be used), and `misorientation` in the `rotations.py` file may be able to work.

psavery (approved):
I added a few more tests, and made a small API change to the symmetry-reduced misorientation so it runs much faster - it now uses the pre-existing `rotations.misorientation` in hexrd for batched, symmetry-aware misorientation instead of a per-pair loop (~40x faster on cubic symmetry). The one behavioral note: with symmetry enabled, one of the two operands must be a single reference orientation (it raises otherwise), which matches how the kernel is actually evaluated.

I'm rebasing #939 and #941 onto these changes, so those don't need fixing on your end. This PR looks good to me!

psavery:
@bnmajor merge when ready!

Issue #866, opened 2025-12-15
Inconsistent behavior in material.unitcell.color_orientations()


rjk5987:
### Describe the bug

Hello,

The function color_orientations() under material/unitcell behaves differently depending on whether a single rotation matrix (3x3) or a stack of rotation matrices (Nx3x3) is passed in. I don't think this is the intended behavior.

In the case of a single rotation matrix (currently lines 1390-1391), the if statement calls `np.atleast_3d(rmats).T`, which reshapes the array to 1x3x3 but also transposes the rotation matrix. For example, if I generate the following matrix:
```
np.arange(9).reshape(3, 3)
Out: 
array([[0, 1, 2],
       [3, 4, 5],
       [6, 7, 8]])
```
and apply this function, the following is returned:
```
np.atleast_3d(np.arange(9).reshape(3, 3)).T
Out: 
array([[[0, 3, 6],
        [1, 4, 7],
        [2, 5, 8]]])
```
If a stack of rotation matrices is passed in (else case, currently lines 1392-1396), then no transpose is applied to the matrices. As such, different colors will be returned depending on whether the rotation matrices are passed in individually or all at once. As currently written, input rotation matrices would need to be crystal-to-sample for the stack (since these only get transposed once in the for loop in lines 1403-1404) versus sample-to-crystal for individual matrices (since these get transposed twice) to obtain the intended result.

`np.atleast_3d(rmats.T).T` or `rmats[None]` are examples of potential replacements for line 1391 that would unify the behavior of the function by reshaping without transposing:
```
np.atleast_3d(np.arange(9).reshape(3, 3).T).T
Out: 
array([[[0, 1, 2],
        [3, 4, 5],
        [6, 7, 8]]])

np.arange(9).reshape(3, 3)[None]
Out[70]: 
array([[[0, 1, 2],
        [3, 4, 5],
        [6, 7, 8]]])
```

### Impact

Will eventually cause someone a problem

### Reproduction

Below is an array with two example rotation matrices:
```
rots = np.array([[[ 0.8513851 ,  0.51820825, -0.08126269],
                  [-0.51602112,  0.85525329,  0.04758169],
                  [ 0.0941574 ,  0.00142293,  0.99555631]],
                 [[ 0.93586175, -0.34601698, -0.06659611],
                  [ 0.3489187 ,  0.93637851,  0.03809232],
                  [ 0.04917857, -0.05888577,  0.99705262]]])
```
[...]

PR #927, opened 2026-04-24
Update unitcell.py

rjk5987:
Submitting pull request at the request of @darrencpagan. Unified color_orientations() behavior for individual rotation matrices vs. stacks of rotation matrices (see issue #866).

# Overview

See issue #866 for a description of the issue, including minimal test cases for reproducing the behavior.

# Affected Workflows

Because the behavior differs between individual rotation matrices and stacks of rotation matrices, this may result in unintended behavior (incorrect coloring) for users implementing this function in their visualization workflows.

### DIFF
--- a/hexrd/core/material/unitcell.py
+++ b/hexrd/core/material/unitcell.py
@@ -1358,7 +1358,7 @@ class unitcell:
         first make sure that the rotation matric is size nx3x3
         '''
         if rmats.ndim == 2:
-            rmats = np.atleast_3d(rmats).T
+            rmats = np.atleast_3d(rmats.T).T
         else:
             assert (
                 rmats.ndim == 3


ZackAttack614:
Thank you for this PR. For us to review this properly, can you provide pytest coverage in the `tests` folder?

I will also propose using the following line instead, to adhere to standard Pythonic practice - please give it a try:

```python
rmats = rmats.reshape(-1, 3, 3)
```

ZackAttack614:
Thank you very much for this PR. We've gone ahead with a [fix](https://github.com/HEXRD/hexrd/pull/929) here, and this should cover your use case. Please reach out if this doesn't solve your problem, and we can re-open the PR.

Cheers,
Zack

PR #929, opened 2026-05-01
fix inconsistency in behavior of single and many rotation matrix passed to Materials.unitcell.color_orientations.

saransh13:
Also includes some ruff linter fixes for `Materials.unitcell` 

actual code change 1360-1380

This fixes issue #866 

A test for `Material.unitcell.color_orientations` will be submitted as a separate PR.

## Overview
If a single rotation matrix was passed, it was converted to the shape `(3, 3, 1)` using the `numpy.atleast_3d` function. However, if many rotation matrices were passed, the shape was assumed to be `(n, 3, 3)` where `n` was the number of rotation matrices. The fix makes the shape of a single rotation matrix to to `(1, 3, 3)` and is now consistent with the expected shape.

The reference directions were then converted to the crystal frame directions by looping over the rotation matrices. That has now been made simpler by separating the transpose  (`line 1372`) with the matrix-vector multiplication (`line 1373`).

## tests added
add tests for color_orientations function. The function specifically tests:
- the single matrix and multiple matrix behavior is the same.
- calculation of directions to fundamental triangle
- calculation of `theta` and `rho` values which are at the heart of the color calculation
- color rgb values


ZackAttack614 (approved):
Thank you, Saransh!

PR #930, opened 2026-05-01
Color orientations tests

saransh13:
## Overview
add tests for color_orientations function. The function specifically tests:
- the single matrix and multiple matrix behavior is the same.
- calculation of directions to fundamental triangle
- calculation of `theta` and `rho` values which are at the heart of the color calculation
- color rgb values


saransh13:
These changes were submitted along with #929

PR #932, opened 2026-05-01
Call `conda-build` executable directly

psavery:
# Overview

I think `conda build` only works if `conda-build` is installed in the base environment. Not sure how it used to work... but `conda-build` as a command will hopefully work.


ZackAttack614 (approved):
This seems to have solved the linux build issue. Thank you @psavery !

PR #934, opened 2026-05-06
Update WPPF params, even if none are varied

psavery:
# Overview

Ensure we update the WPPF parameters before computing a spectrum, even if no WPPF parameters are varied.

This allows users and callers to modify parameters manually, and run the refinement with no varied parameters, in order to get an updated spectrum that includes their modifications.

# Affected Workflows

Powder. Specifically, WPPF

# Documentation Changes

None


psavery:
Reviewed by @saransh13

PR #936, opened 2026-05-15
Fix exclusion ordering on lattice param change

psavery:
# Overview

Previously, `tThSort` and `tThSortInv` would effectively cancel each other out, as long as the two theta ordering stayed constant (which is usually true for high symmetry systems). This issue would not be encountered by most users, since most users are dealing with high symmetry systems.

In low symmetry systems, the two theta ordering can be more easily re-arranged on lattice parameter changes, so this would land the exclusions at incorrect indices.

An example of this issue in the GUI: if you had a triclinic material and selected HKL `002`, then changed the `c` lattice parameter, the selected HKL could be changed to something else, like `-1 -1 2`.

A simple test was added that reproduced the issue, and that also verifies that the issue is fixed.

# Affected Workflows

All

# Documentation Changes

None

### DIFF
--- a/hexrd/core/material/crystallography.py
+++ b/hexrd/core/material/crystallography.py
@@ -900,7 +900,7 @@ class PlaneData(object):
         retval = np.zeros(self.getNhklRef(), dtype=bool)
         if self._exclusions is not None:
             # report in current hkl ordering
-            retval[:] = self._exclusions[self.tThSortInv]
+            retval[:] = self._exclusions[self.tThSort]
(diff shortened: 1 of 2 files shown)


saransh13 (approved):
fixes the issue

psavery:
Rebased on master to fix the packaging

psavery:
This actually seemed to be a relatively important issue with calibrating, especially with low symmetry materials, because if you calibrated lattice parameters, you could end up with wrong exclusions selected.

PR #938, opened 2026-06-01
Fix all deprecation warnings from the tests

psavery:
# Overview

* material/utils.py & wppf/phase.py — importlib.resources.open_binary -> files()/as_file(): clears open_binary `DeprecationWarning` and the unclosed-file `ResourceWarning` (mu_en.h5 / Anomalous.h5).
* rotations.py — np.float64(array) -> np.asarray(...): clears the NumPy "conversion of an array with ndim > 0 to a scalar is deprecated" `DeprecationWarning`.
* matrixutil.py — np.asmatrix -> np.atleast_2d: clears the np.matrix subclass `PendingDeprecationWarning`.
* gridutil.py — np.cross on 2-vectors -> explicit scalar form: clears the NumPy 2.0 "arrays of 2-dimensional vectors are deprecated" `DeprecationWarning`.

# Affected Workflows

None

# Documentation Changes

None

### DIFF
--- a/hexrd/core/gridutil.py
+++ b/hexrd/core/gridutil.py
@@ -178,7 +178,9 @@ def computeArea(polygon):
     area = 0
     for [s1, s2] in triv:
         tvp = np.diff(np.hstack([polygon[s1, :], polygon[s2, :]]), axis=0).flatten()
-        area += 0.5 * np.cross(tvp[:2], tvp[2:])
+        # 2D cross product (z-component); np.cross on 2-vectors is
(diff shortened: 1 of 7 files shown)


psavery:
Also just added a small commit to use `yaml.safe_load()` instead of `yaml.load()` in a couple of places.

ZackAttack614 (approved):
Glad to have things cleaned up a bit more! We're on our way to a quieter terminal, thanks @psavery .

PR #939, opened 2026-06-03
Texture uniform odf

bnmajor:
# Overview

Adds `UniformODF`, a constant orientation distribution function representing completely random (isotropic) texture. The `ODF` evaluates to a constant **1 MRD** (multiples of a random distribution) for every orientation - the standard reference density for texture analysis on SO(3).

## New API (additive only — no existing signatures changed)

- `hexrd.phase_transition.texture.UniformODF`/`hexrd.phase_transition.texture.uniform_odf`:
  - `.eval(orientations: np.ndarray) -> float | np.ndarray` - returns a scalar `1.0` for a single `(3, 3)` matrix and an array of ones for any input whose shape ends in `(3, 3)`; raises `ValueError` otherwise.

## Tests added

- Package export, constant `1 MRD` value, and MRD convention.
- Symmetry validation (valid + invalid crystal/sample)
- Evaluation: single orientation (scalar), batched `(N, 3, 3)` and multi-dimensional `(2, 2, 3, 3)` inputs, and invalid-shape handling.
- All orientations return equal values; value is `1 MRD` across crystal symmetries.
- Guard the mirrored symmetry sets against drift from `WPPF`'s `SYMLIST`

## Who might need to know

- Anyone building texture/ODF functionality on `hexrd.phase_transition.texture`, since `UniformODF` establishes the `ODF` `eval()` interface contract.
- Maintainers of `hexrd.powder.wppf.texture` - `UniformODF` mirrors (to avoid depending on `WPPF`) `SYMLIST`, and a test will fail if the crystal symmetry list there changes.

# Affected Workflows

No existing workflows change behavior - this is purely additive. New texture/ODF code can now use `UniformODF` as a random-texture reference. Nno change to existing public APIs, CLI, or file formats.

# Documentation Changes

- No separate docs site/repo updates required; the public API is documented via docstrings
  
## Other notes

- Symmetry is validated and stored but intentionally inert. The uniform ODF value is constant regardless of symmetry; the arguments exist to satisfy the common ODF interface planned for the upcoming `UnimodalODF` work.
- `'axial'` sample symmetry was dropped. The sample set is restricted to `{triclinic, monoclinic, orthorhombic}` to match `DeLaValleePoussinKernel`s supported sample symmetries. `'axial'` has no current consumer.
- Symmetry labels are mirrored from `hexrd.powder.wppf.texture.SYMLIST` rather than imported, to keep the new texture package decoupled from WPPF. A regression test guards against drift. If depending on `WPPF` is acceptable this can be changed to import the labels and the regression test can be dropped.

Depends on #926

### DIFF
--- a/hexrd/phase_transition/texture/__init__.py
+++ b/hexrd/phase_transition/texture/__init__.py
@@ -2,8 +2,10 @@ from hexrd.phase_transition.texture.kernels import (
     DeLaValleePoussinKernel,
     SO3Kernel,
 )
+from hexrd.phase_transition.texture.uniform_odf import UniformODF
 
 __all__ = [
     'DeLaValleePoussinKernel',
     'SO3Kernel',
+    'UniformODF',
 ]
--- a/hexrd/phase_transition/texture/uniform_odf.py
+++ b/hexrd/phase_transition/texture/uniform_odf.py
@@ -7,21 +7,14 @@ orientations are equally likely. The uniform ODF has a constant value of
 for texture analysis on SO(3).
 """
 
-from typing import Union
+from typing import Optional, Union
 
 import numpy as np
 
-# Valid symmetry labels. Crystal symmetries mirror the crystal portion of
-# hexrd.powder.wppf.texture.SYMLIST. Sample symmetries are the subset
-# currently supported by DeLaValleePoussinKernel; 'axial' is intentionally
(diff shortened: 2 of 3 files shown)


psavery (approved):
@bnmajor I made some minor modifications. Looks good otherwise. Merge when ready.

PR #942, opened 2026-06-05
numba accelerated special functions

saransh13:
a few of the functions such as `exp1` implemented previously were not correct for complex arguments. The latest implementation fixes that. 

Also included is a new special function `wofz`which implements the Faddeeva function for complex arguments.

Both these are needed for the new peak shapes to deal with x-ray induced heating experiments.

The PR removes some unused constants in the `herd.core.constants` and moves all the special function implementation to `hexrd.core.fitting.special`. The paths in `hexrd.powder.wppf.peakfunctions` and `hexed.core.fitting.peakfunctions` have been updated to reflect this move.

Also added tests for these new functions by comparing them to `spicy.special`.


psavery (approved):
Made a few minor fixes. LGTM

PR #943, opened 2026-06-08
Ensure conda packaging finds xsimd and eigen

psavery:
# Overview

Before, it appeared to be performing the fallback install of xsimd and eigen via the `install_build_dependencies.py` script.

Now, it should be using the packages from conda as it should, and thus the build packaging should update the versions automatically.

# Affected Workflows

All, via dependency updates

# Documentation Changes

None


psavery:
This is now suddenly needed for other PRs to actually run their packaging CI successfully, so it's a more urgent one. Merging....

PR #944, opened 2026-06-10
Add pink-beam heating peak shape

psavery:
# Overview

Adds a sigma-based **heating** pink-beam peak shape (`pvheating`, peakshape `3`)
for heating experiments, and splits the existing tau-based math into its own
**exponential** shape (`pvexponential`, peakshape `4`). Both are wired through
`fitpeak`, `spectrum`, and WPPF, with tests for each. Also adds a ruff config.

# Affected Workflows

Core: new peak shapes added, and `pink_beam_dcs` issues fixed in a few places
Powder: same but for WPPF

# Documentation Changes

None


saransh13 (approved):
LGTM. can merge after windows condo_package issue is resolved.

PR #945, opened 2026-06-11
Wppf fixes and background adjusted rwp

saransh13:
## Overview
Many of the HED experiments have an unusually high background. This results in artificially low value of the weighted R-p factor (`Rwp`) for whole pattern fits for these datasets. This PR introduces a background adjusted weighted R-p factor which is more representative of the true quality of the fit.

In addition to this changes, the computation for the chi^2 factor is greatly simplified. A bug was also discovered where the sqrt of the values that should have been reported was getting calculated.

the difference vector that was used for the minimization problem now conforms to the values used in GSAS. There was a sqrt operation that was extra in hexrd before this PR.

Finally, there are some bug fixes in `hexrd.wppf.spectrum` which fixes the missing imports and some undefined error types.

## Workflows affected
- WPPF


psavery (approved):
Made a few minor changes. Rwps are looking good. LGTM!

PR #947, opened 2026-06-15
Support fixed pink-beam asymmetry in powder calibration

psavery:
# Overview

Add an optional `fixed_pink_asymmetry` argument to PowderCalibrator and SpectrumModel that pins the pink-beam shape parameters (alpha/beta for dcs, sigma for heating, tau for exponential) from a prior WPPF refinement and holds them fixed during the fit.

This is needed because WPPF can often refine asymmetry parameters to very accurate values since the parameters are being refined for the entire spectrum. But in the PowderCalibrator and SpectrumModel, asymmetry parameters are refined for each peak, which can sometimes lead to unrealistic results.

Allowing the asymmetry params to be set from WPPF allows us to fix them to be realistic for powder calibration.

# Affected Workflows

Powder

### DIFF
--- a/hexrd/core/fitting/calibration/powder.py
+++ b/hexrd/core/fitting/calibration/powder.py
@@ -35,6 +35,7 @@ class PowderCalibrator(Calibrator):
         tth_distortion=None,
         calibration_picks=None,
         xray_source: Optional[str] = None,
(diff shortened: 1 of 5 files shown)


saransh13 (approved):
LGTM

PR #952, opened 2026-06-26
Compton background

saransh13:
Add optional argument to include compton background in the TDS calculation. This will be an important correction in temperature determination using the TDS signal.


psavery (approved):
LGTM!

PR #956, opened 2026-07-01
Stop installing conda-libmamba-solver in the package workflows

psavery:
# Overview

Installing conda-libmamba-solver into the base env pulled in conda 26.5.3 (requires Python 3.13), corrupting base conda and failing the Linux build with `ModuleNotFoundError: No module named 'conda'`.

libmamba has been the default solver since conda 23.10, so drop the explicit install/config from package.yml and container_build.sh.

# Affected Workflows

None (packaging only)

# Documentation Changes

None


psavery:
This is somewhat urgent to fix the packaging. It's only packaging related. Merging...

PR #963, opened 2026-08-31
Add old "_newton" variants of ge_41rt

psavery:
If a user specifies "_newton" at the end of the ge_41rt name, the old newton methods will be used.

We actually really need the old newton method in some cases, including a GE detector at APS, which appears to have a pincushion effect. The new ge_41rt closed-form cubic solver method **cannot** model the pincushion effect, so we must use the old newton method for it.


saransh13 (approved):
LGTM

PR #965, opened 2026-09-04
Read the space-group setting from CIF files

psavery:
# Overview

The reader discarded the setting and always used the default, so a CIF written in the second origin choice imported with its atoms in the wrong places.  Read it from the coordinate-system code or the Hermann-Mauguin suffix, reject settings hexrd cannot represent, and default consistently with the rest of hexrd when a file omits it.

# Affected Workflows

All. But only for using the second sgsetting, which is somewhat uncommon.

### DIFF
--- a/hexrd/core/material/material.py
+++ b/hexrd/core/material/material.py
@@ -72,10 +72,14 @@ def _key(x):
 
 
 def get_default_sgsetting(sgnum):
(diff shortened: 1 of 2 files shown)


psavery:
This will involve more extensive changes to the symmetry operations in hexrd, so putting this in a draft til we get those in.

PR #967, opened 2026-09-08
Avoid overflow in eta-omega histograms

### LINKED ISSUE #964: uint16 image weights silently overflow when generating eta-omega histograms
### Describe the bug

In `hexrd/core/instrument/hedm_instrument.py`, `_run_histograms()` preserves the input image `dtype` [when it is cast to an array]((https://github.com/hexrd/hexrd/blob/bea0d5281e07882f52b2cd9d2e8d88cf8702c49f/hexrd/core/instrument/hedm_instrument.py#L2540)).

Detector images can have dtype `uint16`. The image values are subsequently passed to `numpy.histogram()` as weights:

```python
result = histogram(
    pixel_etas,
    bins=eta_edges,
    weights=image[pixel_ids],
)[0]
```

numpy uses the dtype of `weights` as the dtype of the histogram accumulator, so, a `uint16` image produces a `uint16` histogram. Any eta-bin sum greater than 65,535 silently wraps around rather than being represented correctly.

A possible fix is to convert the histogram weights to a sufficiently wide type before accumulation, preferably when preparing the image:

```python
image = np.asarray(image, dtype=np.uint32)
```

### Impact

I cannot work around the issue

### Reproduction

Run the following Python snip to see the issue firsthand.

```python
import numpy as np

values = np.array([0.1, 0.2])
weights = np.array([40_000, 40_000], dtype=np.uint16)

hist, _ = np.histogram(values, bins=[0.0, 1.0], weights=weights)

print("Expected:", 80_000)
print("Actual:", hist[0])
print("Histogram dtype:", hist.dtype)
```

### Version

```shell

```

### Affected Workflows

- [x] core
- [ ] EDD
- [x] HEDM
- [ ] Powder
- [ ] Laue

psavery:
# Overview

Promote selected histogram weights to float64 before accumulation. Add a regression test for uint16 bin totals above 65,535.

# Affected Workflows

HEDM

Fixes: #964

### DIFF
--- a/hexrd/core/instrument/hedm_instrument.py
+++ b/hexrd/core/instrument/hedm_instrument.py
@@ -2553,7 +2553,11 @@ def _run_histograms(rows, ims, tth_ranges, ring_maps, ring_params, threshold):
 
             # Unpack the params
             pixel_etas, eta_edges, pixel_ids, bins_on_detector = params
-            result = histogram(pixel_etas, bins=eta_edges, weights=image[pixel_ids])[0]
+            result = histogram(
+                pixel_etas,
+                bins=eta_edges,
+                weights=np.asarray(image[pixel_ids], dtype=np.float64),
+            )[0]
 
             # Note that this preserves nan values for bins not on the detector.
             this_map[i_row, bins_on_detector] = result[bins_on_detector]
--- a/tests/core/instrument/test_hedm_instrument.py
+++ b/tests/core/instrument/test_hedm_instrument.py
@@ -1,0 +1,22 @@
+import numpy as np
+
+from hexrd.core.instrument.hedm_instrument import (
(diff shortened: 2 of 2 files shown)


psavery:
@ZackAttack614 Numpy accumulates histogram weights using the weights’ dtype, so converting to uint32 only moves the overflow boundary and does not fix the general problem. Converting just the selected weights to float64 prevents integer wrap while avoiding the memory cost of converting the entire image.

ZackAttack614 (approved):
This is great, thanks for fixing this! I'd love to hear from @donald-e-boyce about whether this resolves some of the find-orientations issues he was seeing earlier in the year.

psavery:
@donald-e-boyce Let us know when you try it out...

PR #968, opened 2026-09-10
Remove the JHE pinhole model

psavery:
# Overview

The JHE approximation disagreed in sign with Eggert's model as implemented in Rygg's own reference script, and its map variant referenced an undefined name, so it could never have run.  The one remaining model is now simply PinholeDistortion, with the old name kept as an alias.

# Affected Workflows

HED

xref: https://github.com/HEXRD/hexrdgui/pull/2051


saransh13 (approved):
LGTM.

PR #970, opened 2026-09-14
Texture basic operations

bnmajor:
# Overview

Adds `+` and `-` between ODF objects, producing a new `CompositeODF` that evaluates as a linear combination of its terms plus a constant offset in MRD:

    f(g) = c + Sum_i a_i * f_i(g)

## API changes

New, exported from `hexrd.phase_transition.texture`: `CompositeODF`, `ODFArithmetic`

Changed:

- `UniformODF` and `UnimodalODF` now inherit from `ODFArithmetic`. They gain +, -, and unary -.
- Constants are supported on either side: `odf + 2.0`, `2.0 + odf`, `odf - 2.0`, `2.0 - odf`
- `DeLaValleePoussinKernel` and `UnimodalODF` expose the resolved symmetry operators as `crystal_symmetry_quats` and `sample_symmetry_quats`.
- `texture_index` emits a `RuntimeWarning` when the sampled ODF takes negative values, since the index is then not meaningful.

## Symmetry compatibility

Operands must share the same crystal and sample symmetry. Compatibility is checked by comparing the resolved symmetry operators rather than the labels, so alias labels agree (`'ci'` vs `'triclinic'`), array-valued symmetries are checked, and an ODF with no symmetry is the identity group rather than a wildcard. Mismatches raise `ValueError`.

## Tests

- Added `tests/test_odf_arithmetic.py` (42 tests)
- Changed `tests/test_delavalleepoussin_kernel.py`: constants replaced by three helpers.
- Changed `tests/test_texture_norm.py` (one docstring)

# Affected Workflows

No existing workflows change behavior.

Who should know:

- Anyone building ODF models on `phase_transition.texture`** gains `+` and `-`, and should understand that results are not renormalized. This is documented on the methods.

# Documentation Changes

Docstrings have been added, no other documentation changes.


psavery:
@bnmajor I made a few fixes/changes in the commits I just pushed. Look over them and let me know if you agree. If you do, I'll approve.

bnmajor:
@psavery Your changes look good to me - thanks!

PR #971, opened 2026-09-15
Pin crossenv <1.6 for osx-arm64 cross build

psavery:
# Overview

crossenv 1.6.1 (conda-forge, 2026-09-14) rejects every *-apple-darwin* triple with "Unknown Apple compiler triple", aborting all osx-arm64 cross builds in the MacOSX packaging job. Pin to the previous release until robotpy/crossenv#166 is released on conda-forge.


psavery:
This was necessary to fix CI

PR #972, opened 2026-09-15
update shkl and bkg before running refinement.

saransh13:
These parameters were not getting updated if no parameters were refined, such as the case when user manually changed parameter and simply wants to compute the spectrum. Now both `shkl` and `bkg` parameters are updated before `computespectrum` is called. Also updated `Rwp`, `Rwpb` and `gofF` is printed.

# Affected Workflows
None


hexrd/powder/wppf/WPPF.py
```diff
@@ -2244,8 +2257,11 @@ def Refine(self):
                 f"Rwp: {self.Rwp * 100.0:.2f} % Rwpb: {self.Rwpb * 100.0:.2f} % and chi^2: {self.gofF:.2f}\n"
             )
         else:
-            logger.info("Nothing to refine.")
             self.computespectrum()
+            logger.info(
+                f"Nothing to refine."
```
psavery:
```suggestion
                "Nothing to refine.\n"
```

hexrd/powder/wppf/WPPF.py
psavery, line 1452: If the user manually edits parameters, we would need to ensure they get updated here before `CalcIobs()` is called.

```suggestion
        # apply any manual parameter edits before partitioning the
        # intensities, since CalcIobs uses the peak shape and shkl values
        self._set_params_vals_to_class(
            self.params, init=False, skip_phases=False, force=True
        )
        self._update_shkl(self.params)
        self._update_bkg(self.params)

        self.CalcIobs()
```

psavery: Prevent a double print of the metrics when nothing was refined.

```suggestion
        # When nothing was refined, Refine() already printed the metrics
        if print_to_screen and self.res is not None:
            logger.info(
                f"Rwp: {self.Rwp * 100.0:.2f} % Rwpb: {self.Rwpb * 100.0:.2f} % and chi^2: {self.gofF:.2f}\n"
            )
```

psavery: I think this suggestion for some reason didn't show that the previous `self.CalcIobs()` should be deleted, or moved to line 1452, so if you accept this, make sure we don't get duplicate `self.CalcIobs()` calls.

psavery: Same for this suggestion. Seems it didn't work properly for some reason.

psavery: Maybe manually copy/paste them

psavery (approved):

